PR #1, opened 2020-09-15
Add TRITONBACKEND_BackendArtifacts

deadeyegoodwin:


deadeyegoodwin:
@Tabrizian  This change adds the TRITONBACKEND API that allows the backend to find the directory its shard library is from. You will need to use this to find the other python backend artifacts (like startup.py).

Tabrizian:
@deadeyegoodwin I will use this API to remove the hard-coded paths in Python backend.

Tabrizian (dismissed):

PR #6, opened 2020-10-30
WIP: Integrating RateLimiter with the server and dynamic batch scheduler

tanmayv25:


include/triton/core/tritonbackend.h
```diff
@@ -34,19 +34,19 @@ extern "C" {
 #endif
 
 #ifdef _COMPILING_TRITONBACKEND
```
deadeyegoodwin, line 36: Are all the changes in this file just from the formatter?

include/triton/core/tritonserver.h
```diff
@@ -1230,6 +1239,37 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error*
...
+/// \param mode The mode to use for the rate limiting.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error*
+TRITONSERVER_ServerOptionsSetRateLimitMode(
+    TRITONSERVER_ServerOptions* options, TRITONSERVER_RateLimitMode mode);
+
+/// Add resource count for rate limiting.
+///
+/// \param options The server options object.
+/// \param device The device identifier for the resource. The resource
```
deadeyegoodwin, line 1262: We know if the resource is global or not from the model configurations (btw, what if the same resource is listed as both global and non-global? Do you log a warning? How do you proceed). 
I would put "device" as the last parameter and, for consistency, I would make device "int" type. 
"The device identifier for the resource. A value of -1 indicates that the specified number of resources are available on every device. Ignored for a global resource."
I guess it is possible to set resources so that a model instance will never be able to run. It would be nice if we could detect and log an error for that. File a separate PR for that functionality.

include/triton/core/tritonserver.h
```diff
@@ -1230,6 +1239,37 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error*
 TRITONSERVER_ServerOptionsSetStrictModelConfig(
     TRITONSERVER_ServerOptions* options, bool strict);
 
+/// Set the rate limit mode in a server options.
+///
+///   TRITONSERVER_RATE_LIMIT_OFF: The rate limiting is turned off and the
+///   inference gets executed whenever an instance is available.
+///
+///   TRITONSERVER_RATE_LIMIT_EXEC_COUNT: The rate limiting prioritizes the
```
deadeyegoodwin, line 1247: This should be the default.

deadeyegoodwin (changes requested):

include/triton/core/tritonbackend.h
tanmayv25: All the changes from this commit 292cfe2649d5ebb3b are from the formatter. I can quickly share a PR for (DLIS-1738)[https://jirasw.nvidia.com/browse/DLIS-1738] to simplify this change. I think I have all the code ready, just need to make sure all the repos are formatted.

include/triton/core/tritonserver.h
tanmayv25: Why? Shouldn't wee keep the rate limiting off by default?

include/triton/core/tritonserver.h
tanmayv25: > We know if the resource is global or not from the model configurations.

What about the resources that are not specified by any loaded models? I understand in this case we will not be using the resource count from CLI options anyway. We can get the resource type while registering model to the rate limiter. See the next response for what it would entail.

> what if the same resource is listed as both global and non-global?

The current implementation will treat the resource with the same name but different kind (global vs non-global) as separate resources. So, even if they have the same name, the counts that are compared will be from their own kinds. However, if we decide to extract kinds from the model configuration then we would need to have unique resource names.

I think it's a good idea and will keep our TRITONSERVER API clean to have int device.

include/triton/core/tritonserver.h
```diff
@@ -1233,6 +1239,38 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error*
...
+    TRITONSERVER_ServerOptions* options, TRITONSERVER_RateLimitMode mode);
+
+/// Add resource count for rate limiting.
+///
+/// \param options The server options object.
+/// \param name The name of the resource.
+/// \param count The count of the resource.
+/// \param device The device identifier for the resource. A value of -1
+/// indicates that the specified number of resources are available on every
+/// device. Ignored for a global resource.
```
deadeyegoodwin, line 1267: I don't see how global resource is handled here. I believe we may also have to make an adjustment on the cmdline to handle global resources.

PR #9, opened 2020-11-20
Add tritonrepoagent.h

deadeyegoodwin:


include/triton/core/tritonrepoagent.h
```diff
@@ -0,0 +1,307 @@
...
+///
+/// \param agent The agent.
+/// \param model The model.
+/// \param artifact_type Returns the artifact type for the location.
+/// \param path Returns the location.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONREPOAGENT_DECLSPEC TRITONSERVER_Error*
+TRITONREPOAGENT_ModelRepositoryLocationAcquire(
+    TRITONREPOAGENT_Agent* agent, TRITONREPOAGENT_AgentModel* model,
+    TRITONREPOAGENT_ArtifactType* artifact_type, const char** location);
```
GuanLuo, line 180: What happens if the function is called multiple time? I assume it will always return the same temporary location?

deadeyegoodwin: Yes. Also, I'm likely going to change artifact_type to an in/out so that the agent can request a certain type of location (initially there is just "local filesystem") and then triton will prefer to return that type of location (but agent must handle any returned type). With that change will then need to always return same artifact type as well if called multiple times.

PR #10, opened 2020-12-02
Add TRITONBACKEND_ModelConfigSerializedToString

CoderHam:
Needed by custom backend to correctly load using the serialized model config

### DIFF
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -27,6 +27,7 @@
 
 #include <stddef.h>
 #include <stdint.h>
+#include <string>
 #include "triton/core/tritonserver.h"
 
 #ifdef __cplusplus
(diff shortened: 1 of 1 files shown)


GuanLuo (changes requested):
In the backends we use combination of `TRITONBACKEND_ModelConfig()` and `TRITONSERVER_MessageSerializeToJson()` to obtain the model config in JSON string, why do we need this?

deadeyegoodwin:
We don't need this. Look at the identity example. You get the config as a TRITONSERVER_Message and then you can serialize that message to json.

CoderHam:
This is to create a string serialized message and not json

CoderHam:
I do not need a Json object. I need a serialized string (using [SerializeToString()](https://github.com/triton-inference-server/server/blob/master/src/backends/custom/custom_backend.cc)), To do this we need the protobuf message which is not a part of the custom library. This is why we need to expose this function. This is different from [Serializing to a Json String](https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.util.json_util#MessageToJsonString.details). 
To clarify we need to use SerializeToString to maintain backward compatibility in the custom backend.

```
bool SerializeToString(string* output) const;: serializes the message and stores the bytes in the given string. Note that the bytes are binary, not text; we only use the string class as a convenient container.
```

GuanLuo:
I see, that's unfortunate that serialized protobuf is implicitly enforced to the old custom backend. In this case, rather than adding the serialize function here, you should add serialize / deserialize protobuf functions onto `TRITONSERVER_Message` object for consistency

deadeyegoodwin:
We cannot have protobuf anything in any of the external Triton APIs. I don't understand why the json string that you can get from the current TRITONBACKEND API can't be parsed into a protobuf and then serialized and passed to the custom backend. We should discuss offline.

CoderHam:
> I see, that's unfortunate that serialized protobuf is implicitly enforced to the old custom backend. In this case, rather than adding the serialize function here, you should add serialize / deserialize protobuf functions onto `TRITONSERVER_Message` object for consistency

That is where I intended on keeping it first but the protobuf functions are not exposed there. Additionally to serialize to string a generic protobuf the message needs to be exposed to this class.

PR #15, opened 2021-04-02
Add Triton unload_dependent option to unload API

GuanLuo:


include/triton/core/tritonserver.h
```diff
@@ -1634,6 +1634,18 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerLoadModel(
 TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerUnloadModel(
     TRITONSERVER_Server* server, const char* model_name);
 
+/// Unload the requested model, and also unload the models that only be used
```
deadeyegoodwin, line 1637: "Unload the requested model, and also unload any dependent model that was loaded along with the requested model (for example, the models composing an ensemble). Unloading a model ... will be returned. The function does not return until the requested model and all dependent models are unloaded or fail to unload. Returned error indicates if all models unloaded successfully or not."

include/triton/core/tritonserver.h
```diff
@@ -1634,6 +1634,18 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerLoadModel(
...
+/// by the requested model. Unloading a model that is not loaded
+/// on server has no affect and success code will be returned.
+/// The function does not return until the model is unloaded or fails to unload.
+/// Returned error indicates if model unloaded successfully or not.
```
deadeyegoodwin, line 1641: What happens if only some of the models fail to unload. Do you still unload the others? Is the returned error just the error from one of the failed unloads?

include/triton/core/tritonserver.h
```diff
@@ -1634,6 +1634,18 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerLoadModel(
...
+///
+/// \param server The inference server object.
+/// \param model_name The name of the model.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerCascadingUnloadModel(
```
deadeyegoodwin, line 1646: TRITONSERVER_ServerUnloadModelAndDependents

src/tritonserver_stub.cc
```diff
@@ -426,6 +426,10 @@ TRITONSERVER_ServerUnloadModel()
 {
 }
 TRITONAPI_DECLSPEC void
+TRITONSERVER_ServerCascadingUnloadModel()
```
deadeyegoodwin, line 429: TRITONSERVER_ServerUnloadModelAndDependents

deadeyegoodwin (changes requested):

include/triton/core/tritonserver.h
GuanLuo: Actually there is no failure case for unloading a model. Even in the case of there is inflight requests, this function doesn't really wait for the unload to complete, it let model manager to drop its smart pointer to the model and return success.

deadeyegoodwin: We should fix the comment then, and in the ServerUnloadModel function as well.

deadeyegoodwin (approved):

PR #16, opened 2021-04-29
Add TRITONBACKEND_EXECUTION_DEVICE_BLOCKING

deadeyegoodwin:


tanmayv25 (approved):
I get the gist of it. We would need to add some callback logic to inform when execution is actually completed so that allotted resources can be returned to the available pool in the rate limiter.

tanmayv25:
This change is merged into `tanmayv-dev` and will me available in main branch when tanmayv-dev is merged into main.

PR #25, opened 2021-07-29
Document the ownership of buffer holding request object pointers

tanmayv25:
This is necessary for the asynchronous backends like TRT.


GuanLuo (approved):
As we discussed, documenting this for now is fine, but I think it would be nice if we can handle that internally in the future so the backend doesn't need to be aware of that kind of detail (i.e. the array is reserved until all request is released by the backend, but that will an overhead that a callback needs to be attached to the requests). Can you create a ticket for that, we can continue the discussion there.

tanmayv25:
>  Can you create a ticket for that, we can continue the discussion there.

DLIS-2752

PR #33, opened 2021-10-07
Add response cache byte size API

tanmayv25:


include/triton/core/tritonserver.h
```diff
@@ -930,6 +930,17 @@ TRITONSERVER_InferenceRequestAppendInputDataWithHostPolicy(
     const void* base, size_t byte_size, TRITONSERVER_MemoryType memory_type,
     int64_t memory_type_id, const char* host_policy_name);
 
+/// Set the total response cache byte size that the server can allocate in CPU
```
deadeyegoodwin, line 933: Move this next to TRITONSERVER_ServerOptionsSetCudaMemoryPoolByteSize() both here and in the stubs

deadeyegoodwin (changes requested):

PR #34, opened 2021-10-07
Added string correlation ID API

aramesh7:


deadeyegoodwin (commented):
Looks good but you need to update the minor version number at the top of the file. Also you need to include any new API in the stubs file.

include/triton/core/tritonbackend.h
```diff
@@ -337,15 +337,26 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_OutputBuffer(
...
     TRITONBACKEND_Request* request, uint64_t* id);
 
+/// Get the correlation ID of the request if it is a string. 
+/// Empty string indicates that the request does not have a correlation ID. 
+/// Returns error if correlation ID for given request is not a string.
+///
+/// \param request The inference request.
+/// \param id Returns the correlation ID.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_RequestCorrelationIdString(
```
deadeyegoodwin, line 357: TRITONBACKEND functions need to go in the stub as well. Also need to bump the minor version in this file.

src/tritonserver_stub.cc
```diff
@@ -694,10 +710,37 @@ TRITONBACKEND_ModelInstanceReportBatchStatistics()
 {
 }
 TRITONAPI_DECLSPEC void
+TRITONBACKEND_Initialize()
```
deadeyegoodwin, line 713: Thanks for auditing... but these don't go in the stub. The stub is needed only for backend/server functions that are called from backend code itself. These functions are called from Triton into the backend.

aramesh7: Ah okay. Fixed. Hopefully there was none in there that DID need go into the stub.

PR #35, opened 2021-11-03
Document the exec policy override in the TRITONBACKEND API

tanmayv25:


include/triton/core/tritonbackend.h
```diff
@@ -755,6 +757,9 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_BackendExecutionPolicy(
 /// the backend's execution policy after calling
 /// TRITONBACKEND_Initialize, so to be recognized changes to the
 /// execution policy must be made in TRITONBACKEND_Initialize.
+/// Also, note that if loading a Sequence Model in the backend,
```
deadeyegoodwin, line 760: Say more explicitly "if using the sequence batcher for the model" as that is exactly when this happens. Is it for both direct and oldest?

tanmayv25: Yes.. The behavior is for both kinds. We previously had this behavior when TRT backend was in server repo. Moving back to multiple runners for sequence batcher models because of the reasoning in the discussion here:
https://github.com/triton-inference-server/server/pull/3516#discussion_r741598340

GuanLuo (dismissed):

deadeyegoodwin (approved):

include/triton/core/tritonbackend.h
```diff
@@ -755,6 +757,9 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_BackendExecutionPolicy(
 /// the backend's execution policy after calling
 /// TRITONBACKEND_Initialize, so to be recognized changes to the
 /// execution policy must be made in TRITONBACKEND_Initialize.
+/// Also, note that if using sequence batcher for the model, Triton will
+/// use TRITONBACKEND_EXECUTION_BLOCKING policy irrespective to which
```
CoderHam, line 761: irrespective **of the policy** specified by this setter function.

CoderHam (changes requested):

CoderHam (approved):

PR #37, opened 2021-11-18
Add allocator query APIs

GuanLuo:
@deadeyegoodwin This is how the new APIs look and feel. The caveat is that the memory type returned by the query API isn't necessarily consistent with the buffer returned by the allocate API. Although for simple allocator (HTTP / GRPC frontend / MLPerf harness), it may not be the issue, but it may be less useful for dynamic allocator (i.e. uses memory pool and has some fallback mechanism).

### DIFF
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -91,7 +91,7 @@ struct TRITONBACKEND_ModelInstance;
 ///   }
 ///
 #define TRITONBACKEND_API_VERSION_MAJOR 1
-#define TRITONBACKEND_API_VERSION_MINOR 6
+#define TRITONBACKEND_API_VERSION_MINOR 7
 
 /// Get the TRITONBACKEND API version supported by Triton. This value
 /// can be compared against the TRITONBACKEND_API_VERSION_MAJOR and
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -594,6 +594,28 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_ResponseOutput(
     const char* name, const TRITONSERVER_DataType datatype,
     const int64_t* shape, const uint32_t dims_count);
 
+/// Query the response allocator for the preferred memory type and memory
+/// type ID, which will be the attributes of the allocated
+/// buffers if explicitly requested in the sequential call of
+/// TRITONBACKEND_OutputBuffer. 'byte_size' is an optional parameter to
+/// provide more information to the allocator. Note that if 'byte_size' is not
+/// provided, the returned values may not match the attributes of the allocated
+/// buffer even if explicitly requested, as the allocator may allocate
+/// buffers differently based on the requested byte size.
+///
+/// \param response The response.
+/// \param byte_size The expected size of the buffer. This is optional
+/// and it should be set to nullptr to indicate that the byte size has
+/// not determined.
+/// \param memory_type The allocator preferred type of memory.
+/// \param memory_type_id The allocator preferred ID of the memory.
+/// \return a TRITONSERVER_Error object if a failure occurs.
+/// A TRITONSERVER_ERROR_UNAVAILABLE error indicates that the allocator
+/// doesn't provide query functionality.
+TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_ResponseAllocatorQuery(
+    TRITONBACKEND_Response* response, size_t* byte_size,
+    TRITONSERVER_MemoryType* memory_type, int64_t* memory_type_id);
+
 /// Send a response. Calling this function transfers ownership of the
 /// response object to Triton. The caller must not access or delete
 /// the response object after calling this function.
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -604,6 +604,9 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_ResponseOutput(
 /// buffers differently based on the requested byte size.
 ///
 /// \param response The response.
+/// \param name The name of the output tensor. This is optional
+/// and it should be set to nullptr to indicate that the tensor name has
+/// not determined.
 /// \param byte_size The expected size of the buffer. This is optional
 /// and it should be set to nullptr to indicate that the byte size has
 /// not determined.
@@ -613,7 +616,7 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_ResponseOutput(
 /// A TRITONSERVER_ERROR_UNAVAILABLE error indicates that the allocator
 /// doesn't provide query functionality.
 TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_ResponseAllocatorQuery(
-    TRITONBACKEND_Response* response, size_t* byte_size,
+    TRITONBACKEND_Response* response, const char* name,  size_t* byte_size,
     TRITONSERVER_MemoryType* memory_type, int64_t* memory_type_id);
 
 /// Send a response. Calling this function transfers ownership of the
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -452,8 +452,12 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_RequestOutputName(
 /// \param byte_size The expected size of the buffer. This is optional
 /// and it should be set to nullptr to indicate that the byte size has
 /// not determined.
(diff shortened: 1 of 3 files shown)


include/triton/core/tritonserver.h
```diff
@@ -378,6 +378,31 @@ typedef TRITONSERVER_Error* (*TRITONSERVER_ResponseAllocatorReleaseFn_t)(
...
+/// \param allocator The allocator that is provided in the call to
+/// TRITONSERVER_InferenceRequestSetResponseCallback.
+/// \param buffer_userp The user-specified value associated
+/// \param tensor_name The name of the output tensor. This is optional
+/// and it should be set to nullptr to indicate that the tensor name has
+/// not determined.
+/// \param byte_size The expected size of the buffer. This is optional
+/// and it should be set to nullptr to indicate that the byte size has
+/// not determined.
+/// \param memory_type The preferred type of memory.
```
deadeyegoodwin, line 398: Make these in/out paramaters just like in alloc_fn. That way the backend can communicate its preference.

include/triton/core/tritonserver.h
```diff
@@ -378,6 +378,31 @@ typedef TRITONSERVER_Error* (*TRITONSERVER_ResponseAllocatorReleaseFn_t)(
 typedef TRITONSERVER_Error* (*TRITONSERVER_ResponseAllocatorStartFn_t)(
     TRITONSERVER_ResponseAllocator* allocator, void* userp);
 
+/// Type for function that is called to query the allocator's preferred memory
```
deadeyegoodwin, line 381: Move this up to be just after the alloc_fn type definition.

include/triton/core/tritonserver.h
```diff
@@ -378,6 +378,31 @@ typedef TRITONSERVER_Error* (*TRITONSERVER_ResponseAllocatorReleaseFn_t)(
...
+/// type and memory type ID, which will be the attributes of the allocated
+/// buffers if explicitly requested. 'tensor_name' and 'byte_size' are optional
+/// parameters for the allocator. Note that if 'tensor_name' or 'byte_size' is
```
deadeyegoodwin, line 384: Replace "Note that..." with
"As much as possible, the allocator should attempt to return the same memory_type and memory_type_id values that will be returned by the subsequent call to TRITONSERVER_ResponseAllocatorAllocFn_t. But the allocator is not required to do so, and the backend must be capable of handling cases where the values differ."

include/triton/core/tritonbackend.h
```diff
@@ -594,6 +594,31 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_ResponseOutput(
     const char* name, const TRITONSERVER_DataType datatype,
     const int64_t* shape, const uint32_t dims_count);
 
+/// Query the response allocator for the preferred memory type and memory
```
deadeyegoodwin, line 597: This needs to be TRITONBACKEND_RequestOutputBufferProperties because the backend may need to determine this information before it generates any of the response objects. Need to document that a return TRITONSERVER_ERROR_UNAVAILABLE indicates that the properties are not available and all other returns indicate an error.

deadeyegoodwin (changes requested):

GuanLuo:
@askhade FYI, this is the new APIs for https://github.com/triton-inference-server/server/issues/3364

include/triton/core/tritonbackend.h
```diff
@@ -439,6 +439,29 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_RequestOutputName(
...
+/// be capable of handling cases where the values differ.
+///
+/// \param request The request.
+/// \param name The name of the output tensor. This is optional
+/// and it should be set to nullptr to indicate that the tensor name has
+/// not determined.
+/// \param byte_size The expected size of the buffer. This is optional
+/// and it should be set to nullptr to indicate that the byte size has
+/// not determined.
+/// \param memory_type Triton preferred type of memory.
```
deadeyegoodwin, line 455: These should be input/output to match the query_fn

include/triton/core/tritonserver.h
```diff
@@ -344,6 +344,33 @@ typedef TRITONSERVER_Error* (*TRITONSERVER_ResponseAllocatorAllocFn_t)(
...
 
+/// Type for function that is called to query the allocator's preferred memory
+/// type and memory type ID. As much as possible, the allocator should attempt
+/// to return the same memory_type and memory_type_id values that will be
+/// returned by the subsequent call to TRITONSERVER_ResponseAllocatorAllocFn_t.
+/// But the allocator is not required to do so.
+///
+/// \param allocator The allocator that is provided in the call to
+/// TRITONSERVER_InferenceRequestSetResponseCallback.
+/// \param buffer_userp The user-specified value associated
```
deadeyegoodwin, line 355: Remove buffer_userp here and in the function.

include/triton/core/tritonserver.h
```diff
@@ -344,6 +344,33 @@ typedef TRITONSERVER_Error* (*TRITONSERVER_ResponseAllocatorAllocFn_t)(
...
+/// \param memory_type Acts as both input and output. On input gives
+/// the memory type preferred by the caller. Returns memory type preferred
+/// by the allocator taken account to the caller preferred type.
+/// \param memory_type_id Acts as both input and output. On input gives
+/// the memory type ID preferred by the caller. Returns memory type ID preferred
+/// by the allocator taken account to the caller preferred type ID.
+/// \return a TRITONSERVER_Error object if a failure occurs.
+typedef TRITONSERVER_Error* (*TRITONSERVER_ResponseAllocatorQueryFn_t)(
+    TRITONSERVER_ResponseAllocator* allocator, void* buffer_userp,
+    const char* tensor_name, size_t* byte_size,
```
deadeyegoodwin, line 371: byte_size should not be a pointer

include/triton/core/tritonserver.h
```diff
@@ -433,6 +460,23 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ResponseAllocatorNew(
...
+/// The thread-safy requirement for query_fn is the same as other allocator
+/// callbacks.
+///
+/// \param allocator The response allocator object.
+/// \param query_fn The function to call to query allocator's preferred memory
+/// type and memory type ID.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error*
+TRITONSERVER_ResponseAllocatorSetQueryFunction(
+    TRITONSERVER_ResponseAllocator** allocator,
```
deadeyegoodwin, line 477: Just single * for allocator

deadeyegoodwin (changes requested):

include/triton/core/tritonserver.h
GuanLuo: This is misnamed. It should be `usrp` which is the user pointer associated to the allocator rather than the buffer.

include/triton/core/tritonserver.h
GuanLuo: Don't we want the byte size to be optional? I use nullptr to distinguish if byte size is provided.

deadeyegoodwin: Right. Keep it as it is.

include/triton/core/tritonbackend.h
```diff
@@ -439,6 +439,33 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_RequestOutputName(
...
+/// \param memory_type Acts as both input and output. On input gives
+/// the memory type preferred by the caller. Returns memory type preferred
+/// by Triton, taken account of the caller preferred type.
+/// \param memory_type_id Acts as both input and output. On input gives
+/// the memory type ID preferred by the caller. Returns memory type ID preferred
+/// by Triton, taken account of the caller preferred type ID.
+/// \return a TRITONSERVER_Error object if a failure occurs.
+/// A TRITONSERVER_ERROR_UNAVAILABLE error indicates that the properties are not
+/// available, other error codes indicate an error.
+TRITONBACKEND_DECLSPEC TRITONSERVER_Error*
```
deadeyegoodwin, line 464: Increment the minor version

include/triton/core/tritonserver.h
```diff
@@ -344,6 +344,35 @@ typedef TRITONSERVER_Error* (*TRITONSERVER_ResponseAllocatorAllocFn_t)(
     TRITONSERVER_MemoryType* actual_memory_type,
     int64_t* actual_memory_type_id);
 
+/// Type for function that is called to query the allocator's preferred memory
```
deadeyegoodwin, line 347: Increment minor version

deadeyegoodwin (changes requested):
Looks good. Just need to increment the API versions

deadeyegoodwin (approved):

PR #40, opened 2022-02-16
Add new load model API for config override

GuanLuo:


include/triton/core/tritonserver.h
```diff
@@ -1862,6 +1880,24 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerModelIndex(
...
+/// - "config" : string parameter that contains a JSON representation of the
+/// model configuration. This config will be used for loading the model instead
+/// of the one in the model directory.
+///
+/// \param server The inference server object.
+/// \param model_name The name of the model.
+/// \param parameters The array of load parameters.
+/// \param parameter_count The number of parameters.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerLoadModel(
```
deadeyegoodwin, line 1897: For C there is no name mangling, so need a different name for this function. Perhaps TRITONSERVER_ServerLoadModelWithParameters

deadeyegoodwin (changes requested):

include/triton/core/tritonserver.h
GuanLuo: Thinking about possible future extension on loading multiple models at once, should we make this multi model (still only support single model currently) so we don't need to add another API when we add the support?

deadeyegoodwin: Is it not sufficient to just call ServerLoadModel multiple times to load multiple models? I'd rather not add "multi" to all the load/unload protocol unless necessary

deadeyegoodwin (approved):

tanmayv25 (dismissed):

CoderHam (dismissed):

tanmayv25 (approved):

tanmayv25 (approved):

CoderHam (approved):

PR #41, opened 2022-03-02
Add buffer attributes API

Tabrizian:
The look-and-feel of the buffer attributes API. I have filed a ticket to add the buffer attributes API to host policy.


include/triton/core/tritonbackend.h
```diff
@@ -64,6 +64,7 @@ struct TRITONBACKEND_Response;
 struct TRITONBACKEND_Backend;
 struct TRITONBACKEND_Model;
 struct TRITONBACKEND_ModelInstance;
+struct TRITONBACKEND_BufferAttributes;
```
deadeyegoodwin, line 67: TRITONSERVER_BufferAttributes and the New/Delete/Get/Set methods should all be defined in tritonserver.h

deadeyegoodwin (changes requested):

include/triton/core/tritonbackend.h
```diff
@@ -711,6 +743,58 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_StateBuffer(
...
+/// not exist for the given buffer, the handle will contain nullptr.  The
+/// returned 'handle' is owned by the 'buffer_attributes' and so should not be
+/// modified or freed by the caller. The lifetime of the 'handle' matches that
+/// of the 'buffer_attributes'.
+///
+/// \param buffer_attributes The buffer attributes object.
+/// \param handle Returns the memory type id of the buffer.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONBACKEND_DECLSPEC TRITONSERVER_Error*
+TRITONBACKEND_BufferAttributesCudaIpcHandle(
```
GuanLuo, line 795: So we will have explicit API for retrieving different buffer attributes instead of something similar to output parameters?

Tabrizian: I discussed this with @deadeyegoodwin. I'm open to having something like response parameters that is more flexible. @deadeyegoodwin what do you think?

deadeyegoodwin: Output parameters API has a generic key-value API because the underlying support is for generic key-value. In this case there are only specific values that are allowed and so we should follow the existing pattern for this kind of thing (like ServerOptions).

include/triton/core/tritonserver.h
```diff
@@ -362,6 +363,35 @@ typedef TRITONSERVER_Error* (*TRITONSERVER_ResponseAllocatorAllocFn_t)(
...
+/// \param allocator The allocator that is provided in the call to
+/// TRITONSERVER_InferenceRequestSetResponseCallback.
+/// \param tensor_name The name of the output tensor to allocate for.
+/// \param userp The user data pointer that is provided as
+/// 'response_allocator_userp' in the call to
+/// TRITONSERVER_InferenceRequestSetResponseCallback.
+/// \param preferred_buffer_attributes The buffer attributes that the caller
+/// prefers for the buffer.
+/// \param actual_buffer_attributes Returns the buffer attributes associated
+/// with the allocated buffer.
```
Tabrizian, line 378: The response allocator accepts a buffer attributes for both preferred and actual arguments.

include/triton/core/tritonbackend.h
```diff
@@ -711,6 +743,67 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_StateBuffer(
...
+    TRITONBACKEND_BufferAttributes* buffer_attributes, void** handle);
```
GuanLuo, line 805: should the "handle" type just be the cuda IPC handle?

include/triton/core/tritonserver.h
```diff
@@ -480,6 +532,62 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ResponseAllocatorNew(
...
+/// tensors.
+/// \param release_fn The function to call when the server no longer
+/// holds a reference to an allocated buffer.
+/// \param start_fn The function to call to indicate that the
+/// subsequent 'alloc_fn' calls are for a new response. This callback
+/// is optional (use nullptr to indicate that it should not be
+/// invoked).
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error*
+TRITONSERVER_ResponseAllocatorNewWithBufferAttributes(
```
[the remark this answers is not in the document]
Tabrizian: Makes sense. Will modify it to use a callback similar to the query API.

include/triton/core/tritonserver.h
```diff
@@ -362,6 +363,37 @@ typedef TRITONSERVER_Error* (*TRITONSERVER_ResponseAllocatorAllocFn_t)(
...
+/// released and will also be returned by TRITONSERVER_InferenceResponseOutput.
+/// \return a TRITONSERVER_Error object if a failure occurs while
+/// attempting an allocation. If an error is returned all other return
+/// values will be ignored.
+typedef TRITONSERVER_Error* (
+    *TRITONSERVER_ResponseAllocatorBufferAttributesAllocFn_t)(
+    TRITONSERVER_ResponseAllocator* allocator, const char* tensor_name,
+    TRITONSERVER_BufferAttributes* preferred_buffer_attributes,
+    TRITONSERVER_BufferAttributes** actual_buffer_attributes, void* userp,
+    void** buffer, void** buffer_userp);
```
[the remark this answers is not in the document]
Tabrizian: The buffer size is included in the TRITONSERVER_BufferAttributes.

include/triton/core/tritonbackend.h
Tabrizian: I thought about this but this would add a dependency to the cuda runtime header file which may not be available in all circumstances (e.g. CPU only builds).

include/triton/core/tritonbackend.h
```diff
@@ -293,6 +294,24 @@ TRITONBACKEND_InputBufferForHostPolicy(
     const uint32_t index, const void** buffer, uint64_t* buffer_byte_size,
     TRITONSERVER_MemoryType* memory_type, int64_t* memory_type_id);
 
+/// Get the buffer attributes associated with the given input buffer. The
+/// returned 'buffer_attributes' is owned by the input so should not be freed by
```
deadeyegoodwin, line 298: "The returned..." is repeated below so I would remove this sentence.

include/triton/core/tritonbackend.h
```diff
@@ -293,6 +294,24 @@ TRITONBACKEND_InputBufferForHostPolicy(
...
+/// accessed after the input tensor object is released.
+///
+/// \param input The input tensor.
+/// \param index The index of the buffer. Must be 0 <= index < buffer_count,
+/// where buffer_count is the value returned by TRITONBACKEND_InputProperties.
+/// \param buffer_attributes Returns the attributes for the given buffer.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_InputBufferAttributes(
+    TRITONBACKEND_Input* input, const uint32_t index,
+    TRITONBACKEND_BufferAttributes* buffer_attributes);
```
deadeyegoodwin, line 313: TRITONSERVER_BufferAttributes

include/triton/core/tritonbackend.h
deadeyegoodwin: Remove this and use TRITONSERVER_BufferAttributes everywhere

deadeyegoodwin (changes requested):

deadeyegoodwin (changes requested):

src/tritonserver_stub.cc
```diff
@@ -574,6 +626,26 @@ TRITONBACKEND_StateUpdate()
 {
 }
 TRITONAPI_DECLSPEC void
+TRITONBACKEND_StateBufferAttributes()
```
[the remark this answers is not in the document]
Tabrizian: Good catch. Fixed.

include/triton/core/tritonserver.h
```diff
@@ -1088,6 +1196,23 @@ TRITONSERVER_InferenceRequestAppendInputDataWithHostPolicy(
     const void* base, size_t byte_size, TRITONSERVER_MemoryType memory_type,
     int64_t memory_type_id, const char* host_policy_name);
 
+/// Assign a buffer of data to an input. The buffer will be appended
+/// to any existing buffers for that input. The 'inference_request'
+/// object takes ownership of the buffer and so the caller should not
```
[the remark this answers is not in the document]
Tabrizian: I updated this but removed it later. Things got a bit complicated if we wanted to transfer the ownership. In order to simplify things each of the buffers (output, input, state) have an internal buffer attributes object. If the frontends want to change it they will provide their own buffer attribute and it will just copy them into the internal buffer and the frontends should delete it separately.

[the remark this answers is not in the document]
GuanLuo: I see, so when this function get called, the content of `buffer_attributes` will be copied to the internal object.

include/triton/core/tritonserver.h
```diff
@@ -362,6 +363,26 @@ typedef TRITONSERVER_Error* (*TRITONSERVER_ResponseAllocatorAllocFn_t)(
...
 
+/// Type for allocation function that allocates a buffer to hold an
+/// output tensor with buffer attributes. The callback function must fill in the
+/// appropriate buffer attributes information related to this buffer.
+///
+/// \param allocator The allocator that is provided in the call to
+/// TRITONSERVER_InferenceRequestSetResponseCallback.
+/// \param tensor_name The name of the output tensor to allocate for.
+/// \param buffer_attributes The buffer attributes associated with the buffer.
+/// \param buffer Returns a pointer to the allocated memory.
```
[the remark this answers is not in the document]
Tabrizian: We discussed this offline.

include/triton/core/tritonserver.h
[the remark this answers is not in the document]
Tabrizian: Correct.

deadeyegoodwin (approved):

PR #44, opened 2022-03-18
Add API for adding raw binary input

GuanLuo:


src/infer_request.cc
```diff
@@ -682,6 +714,59 @@ InferenceRequest::Normalize()
...
+    }
+    int64_t dynamic_axis = -1;
+    size_t element_cnt = 1;
+    for (const auto& dim : config_input.dims()) {
+      if (dim == WILDCARD_DIM) {
+        if (dynamic_axis != -1) {
+          return Status(
+              Status::Code::INVALID_ARG,
+              "The shape of the raw input '" + raw_input_name_ +
+                  "' can not be deduced because there are more than two "
```
deadeyegoodwin, line 747: "... because there is more than one dynamic..."

src/infer_request.cc
```diff
@@ -682,6 +714,59 @@ InferenceRequest::Normalize()
...
+                  "dynamic dimension");
```
deadeyegoodwin, line 748: I think we consistently use the term "variable-sized dimension" instead of "dynamic dimension"

src/infer_request.cc
```diff
@@ -682,6 +714,59 @@ InferenceRequest::Normalize()
...
+      } else {
+        element_cnt *= (size_t)dim;
+      }
+      shape.emplace_back(dim);
+    }
+    if ((config_input.data_type() == inference::DataType::TYPE_STRING) &&
+        ((dynamic_axis != -1) || (element_cnt != 1))) {
+      return Status(
+          Status::Code::INVALID_ARG,
+          "Currently only input shape [1] is allowed for raw input '" +
```
deadeyegoodwin, line 760: "For BYTE datatype raw input, the model must have input shape [1]"

deadeyegoodwin (changes requested):

CoderHam (approved):

deadeyegoodwin (dismissed):

tanmayv25 (approved):

include/triton/core/tritonserver.h
```diff
@@ -1071,6 +1071,20 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_InferenceRequestAddInput(
     const TRITONSERVER_DataType datatype, const int64_t* shape,
     uint64_t dim_count);
 
+/// Add a raw input to a request. The model input name, data type and shape of
+/// the input will be deduced from model configuration, this function must at
+/// most be called once on request with no other input to ensure the deduction
```
CoderHam, line 1076: ".This function must be called at most once per request for a single input. This ensures the deduction will be accurate."

include/triton/core/tritonserver.h
```diff
@@ -1071,6 +1071,20 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_InferenceRequestAddInput(
     const TRITONSERVER_DataType datatype, const int64_t* shape,
     uint64_t dim_count);
 
+/// Add a raw input to a request. The model input name, data type and shape of
```
CoderHam, line 1074: Redundant to have input twice, 

"The model input name, data type and shape will be ..."

CoderHam (changes requested):
Some files need copyright updates and re-word description for TRITONSERVER_InferenceRequestAddRawInput.

Otherwise LGTM

include/triton/core/tritonserver.h
```diff
@@ -1071,6 +1071,20 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_InferenceRequestAddInput(
...
+/// the input will be deduced from model configuration, this function must at
+/// most be called once on request with no other input to ensure the deduction
+/// is accurate.
+///
+/// \param inference_request The request object.
+/// \param name The name of the input. This name is only used as a reference
+/// of the raw input in other Tritonserver APIs, it doesn't assoicate with the
```
dyastremsky, line 1081: Nit: Would be helpful for readability to break this into two sentences. "This name is used as a reference
of the raw input in other Tritonserver APIs. It is not associated with the name used in the model."

include/triton/core/tritonserver.h
GuanLuo: Can you elaborate?

deadeyegoodwin (dismissed):

src/infer_request.cc
```diff
@@ -682,6 +714,72 @@ InferenceRequest::Normalize()
 {
   const inference::ModelConfig& model_config = model_raw_->Config();
 
+  // Fill metadata for raw input
+  if (!raw_input_name_.empty()) {
+    if ((original_inputs_.size() != 1) || (model_config.input_size() != 1)) {
+      return Status(
+          Status::Code::INVALID_ARG,
+          "Raw request must only have 1 input to be deduced but got " +
+              std::to_string(model_config.input_size()) +
```
dyastremsky, line 723: Maybe should add the original_inputs_.size() to error message? Since the if statement includes both conditions, it's possible the model config has 1 input but the original inputs do not.

include/triton/core/tritonserver.h
CoderHam: The statement starts with model input name ... then mentions shape of input again. 

My suggestion was to remove the second occurrence of input

CoderHam (changes requested):

src/infer_request.cc
```diff
@@ -682,6 +714,72 @@ InferenceRequest::Normalize()
 {
   const inference::ModelConfig& model_config = model_raw_->Config();
 
+  // Fill metadata for raw input
+  if (!raw_input_name_.empty()) {
+    if ((original_inputs_.size() != 1) || (model_config.input_size() != 1)) {
```
nv-kmcgill53, line 719: I would suggest creating a local boolean before this if-statement. Something like,
```
...
if (!raw_input_name_.empty()) {
    bool input_size_not_one = (original_inputs_.size() != 1) || (model_config.input_size() != 1);
    if (input_size_not_one) {
...
```
Feel free to reject my naming convention in favor of your own. 

The reason for this is that when glancing over the code in the future, you don't want to have the overhead of having to reason out an if-statement before moving on in your train of thought. During debugging, this makes inspection of the logic easier as you can put a breakpoint directly on the temporary var. During a release build this temp var _should_ get optimized out by the compiler so there is no performance overhead either.

src/infer_request.cc
```diff
@@ -682,6 +714,72 @@ InferenceRequest::Normalize()
...
+      raw_input_size_ = raw_input.Data()->TotalByteSize();
+      RETURN_IF_ERROR(raw_input.PrependData(
+          &raw_input_size_, sizeof(uint32_t), TRITONSERVER_MEMORY_CPU, 0));
+      // Limit the BYTE raw input not to have host policy specific input for
+      // simplicity, such case won't happen given the current protocol spec.
+      // Will need to extend Input::PrependData() if needed.
+      if (!raw_input.HostPolicyData().empty()) {
+        return Status(
+            Status::Code::INVALID_ARG,
+            "Raw input with host policy specific data is not currently "
```
dyastremsky, line 773: Nit: You may want to remove "specific" to make this clearer. Unless it adds additional meaning.*

*If you want to leave specific in, it's a bit ambiguous. The unambiguous way to write it would be host-policy-specific data, though that's a bit long.

dyastremsky (changes requested):
Looks good! Added a few minor comments.

include/triton/core/tritonserver.h
tanmayv25: Something like: "The name, data type and shape of the model input will be deduced from model configuration."

GuanLuo: Maybe "the name recognized by the model" then, "model input name" is used to distinguish from the use of  "name" below

CoderHam (approved):

dyastremsky (approved):

CoderHam (approved):

PR #46, opened 2022-03-21
Generic Metrics API

rmccorm4:
- [x] Test build variant with metrics disabled
- [x] Create unit test to validate new APIs
```
root@f862e5554bf3:/mnt/triton/jira/3284-metrics-api/core/build# LD_LIBRARY_PATH=$PWD/triton-core:$LD_LIBRARY_PATH ./triton-core/test/metrics_api_test 
[==========] Running 2 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 2 tests from MetricsApiTest
[ RUN      ] MetricsApiTest.TestCounterEndToEnd
# HELP api_counter_example this is an example counter metric added via API.
# TYPE api_counter_example counter
api_counter_example{example1="counter_label1",example2="counter_label2"} 10.000000

[       OK ] MetricsApiTest.TestCounterEndToEnd (2 ms)
[ RUN      ] MetricsApiTest.TestGaugeEndToEnd
# HELP api_gauge_example this is an example gauge metric added via API.
# TYPE api_gauge_example gauge
api_gauge_example{example1="gauge_label1",example2="gauge_label2"} 42.000000

[       OK ] MetricsApiTest.TestGaugeEndToEnd (1 ms)
[----------] 2 tests from MetricsApiTest (3 ms total)

[----------] Global test environment tear-down
[==========] 2 tests from 1 test suite ran. (3 ms total)
[  PASSED  ] 2 tests.
```
- [x] Test calling actual `/metrics` endpoint equivalent (`Metrics::SerializedMetrics`) to view custom metrics
- [ ] Documentation - Separate PR to come with an example

**NOTE**

End-to-end testing from a sample backend will come with a separate PR, but does not necessarily have to block this PR given the unit tests added here.


rmccorm4:
End-to-end testing from a sample backend will come with a separate PR, but does not necessarily have to block this PR given the unit tests added here.

include/triton/core/tritonserver.h
```diff
@@ -89,7 +91,7 @@ struct TRITONSERVER_ServerOptions;
 ///   }
 ///
 #define TRITONSERVER_API_VERSION_MAJOR 1
-#define TRITONSERVER_API_VERSION_MINOR 10
+#define TRITONSERVER_API_VERSION_MINOR 11
```
rmccorm4, line 94: Please verify if this requires a version bump or not

tanmayv25: Yes. Any change in this API should come with a version bump.

src/tritonserver.cc
```diff
@@ -2731,12 +2731,17 @@ TRITONSERVER_ServerInferAsync(
...
+    const char* name, const char* description, void* registry_ptr)
 {
 #ifdef TRITON_ENABLE_METRICS
-  const auto& registry = tc::Metrics::GetRegistry();
+  const auto& registry =
+      reinterpret_cast<std::shared_ptr<prometheus::Registry>*>(registry_ptr);
+  // TODO: Remove this after review
+  // TODO: Verify passing registry_ptr from API caller is OK.
+  //       Calling GetRegistry() here did not add the new metrics to registry.
+  // const auto& registry = tc::Metrics::GetRegistry();
```
rmccorm4, line 2742: Need another set of eyes on this. 

Originally I retrieved the Registry object in the API implementation here:
```
const auto& registry = tc::Metrics::GetRegistry();
*family = reinterpret_cast<TRITONSERVER_MetricFamily*>(
      new tc::MetricFamily(kind, name, description, registry))
```

But the new metrics did not show up when collecting from the registry.

However, by passing the registry object from the API caller (unit test in this case), the new metrics were correctly added to the registry when calling `triton::core::Metrics::SerializedMetrics();`

include/triton/core/tritonserver.h
```diff
@@ -2141,6 +2143,98 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerInferAsync(
...
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_MetricValue(
+    TRITONSERVER_Metric* metric, double* value);
+
+/// Increase the current value of metric by value.
+///
+/// \param metric The metric object to update.
+/// \param value The amount to increment the metric's value by.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_MetricIncrement(
```
deadeyegoodwin, line 2211: Increment/decrement only work for COUNTER? Set only works for GAUGE? If so document that and also the error code that is returned if called on incorrect kind metric.

deadeyegoodwin (changes requested):

include/triton/core/tritonserver.h
```diff
@@ -2141,6 +2143,98 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerInferAsync(
...
+/// \param kind The TRITONSERVER_MetricKind type of MetricFamily to create.
+/// \param name The name of the metric family seen when calling the metrics
+/// endpoint.
+/// \param description The description of the metric family seen when
+/// calling the metrics endpoint.
+/// \param kind Pointer to the metric registry. // TODO
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_MetricFamilyNew(
+    TRITONSERVER_MetricFamily** family, const TRITONSERVER_MetricKind kind,
+    const char* name, const char* description, void* registry_ptr);
```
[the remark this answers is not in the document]
rmccorm4: I agree, but I think I have a misunderstanding in how the static resource works, as getting the registry object in the API implementation did not end up capturing the new metrics. Would appreciate any ideas - see this comment: https://github.com/triton-inference-server/core/pull/46#discussion_r833841249

src/metric_family.cc
```diff
@@ -0,0 +1,231 @@
...
+    case TRITONSERVER_METRIC_KIND_GAUGE: {
+      auto gauge_ptr = reinterpret_cast<prometheus::Gauge*>(metric_);
+      gauge_ptr->Decrement(value);
+      return nullptr;  // Success
+    }
+    default:
+      return TRITONSERVER_ErrorNew(
+          TRITONSERVER_ERROR_UNSUPPORTED,
+          "Unsupported TRITONSERVER_MetricKind");
+  }
```
tanmayv25, line 205: Have a single return nullptr at the end and replace it with break in KIND_GAUGE case.

src/tritonserver.cc
```diff
@@ -2725,4 +2725,132 @@ TRITONSERVER_ServerInferAsync(
...
+{
+#ifdef TRITON_ENABLE_METRICS
+  const auto& registry =
+      reinterpret_cast<std::shared_ptr<prometheus::Registry>*>(registry_ptr);
+  // TODO: Remove this after review
+  // TODO: Verify passing registry_ptr from API caller is OK.
+  //       Calling GetRegistry() here did not add the new metrics to registry.
+  // const auto& registry = tc::Metrics::GetRegistry();
+  *family = reinterpret_cast<TRITONSERVER_MetricFamily*>(
+      new tc::MetricFamily(kind, name, description, *registry));
```
tanmayv25, line 2744: Use try-catch in conjunction with my above comments to return errors when the MetricFamilu creation failed.

src/tritonserver.cc
```diff
@@ -2725,4 +2725,132 @@ TRITONSERVER_ServerInferAsync(
...
+    TRITONSERVER_Metric** metric, TRITONSERVER_MetricFamily* family,
+    const TRITONSERVER_Parameter** labels, const uint64_t label_count)
+{
+#ifdef TRITON_ENABLE_METRICS
+  std::vector<const tc::InferenceParameter*> labels_vec;
+  for (size_t i = 0; i < label_count; i++) {
+    labels_vec.emplace_back(
+        reinterpret_cast<const tc::InferenceParameter*>(labels[i]));
+  }
+  *metric = reinterpret_cast<TRITONSERVER_Metric*>(
```
tanmayv25, line 2778: try-catch here as well? Or may be not as it won't fail for a valid family with a kind.

src/metric_family.cc
```diff
@@ -0,0 +1,231 @@
...
+      gauge_ptr->Set(value);
+      return nullptr;  // Success
+    }
+    default:
+      return TRITONSERVER_ErrorNew(
+          TRITONSERVER_ERROR_UNSUPPORTED,
+          "Unsupported TRITONSERVER_MetricKind");
+  }
+}
+
```
tanmayv25, line 228: same as above

tanmayv25 (changes requested):

include/triton/core/tritonserver.h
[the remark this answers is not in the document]
deadeyegoodwin: Did you enable the metrics? As is done by the server here: https://github.com/triton-inference-server/core/blob/main/src/tritonserver.cc#L1959-L1965

src/tritonserver.cc
rmccorm4: Resolved by using C APIs to get metrics like `TRITONSERVER_ServerMetrics` and `TRITONSERVER_MetricsFormatted` instead of `tc::Metrics::SerializedMetrics()` in the unit test

include/triton/core/tritonserver.h
```diff
@@ -2141,6 +2143,107 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerInferAsync(
...
+    TRITONSERVER_Metric* metric, double value);
+
+/// Decrease the current value of metric by value.
+/// Supports metrics of kind TRITONSERVER_METRIC_KIND_GAUGE and returns
+/// TRITONSERVER_ERROR_UNSUPPORTED for unsupported TRITONSERVER_MetricKind.
+///
+/// \param metric The metric object to update.
+/// \param value The amount to decrement the metric's value by.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_MetricDecrement(
```
deadeyegoodwin, line 2226: How is Decrement different than Increment with a negative value? Do we really need Decrement?

src/metric_family.cc
```diff
@@ -0,0 +1,241 @@
+// Copyright (c) 2022, NVIDIA CORPORATION. All rights reserved.
```
GuanLuo, line 1: `NVIDIA CORPORATION & AFFILIATES`

src/test/CMakeLists.txt
```diff
@@ -424,3 +426,61 @@ install(
...
+
+
+if(${TRITON_ENABLE_METRICS})
+  #
+  # Unit test for Generic Metrics API
+  #
+  add_executable(
+    metrics_api_test
+    metrics_api_test.cc
+    ../metric_family.cc
```
GuanLuo, line 438: If the unit test only interact with Triton core via C APIs, I don't think you need to include these files

include/triton/core/tritonserver.h
rmccorm4: Good point, I will consolidate them.

include/triton/core/tritonserver.h
```diff
@@ -2141,6 +2143,98 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerInferAsync(
...
+/// Increment the current value of metric by value.
+/// Supports metrics of kind TRITONSERVER_METRIC_KIND_GAUGE for any value,
+/// and TRITONSERVER_METRIC_KIND_COUNTER for non-negative values. Returns
+/// TRITONSERVER_ERROR_UNSUPPORTED for unsupported TRITONSERVER_MetricKind
+/// or for negative values on a TRITONSERVER_METRIC_KIND_COUNTER metric.
+///
+/// \param metric The metric object to update.
+/// \param value The amount to increment the metric's value by.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_MetricIncrement(
```
rmccorm4, line 2217: @deadeyegoodwin there are some conflicting dictionary definitions of Increment for positive/negative values, though the mathematical definition doesn't require positive-ness. 

If you have any preferred name like Adjust/Change/etc. for clarity, let me know. I updated the documentation above the header to be more clear.

GuanLuo (dismissed):
We should have a ticket to track upgrading to newer release of prometheus-cpp

src/metric_family.cc
```diff
@@ -0,0 +1,233 @@
...
+      auto gauge_ptr = reinterpret_cast<prometheus::Gauge*>(metric_);
+      // Gauge::Increment works for both positive and negative values as of
+      // prometheus-cpp v1.0 but for now on v0.7 we defer call to
+      // Increment/Decrement based on the sign of value
+      // https://github.com/jupp0r/prometheus-cpp/blob/master/core/src/gauge.cc
+      if (value < 0.0) {
+        gauge_ptr->Decrement(-1.0 * value);
+      } else {
+        gauge_ptr->Increment(value);
+      }
```
rmccorm4, line 196: @deadeyegoodwin If we ever upgrade our prometheus-cpp dependency, this could be a few lines simpler and simply call `gauge_ptr->Increment(value)` regardless of sign. But I don't think we need to upgrade and risk potential changes to save 3 lines of code here.

rmccorm4: Similarly we could probably be fancy and do something like `gauge_ptr->Set(gauge_ptr->Value() + value);` (regardless of sign) here even with the current prometheus-cpp version, but I think that's less clear to read.

include/triton/core/tritonserver.h
```diff
@@ -2141,6 +2143,98 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerInferAsync(
...
+/// \param value Returns the current value of the metric object.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_MetricValue(
+    TRITONSERVER_Metric* metric, double* value);
+
+/// Increment the current value of metric by value.
+/// Supports metrics of kind TRITONSERVER_METRIC_KIND_GAUGE for any value,
+/// and TRITONSERVER_METRIC_KIND_COUNTER for non-negative values. Returns
+/// TRITONSERVER_ERROR_UNSUPPORTED for unsupported TRITONSERVER_MetricKind
+/// or for negative values on a TRITONSERVER_METRIC_KIND_COUNTER metric.
```
deadeyegoodwin, line 2212: Looks like you actually return INVALID_VALUE for a negative value with COUNTER

deadeyegoodwin (changes requested):

include/triton/core/tritonserver.h
rmccorm4: fixed

tanmayv25 (approved):

deadeyegoodwin (approved):

rmccorm4:
✔️ Pipeline #4238580

PR #48, opened 2022-03-21
Remove redefinition

GuanLuo:
The macros have been defined in [server_message.h](https://github.com/triton-inference-server/core/blob/main/src/server_message.h#L31-L35)


deadeyegoodwin (approved):

GuanLuo:
Addressed in https://github.com/triton-inference-server/core/pull/44

PR #49, opened 2022-03-23
Avoid deadlock between concurrent model load/unload and inference req…

tanmayv25:
…uests

The order of mutex acquisition is reverse for LoadUnload and consumer threads(InferHandler, InferStatistics).

The LoadUnload thread was getting stuck at acquiring map_mtx_ here: https://github.com/triton-inference-server/core/blob/28c2b892765b497d7fcc855f8fca6ea2e9a8dd60/src/model_repository_manager.cc#L830

### DIFF
--- a/src/model_repository_manager.cc
+++ b/src/model_repository_manager.cc
@@ -35,9 +35,9 @@
 #include "constants.h"
 #include "ensemble_utils.h"
 #include "filesystem.h"
-#include "triton/common/logging.h"
 #include "model.h"
 #include "model_config_utils.h"
+#include "triton/common/logging.h"
 #include "triton_repo_agent.h"
 
 #ifdef TRITON_ENABLE_GPU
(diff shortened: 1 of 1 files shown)


GuanLuo (approved):

CoderHam (approved):

Tabrizian (approved):
Just trying to understand this change, the issue is that another thread uses `map_mtx_` and the progress in this function depends on the progress in LoadUnload and we'll retry again later so that LoadUnload can make progress and release the mutex?

tanmayv25:
> Just trying to understand this change, the issue is that another thread uses map_mtx_ and the progress in this function depends on the progress in LoadUnload and we'll retry again later so that LoadUnload can make progress and release the mutex? 

That is correct.

> Can the LoadUnload model take a long time to release the lock? If yes, does it make sense to use a timed_mutex to not keep the CPU busy? If yes, does it make sense to use a timed_mutex to not keep the CPU busy?

The logic is in critical path of running the inference. We would like to release map_mtx_ as soon as possible for the LoadUnload thread to progress, hence I wanted it to be a non-blocking call here. The LoadUnload thread might be processing a specific model version at a time, so there is a high chance that user might have multiple model versions already loaded and in READY state. Upon the retries, my thought is that we might be able to capture any stable model version.

Tabrizian:
Thanks for clarifying.

PR #51, opened 2022-03-24
Support optional ensemble input

GuanLuo:


src/ensemble_scheduler.cc
```diff
@@ -472,6 +472,16 @@ EnsembleContext::EnsembleContext(
...
       }
     }
+
+    // Iterate the ensemble optional inputs and add empty tensor data entry
+    // if the input is not provided
+    for (const auto& name : info_->optional_inputs_) {
+      auto it = tensor_data_.find(name);
+      if ((it != tensor_data_.end()) && it->second.tensor_.empty()) {
+        it->second.AddTensor(nullptr);
+        it->second.batch_size_ = lrequest->BatchSize();
```
tanmayv25, line 482: Why are we required to set this batch_size?

GuanLuo: There is a ensemble feature that you can connect tensor between batching model / non-batching model as long as the full shapes match and this is used to perform the conversion

tanmayv25: Thanks! Should be document this?

tanmayv25 (approved):

tanmayv25 (approved):

PR #53, opened 2022-03-31
Create API to Register/Unregister Model Repository

dyastremsky:
Allow users to register and unregister model repositories via the C API.
Maintain a model mapping between directories and model names to make it possible for users to run models where the model repository and model have different names.

### DIFF
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -1887,11 +1887,12 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerStop(
 ///
 /// \param server The inference server object.
 /// \param repository_path The full path to the model repository.
-/// \param name_mapping List of name_mapping parameters. Each path mapping has
+/// \param name_mapping List of name_mapping parameters. Each mapping has
 /// the model directory name as its key, overriden model name as its value.
-/// These map each directory to overwritten model names. This is for cases where
-/// the directory name does not match the model. \param model_count Number of
-/// mappings provided \return a TRITONSERVER_Error indicating success or
+/// If the model directory name is empty, there must only be one model directory
+/// in the repository. In that case, the name will be set to that model directory.
+/// \param model_count Number of mappings provided.
+/// \return a TRITONSERVER_Error indicating success or
 /// failure.
 TRITONSERVER_DECLSPEC TRITONSERVER_Error*
 TRITONSERVER_ServerRegisterModelRepository(
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -1883,6 +1883,30 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerDelete(
 TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerStop(
     TRITONSERVER_Server* server);
 
+/// Register a new model repository
+///
+/// \param server The inference server object.
+/// \param repository_path The full path to the model repository.
+/// \param name_mapping List of name_mapping parameters. Each path mapping has
+/// the model directory name as its key, overriden model name as its value.
+/// These map each directory to overwritten model names. This is for cases where
+/// the directory name does not match the model. \param model_count Number of
+/// mappings provided \return a TRITONSERVER_Error indicating success or
+/// failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error*
+TRITONSERVER_ServerRegisterModelRepository(
+    TRITONSERVER_Server* server, const char* repository_path,
+    const TRITONSERVER_Parameter** name_mapping, const uint32_t mapping_count);
+
+/// Unregister a model repository
+///
+/// \param server The inference server object.
+/// \param repository_path The full path to the model repository.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error*
+TRITONSERVER_ServerUnregisterModelRepository(
+    TRITONSERVER_Server* server, const char* repository_path);
+
 /// Check the model repository for changes and update server state
 /// based on those changes.
 ///
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -1889,10 +1889,8 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerStop(
 /// \param repository_path The full path to the model repository.
 /// \param name_mapping List of name_mapping parameters. Each mapping has
 /// the model directory name as its key, overriden model name as its value.
-/// If the model directory name is empty, there must only be one model directory
-/// in the repository. In that case, the name will be set to that model
-/// directory. \param model_count Number of mappings provided. \return a
-/// TRITONSERVER_Error indicating success or failure.
+/// \param model_count Number of mappings provided.
+/// \return a TRITONSERVER_Error indicating success or failure.
 TRITONSERVER_DECLSPEC TRITONSERVER_Error*
 TRITONSERVER_ServerRegisterModelRepository(
     TRITONSERVER_Server* server, const char* repository_path,
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -1883,7 +1883,7 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerDelete(
 TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerStop(
     TRITONSERVER_Server* server);
 
-/// Register a new model repository
+/// Register a new model repository. Not available in polling mode.
 ///
 /// \param server The inference server object.
 /// \param repository_path The full path to the model repository.
@@ -1896,7 +1896,7 @@ TRITONSERVER_ServerRegisterModelRepository(
     TRITONSERVER_Server* server, const char* repository_path,
     const TRITONSERVER_Parameter** name_mapping, const uint32_t mapping_count);
 
-/// Unregister a model repository
+/// Unregister a model repository. Not available in polling mode.
 ///
 /// \param server The inference server object.
 /// \param repository_path The full path to the model repository.
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -1890,10 +1890,9 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerStop(
 /// \param name_mapping List of name_mapping parameters. Each mapping has
 /// the model directory name as its key, overriden model name as its value.
 /// If the model directory name is empty, there must only be one model directory
-/// in the repository. In that case, the name will be set to that model directory.
-/// \param model_count Number of mappings provided.
-/// \return a TRITONSERVER_Error indicating success or
-/// failure.
+/// in the repository. In that case, the name will be set to that model
+/// directory. \param model_count Number of mappings provided. \return a
+/// TRITONSERVER_Error indicating success or failure.
 TRITONSERVER_DECLSPEC TRITONSERVER_Error*
 TRITONSERVER_ServerRegisterModelRepository(
     TRITONSERVER_Server* server, const char* repository_path,
--- a/src/backend_model.h
+++ b/src/backend_model.h
@@ -47,7 +47,7 @@ class TritonModelInstance;
 class TritonModel : public Model {
  public:
   static Status Create(
-      InferenceServer* server, const std::string& model_repository_path,
+      InferenceServer* server, const std::string& original_model_path,
       const BackendCmdlineConfigMap& backend_cmdline_config_map,
       const HostPolicyCmdlineConfigMap& host_policy_map,
       const std::string& model_name, const int64_t version,
--- a/src/backend_model.h
+++ b/src/backend_model.h
@@ -47,7 +47,7 @@ class TritonModelInstance;
 class TritonModel : public Model {
  public:
   static Status Create(
-      InferenceServer* server, const std::string& original_model_path,
+      InferenceServer* server, const std::string& model_path,
       const BackendCmdlineConfigMap& backend_cmdline_config_map,
       const HostPolicyCmdlineConfigMap& host_policy_map,
       const std::string& model_name, const int64_t version,
--- a/src/model_repository_manager.cc
+++ b/src/model_repository_manager.cc
@@ -1163,8 +1163,8 @@ ModelRepositoryManager::ModelLifeCycle::CreateModel(
 #ifdef TRITON_ENABLE_ENSEMBLE
     if (model_info->is_ensemble_) {
       status = EnsembleModel::Create(
-          server_, model_path, version, model_config, min_compute_capability_,
-          &is);
+          server_, model_info->model_path_, version, model_config,
+          min_compute_capability_, &is);
       // Complete label provider with label information from involved models
       // Must be done here because involved models may not be able to
       // obtained from server because this may happen during server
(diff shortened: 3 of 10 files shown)


include/triton/core/tritonserver.h
```diff
@@ -1883,6 +1883,30 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerDelete(
...
     TRITONSERVER_Server* server);
 
+/// Register a new model repository
+///
+/// \param server The inference server object.
+/// \param repository_path The full path to the model repository.
+/// \param name_mapping List of name_mapping parameters. Each path mapping has
+/// the model directory name as its key, overriden model name as its value.
+/// These map each directory to overwritten model names. This is for cases where
+/// the directory name does not match the model. \param model_count Number of
```
GuanLuo, line 1893: Add comment about the special case where directory name is not provided (unknown) and user want the model inside has an overwritten name. This is the case we actually need to support.

GuanLuo: And newline for \param and \return

GuanLuo (commented):
@deadeyegoodwin for viz.

include/triton/core/tritonserver.h
```diff
@@ -1883,6 +1883,31 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerDelete(
 TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerStop(
     TRITONSERVER_Server* server);
 
+/// Register a new model repository
+///
+/// \param server The inference server object.
+/// \param repository_path The full path to the model repository.
+/// \param name_mapping List of name_mapping parameters. Each mapping has
+/// the model directory name as its key, overriden model name as its value.
+/// If the model directory name is empty, there must only be one model directory
```
deadeyegoodwin, line 1892: If the map doesn't contain an entry for a subdir or the model repo, then the model name used is just the name of the subdir... that is how our current model repository works. So the map can be empty or can list only some of the model subdirs... but there is no requirement around having only 1 model subdirectory.

deadeyegoodwin:
@dyastremsky  The formatting changes are obscuring your actual changes. Can you format everything in a separate branch/PR and then base this PR on that branch so we can see just the relevant changes.

deadeyegoodwin (changes requested):

include/triton/core/tritonserver.h
GuanLuo: I think for SageMaker, the endpoint doesn't know the file structure under the repository to be added, so they don't know what the subdir name is while the endpoint still want to use different name for the model.

dyastremsky: I can try to word this better. This is specifically for the case we are trying to support. If the user does not know the name of the model repository nor the name of the model, they can still use the model with triton.

They would register a model repository with that single model. The parameter would have an empty directory name a given model name to map it to. Then, they could use that model name with Triton (e.g. to load, run, or unload it).

deadeyegoodwin: It is not that the user does not know the name of the model. In the case of sagemaker they will be registering model repositories that have only a single model, and the subdir name for that model will always be 'model'. But the SM load-model API allows the user to assign a name to the model and that name is what triton should use. So for SM they will always call the register API with map { 'model' -> '<name user specifies in SM load model API>' }. But we should not need to do anything special for SM... they will just use the generic capability we are designing into the register API.

GuanLuo:
FYI, need to revisit the `ModelRepositoryManager::RepositoryIndex` as well ([model index protocol](https://github.com/triton-inference-server/server/blob/main/docs/protocol/extension_model_repository.md#index-1) for reference), this function iterates all monitored repository and currently use directory name for model name. I think it also need to handle the case of overridden model name

dyastremsky:
> FYI, need to revisit the `ModelRepositoryManager::RepositoryIndex` as well ([model index protocol](https://github.com/triton-inference-server/server/blob/main/docs/protocol/extension_model_repository.md#index-1) for reference), this function iterates all monitored repository and currently use directory name for model name. I think it also need to handle the case of overridden model name. Can be addressed in a separate ticket / PR if it requires non-trivial amount of effort

Thanks, Guan. Updated that function.

dyastremsky:
This feature is ready for review. Tests are passing locally.

src/backend_model.h
```diff
@@ -47,7 +47,7 @@ class TritonModelInstance;
 class TritonModel : public Model {
  public:
   static Status Create(
-      InferenceServer* server, const std::string& model_repository_path,
+      InferenceServer* server, const std::string& original_model_path,
```
GuanLuo, line 50: Just call `model_path`

src/model_repository_manager.cc
```diff
@@ -359,7 +360,7 @@ class ModelRepositoryManager::ModelLifeCycle {
   // not specified in the load will be unloaded after the load is finished.
   Status AsyncLoad(
       const std::string& repository_path, const std::string& model_name,
```
GuanLuo, line 362: Is `repository_path` still being used? I don't think we need to keep it given `model_path`

src/model_repository_manager.cc
```diff
@@ -1803,27 +1818,57 @@ ModelRepositoryManager::Poll(
...
-        } else if (exists_in_this_repo) {
-          auto res = model_to_repository.emplace(model.first, repository_path);
-          if (res.second) {
-            exists = true;
-          } else {
-            exists = false;
-            model_to_repository.erase(res.first);
-            LOG_ERROR << "failed to poll model '" << model.first
-                      << "': not unique across all model repositories";
+        }
```
GuanLuo, line 1835: I don't think you need the duplication check here, since `model_mappings_` ensures the model name is mapped to one directory and we decided not to poll other repos if we have found a mapping, duplication can't happen.

However, you should have `FileExists()` check as below to make sure the mapping is valid, i.e. user may delete the model directory after registering the repo.

src/model_repository_manager.cc
```diff
@@ -1803,27 +1818,57 @@ ModelRepositoryManager::Poll(
...
+            bool mapped = false;
+            for (auto const& mapping : model_mappings_) {
+              if (mapping.second.second == full_path) {
+                mapped = true;
+                break;
+              }
+            }
+            if (mapped) {
+              continue;
+            }
```
GuanLuo, line 1858: Shouldn't need this, if control reaches this point there shouldn't be an existing mapping. `else` in line 1836 is equivalent to `model_mappings_.find(model.first) == model_mappings_.end()`

src/model_repository_manager.cc
```diff
@@ -1902,7 +1953,8 @@ ModelRepositoryManager::Poll(
     }
 
     if (status.IsOk() && (model_poll_state != STATE_UNMODIFIED)) {
-      model_info.reset(new ModelInfo(mtime_ns, prev_mtime_ns, repository));
+      model_info.reset(
+          new ModelInfo(mtime_ns, prev_mtime_ns, repository, full_path));
```
GuanLuo, line 1957: The same here, is `repository` still needed? We may just store `full_path` in `model_to_repository` (then this variable name should be updated)?

src/model_repository_manager.cc
```diff
@@ -2162,6 +2220,67 @@ ModelRepositoryManager::UpdateDependencyGraph(
   return Status::Success;
 }
 
+Status
+ModelRepositoryManager::AddModelRepositoryPath(const std::string& path)
+{
+  if (!repository_paths_.insert(path).second) {
```
GuanLuo, line 2226: Need mutex to protect concurrent calls to the function here and below`std::lock_guard<std::mutex> lock(poll_mu_);`

src/model_repository_manager.h
```diff
@@ -320,10 +344,19 @@ class ModelRepositoryManager {
...
   bool CheckNode(DependencyNode* node);
 
+  /// Get the model mappings.
+  /// \return Map with model name as key, pair of repository path and full path
+  // as value.
+  const std::unordered_map<std::string, std::pair<std::string, std::string>>&
+  ModelMappings() const
+  {
+    return model_mappings_;
+  }
```
GuanLuo, line 354: Doesn't seem to be used anywhere, can remove this

src/test/register_api_test.cc
```diff
@@ -0,0 +1,733 @@
...
+  FAIL_TEST_IF_ERR(
+      TRITONSERVER_ServerRegisterModelRepository(
+          server_, "models_1", name_map.data(), name_map.size()),
+      "registering model repository 'models_1'");
+
+  // Request to load "model_1"
+  FAIL_TEST_IF_ERR(
+      TRITONSERVER_ServerLoadModel(server_, "model_1"),
+      "loading model 'model_1'");
+  // [FIXME] check if the correct model is loaded (directory "model_0")
```
GuanLuo, line 270: I will push more changes to address the FIXME in this test and some other tests

src/tritonserver.cc
```diff
@@ -2157,6 +2157,112 @@ TRITONSERVER_ServerStop(TRITONSERVER_Server* server)
...
+      default:
+        mode = "UNKNOWN";
+        break;
+    }
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_UNSUPPORTED,
+        (std::string("Unregister API is unsupported in ") + mode +
+         " model control mode")
+            .c_str());
+  }
```
GuanLuo, line 2260: The control mode check should be moved to `ModelRepositoryManger`, similar to the check in `ModelRepositoryManager::LoadUnloadModel`, we can update the test to expect error message like `repository register is not allowed if model control mode is EXPLICIT` for simplicity.

src/tritonserver.cc
```diff
@@ -2157,6 +2157,112 @@ TRITONSERVER_ServerStop(TRITONSERVER_Server* server)
...
+  if (!status.IsOk() || !is_directory) {
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_INVALID_ARG,
+        (std::string("failed to register '") + repository_path +
+         "', repository not found")
+            .c_str());
+  }
+
+  RETURN_IF_STATUS_ERROR(lserver->AddModelRepositoryPath(repository_path));
+  std::unordered_map<std::string, std::string> model_mapping;
```
GuanLuo, line 2201: Why not just make one function `InferenceServer::AddModelRepositoryPath(const std::string& repository, std::unordered_map<std::string, std::string> model_mapping);`, then the implementation here is just translating Triton objects to C++ objects and propagating it downwards. Where the validations will be done in model manager.

In this way, you don't need to worry about unregistering the repository if error is returned

GuanLuo (changes requested):

src/test/register_api_test.cc
GuanLuo: @dyastremsky FYI, I pushed to this branch directly

include/triton/core/tritonserver.h
dyastremsky: Thanks for clarifying. Removed this edge case based on our conversation.

src/backend_model.h
dyastremsky: That would cause a conflict with a local variable. https://github.com/triton-inference-server/core/blob/388e9eb970eacff0214496aae72ecd077b316f28/src/backend_model.cc#L107

model_path represents the model path used, whether local or original. It seems like the call to LocalizeDirectory() sets these differently for different filesystems, so I thought keeping the nomenclature original_path made sense. If you'd prefer something else, let me know.

src/model_repository_manager.cc
dyastremsky: This is an edge case, revealed by one of the tests (DifferentMappings, I believe). Say we have two repositories: models_0 with model_0 mapped to "name_0," and models_1 with model_0 and model_1 (no mappings). If you don't add this check, model_0 from models_0 could be loaded, yet that's counter to the behavior the user requested by mapping that subdirectory to a different name ("name_0").

GuanLuo: I see, resolved

src/backend_model.h
GuanLuo: I see, then I am okay with that. Or you may change the local variable `model_path` to `localized_model_path`.

src/model_repository_manager.cc
dyastremsky: Updated.

src/model_repository_manager.h
dyastremsky: Done.

src/tritonserver.cc
dyastremsky: Done.

src/tritonserver.cc
dyastremsky: Done.

src/model_repository_manager.cc
dyastremsky: Good catch, updated.

src/backend_model.h
dyastremsky: Updated.

src/model_repository_manager.cc
dyastremsky: Thanks, removed repository_path throughout for loading (and ModelInfo). Looks like there was a FIXME in there to get rid of it.

src/model_repository_manager.cc
dyastremsky: Done.

src/model_repository_manager.cc
```diff
@@ -2162,6 +2204,100 @@ ModelRepositoryManager::UpdateDependencyGraph(
...
+
+    std::set<std::string> models_to_delete;
+    for (auto const& mapping : model_mappings_) {
+      if (mapping.second.first == repository) {
+        models_to_delete.insert(mapping.first);
+      }
+    }
+    for (auto const& model : models_to_delete) {
+      if (model_mappings_.erase(model) != 1) {
+        all_found = false;
```
GuanLuo, line 2287: Actually this is not likely to happen, `models_to_delete` is generated based on the item in `model_mappings_`

src/model_repository_manager.cc
```diff
@@ -1703,14 +1700,28 @@ ModelRepositoryManager::RepositoryIndex(
   std::set<std::string> seen_models;
   std::set<std::string> duplicate_models;
   for (const auto& repository_path : repository_paths_) {
+    // For any mapped models in this repository, save the mapping
+    // from their subdirectory name to model name.
+    std::map<std::string, std::string> models_in_repo;
+    for (const auto& mapping_it : model_mappings_) {
+      if (mapping_it.second.first == repository_path) {
+        models_in_repo.emplace(mapping_it.second.first, repository_path);
```
GuanLuo, line 1708: I think this should be `models_in_repo.emplace(BaseName(mapping_it.second.second), mapping_it.first);` (`key->directory_name`; `value->override_name`). In line 1715, you want to search if the directory name is a key of `models_in_repo`, and set `model` to the corresponding value.

src/model_repository_manager.cc
```diff
@@ -2162,6 +2204,100 @@ ModelRepositoryManager::UpdateDependencyGraph(
...
+      }
+    }
+  }
+
+  LOG_INFO << "Model repository unregistered: " << repository;
+  if (!all_found) {
+    return Status(
+        Status::Code::INTERNAL,
+        "Not all registered model mappings found for '" + repository + "'");
+  }
```
GuanLuo, line 2297: This is also not likely to happen

GuanLuo (changes requested):
Can you also create a follow-up ticket to add testing on checking model index in the case of override model name?

src/model_repository_manager.cc
dyastremsky: Removed. I may have been over-rigorous with the error checking.

src/model_repository_manager.cc
dyastremsky: Thanks for catching this. So many changes to the mapping design haha.

As per your comment above, I made a follow-up ticket to test the model indexes for mapped models.

src/model_repository_manager.cc
dyastremsky: Removed.

CoderHam (approved):

GuanLuo (approved):

deadeyegoodwin (approved):

nskool:
Does `tritonserver_stub.cc` need to change as well?

I get the following error during linking unless the function declarations are added to tritonserver_stub.cc as well.
```
libhttp-endpoint-library.a(sagemaker_server.cc.o): In function `triton::server::SagemakerAPIServer::ParseRequest(evhtp_request*)':
sagemaker_server.cc:(.text+0x26f6): undefined reference to `TRITONSERVER_ServerRegisterModelRepository'
```

dyastremsky:
Thanks for catching that, @nskool. Created a follow-up PR to fix this quickly.

dyastremsky:
@nskool I can't tag you on the other pull request. Let me know if the fix is working for your build.

nskool:
Thanks @dyastremsky - yes, the fix works for my build.

dyastremsky:
Perfect, thanks for verifying, @nskool!

PR #55, opened 2022-04-01
Add sequence-aware shutdown

GuanLuo:


dyastremsky:
Would you be able to rebase this on main? The main branch should be formatted now, so it'll make your changes easier to see.

src/model_repository_manager.h
```diff
@@ -170,8 +170,20 @@ class ModelRepositoryManager {
...
+  /// Instruct all models to stop accepting new inference requests. However,
+  /// the models are still capable of processing inference requests
+  /// if the model considers them as part of the in-flight inference.
+  /// \return error status.
+  Status StopAllModels();
+
+  /// \return the number of in-flight inferences for the all versions of all
+  /// models. Note that a model version will not be included if it doesn't
+  /// have in-flight inferences.
+  const std::set<std::tuple<std::string, int64_t, size_t>> InflightStatus();
```
dyastremsky, line 185: Could be useful to note what the first and second values are for the tuples

src/model_repository_manager.cc
```diff
@@ -519,6 +526,50 @@ ModelRepositoryManager::ModelLifeCycle::LiveModelStates(bool strict_readiness)
   return live_model_states;
 }
 
+Status
+ModelRepositoryManager::ModelLifeCycle::StopAllModels()
+{
+  LOG_VERBOSE(1) << "StopAllModels()";
```
Tabrizian, line 532: Is it a debug statement?

src/model_repository_manager.cc
```diff
@@ -519,6 +526,50 @@ ModelRepositoryManager::ModelLifeCycle::LiveModelStates(bool strict_readiness)
...
+      }
+    }
+  }
+  return Status::Success;
+}
+
+const std::set<std::tuple<std::string, int64_t, size_t>>
+ModelRepositoryManager::ModelLifeCycle::InflightStatus()
+{
+  LOG_VERBOSE(1) << "InflightStatus()";
```
Tabrizian, line 551: Same

src/server.cc
```diff
@@ -250,33 +250,59 @@ InferenceServer::Stop(const bool force)
...
-                         << ModelReadyStateString(v.second.first);
+    if (!unloading_model) {
+      // Check if all in-flight inference requests / sequences are completed
+      const auto& inflight_status = model_repository_manager_->InflightStatus();
+      LOG_INFO << "Timeout " << exit_timeout_iters << ": Found "
+               << inflight_status.size()
+               << " model versions that have in-flight inferences";
+      if (LOG_VERBOSE_IS_ON(1)) {
+        for (const auto& inflight : inflight_status) {
+          LOG_VERBOSE(1) << std::get<0>(inflight) << " v"
```
Tabrizian, line 271: Do we want to include more information when logging the inflight requests here? For example, "Model std::get<0>(inflight) (version std::get<1>(inflight)) has std::get<2>(inflight) in-flight inference request(s)"

Tabrizian (commented):
I was also curious whether this will also help BLS with more graceful shutdown when there are inflight inference requests? When I was using Perf Analyzer's C-API mode, I noticed that when Perf Analyzer was exiting, BLS returned an error indicating that the model that BLS was performed on is no longer ready.

src/dynamic_batch_scheduler.cc
```diff
@@ -163,6 +163,11 @@ DynamicBatchScheduler::~DynamicBatchScheduler()
 Status
 DynamicBatchScheduler::Enqueue(std::unique_ptr<InferenceRequest>& request)
 {
+  if (stop_) {
+    return Status(
+        Status::Code::UNAVAILABLE,
+        "Scheduler has stopped accepting new inference");
```
deadeyegoodwin, line 169: Perhaps say something indicating why... "Server is stopping, scheduler for model <xyz> has stopped accepting new inference requests:

src/ensemble_scheduler.cc
```diff
@@ -1250,6 +1250,12 @@ EnsembleScheduler::Create(
 Status
 EnsembleScheduler::Enqueue(std::unique_ptr<InferenceRequest>& request)
 {
+  if (stop_) {
+    return Status(
+        Status::Code::UNAVAILABLE,
+        "Scheduler has stopped accepting new inference");
```
deadeyegoodwin, line 1256: See above comment...

src/sequence_batch_scheduler.cc
```diff
@@ -546,6 +546,15 @@ SequenceBatchScheduler::Enqueue(std::unique_ptr<InferenceRequest>& irequest)
   const bool seq_end =
       ((irequest->Flags() & TRITONSERVER_REQUEST_FLAG_SEQUENCE_END) != 0);
 
+  // Check if the request is one of the in-flight sequence (not starting new
+  // sequence), currently we consider sequences in backlog are also in-flight
```
deadeyegoodwin, line 550: "sequence). We consider sequences in backlog as also in-flight."

Not just for simplicity but also because we want that semantic. Any sequence that client has been able to start in the server we want to allow to finish.

src/sequence_batch_scheduler.cc
```diff
@@ -546,6 +546,15 @@ SequenceBatchScheduler::Enqueue(std::unique_ptr<InferenceRequest>& irequest)
...
+  // for simplicity.
+  if (stop_ && seq_start) {
+    return Status(
+        Status::Code::UNAVAILABLE,
+        "Scheduler has stopped accepting new inference");
```
deadeyegoodwin, line 555: Fix as above...

deadeyegoodwin (changes requested):

src/model_repository_manager.cc
GuanLuo: Yes, I think I should put it to VERBOSE(2)

src/sequence_batch_scheduler.h
```diff
@@ -127,6 +137,8 @@ class SequenceBatchScheduler : public Scheduler {
   // The max_sequence_idle_microseconds value for this scheduler.
   uint64_t max_sequence_idle_microseconds_;
 
+  bool stop_;
```
tanmayv25, line 140: Shouldn't this stop_ be a part of scheduler base class? Also, all scheduler seems to be setting this flag to true. So, shouldn't there be a single implementation of Stop function in scheduler base class?

src/model_repository_manager.cc
```diff
@@ -1607,6 +1658,18 @@ ModelRepositoryManager::UnloadAllModels()
   return Status::Success;
 }
 
+Status
+ModelRepositoryManager::StopAllModels()
+{
+  return model_life_cycle_->StopAllModels();
+}
+
+const std::set<std::tuple<std::string, int64_t, size_t>>
```
tanmayv25, line 1667: The set can be made more descriptive to make it more comprehensible.

tanmayv25 (changes requested):

src/sequence_batch_scheduler.h
GuanLuo: I have thought about it, the Scheduler base class was designed to be just an interface so I keep it this way: pure virtual can make the implementer to think twice about what the scheduler should behave in those cases. And now the `Ensemble::Stop()` will just be no-op as discussed.

tanmayv25: Yes. This can be resolved.

src/model_repository_manager.cc
GuanLuo: Added comment about the return tuple in the function description, not sure if we need to `using ..` for it as it is not super nested that needs a descriptive name.

GuanLuo:
@Tabrizian I don't think this will improve the BLS case, like the ensemble, it will try to send request to the inner model and the inner model will still reject the request (but with different error message). To allow BLS to complete the current request, the inner models will need to be aware of "who" is sending the request and not rejecting the request.

Tabrizian (approved):

tanmayv25 (approved):

deadeyegoodwin (approved):

PR #57, opened 2022-04-06
Add basic tests

GuanLuo:
@dyastremsky this is some of the test cases for the register API, it is not completed yet but you should be able to (cherry-pick /) build it and use for some testing. I didn't run the actual test but it should define the expected behavior, feel free to suggest changes to the test case to better reflect the implementation and to have proper expected behavior.

### DIFF
--- a/src/test/register_api_test.cc
+++ b/src/test/register_api_test.cc
@@ -1,0 +1,374 @@
+// Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions
+// are met:
+//  * Redistributions of source code must retain the above copyright
+//    notice, this list of conditions and the following disclaimer.
+//  * Redistributions in binary form must reproduce the above copyright
+//    notice, this list of conditions and the following disclaimer in the
+//    documentation and/or other materials provided with the distribution.
+//  * Neither the name of NVIDIA CORPORATION nor the names of its
+//    contributors may be used to endorse or promote products derived
+//    from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+// PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+// OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include <iostream>
+#include <thread>
+#include "gtest/gtest.h"
+#include "triton/core/tritonserver.h"
+
+namespace {
+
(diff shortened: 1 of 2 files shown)


src/test/register_api_test.cc
```diff
@@ -0,0 +1,374 @@
...
+TEST_F(RegisterApiTest, RegisterWithRepeatedMap)
+{
+  // Registering a repository "models_1" where contains "model_0" and "model_1",
+  // map "model_0" to "model_1" which creates confliction, however,
+  // in EXPLICIT mode, mapping lookup will have higher priority than
+  // repository polling so the confliction will be resolved by always loading
+  // the model from mapped directory.
+  // [FIXME] should this be expected behavior? Should this be the same for
+  // POLLING mode? Or it should be treated as having dupliciate models since
+  // model manager always iterate the repositories in POLLING.
```
GuanLuo, line 219: @deadeyegoodwin @dyastremsky What do you think about the above behavior?

deadeyegoodwin: How about we not allow the new Register/Unregister repo API with POLLING mode. Mode must be either NONE or EXPLICIT or the API calls should fail.

src/test/register_api_test.cc
```diff
@@ -0,0 +1,374 @@
...
+          server_, "models_0", nullptr, 0),
+      TRITONSERVER_ERROR_ALREADY_EXISTS,
+      "model repository 'models_0' has been registered",
+      "registering model repository 'models_0'");
+}
+
+TEST_F(RegisterApiTest, RegisterWithMultiMap)
+{
+  // Registering a repository "models_0" where contains "model_0",
+  // and "model_0" is mapped to two different names
```
dyastremsky, line 180: @GuanLuo @deadeyegoodwin Is this a case we want/need to support?

I can change the mapping values to hold a list of model names, but I feel that it'd introduce complexity for an edge case that seems to have very limited utility or use. Do users need the ability to map the same subdirectory to multiple model names?

deadeyegoodwin: I agree. Let's enforce that there can be only a single name for any model. There needs to be some testing around this.

src/test/register_api_test.cc
dyastremsky: Right now, I believe the behavior is that model lookup happens at the time of polling in any mode. So there is no priority. Is the priority desirable for us?

I hadn't considered the introduction of conflict from different namings during polling. Instead of taking the first one we find (as was the behavior before), I can report an error if multiple models with same name are found. That would mean no priority, e.g. RegisterMultiWithMap would fail due to reporting an error for duplicate names upon loading rather than loading based on priority.

We can get rid of the feature in POLLING mode if we want, but I think the current behavior plus the above change will get us what we're looking for. If I'm understanding the above comments, this exists in EXPLICIT mode as well, since you may have a model already loaded then register a repository with a duplicate model name, then try to register that model again.

dyastremsky: Also, this feature would do nothing in NONE mode, right? It's only registering the model, but you can't load it in that mode.

dyastremsky:
Thanks for being so thorough with these test cases, Guan. The code is working (and passing CI currently, with the pipeline almost done). I'll have to make a few changes based on the good tests you wrote (see comments/discussion), and then build your tests and get them passing. Feel free to run against the current PR's changes to get a feel for what happens.

dyastremsky:
Merging PR for ease, as Guan updates/adds tests. Feel free to continue the conversations above.

src/test/register_api_test.cc
GuanLuo: I think user may want to register / unregister with POLLING mode, but we can disallow it until further requested.

The "priority" introduced here is to avoid polling many repositories when a model name has been mentioned in the mapping. Currently in any mode, we poll all repositories to perform duplication check, this can be slow if many repositories are registered, but with "priority", then we can say if the model name is found in the mapping, we will load the model based on the mapping and don't bother to check if there is a model directory with the same name (we know which directory to be used even if there is duplication).

The above also explains why I wasn't sure about the "priority' in POLLING mode, as we already poll all repository anyway. We may want this for consistency but we will need additional handling when we detects duplication. And disallowing the register API in POLLING can defer the resolution of this concern.

dyastremsky: You're talking about the following lines, right?

Polling: https://github.com/triton-inference-server/core/blob/cd4d42781cf081f307241d73fdd64b1c07d5b3b4/src/model_repository_manager.cc#L1790-L1816
Non-polling: https://github.com/triton-inference-server/core/blob/cd4d42781cf081f307241d73fdd64b1c07d5b3b4/src/model_repository_manager.cc#L1834-L1848

For polling, I still do the model lookups at the same time. For each subdirectory, if a mapping exists in that subdirectory, it is used; else, the default behavior.

For specific models (EXPLICIT/NONE), the default behavior applies as well except for a check for that mapping (if exists). It sounds like the issue is that a model may be loaded when duplicate models are registered. I don't know that removing POLLING would impact that. We either need to do a duplicate check on loading (poll all model repositories) or load the first model found with that name, which would be random based on whichever repository the map has saved first. The latter is what the code would do now.

PR #63, opened 2022-04-22
Merge PersistentBackendManager into TritonBackendManager

deadeyegoodwin:


src/backend_manager.cc
```diff
@@ -315,11 +279,26 @@ TRITONBACKEND_BackendSetState(TRITONBACKEND_Backend* backend, void* state)
 //
 // TritonBackendManager
 //
-TritonBackendManager&
-TritonBackendManager::Singleton()
+
+static std::weak_ptr<TritonBackendManager> backend_manager_;
```
nv-kmcgill53, line 283: Can you explain this pattern please? I'm not sure I understand why there needs to be a global weak_ptr to the backend_manager when the server is the one holding ownership of this object. I'm likely missing something here.

nv-kmcgill53: My understanding of this is the following:
- On startup the server creates a new backend_manager where both the weak_ptr and the server's shared_ptr are null.
- On line 293 we ask the weak_ptr for the ptr but it's null so it returns null
- We generate a new TritonBackendManager on like 298 for the server's shared_ptr

At this point if we were to call TritonBackendManager::Create again, we can just check this server's manager and skip the check of the weak_ptr. I guess another question I have is do we have a use case where we create multiple InferenceServer objects or do we want to leave ourselves open to that? If yes, then the weak_ptr is needed; otherwise, I don't think it is.

deadeyegoodwin: We do support the creation of multiple InferenceServer objects in the same process, though we don't use that capability ourselves in the tritonserver executable and it is unlikely that anyone else uses it either. So this pattern is making sure we have a shared BackendManager across all those InferenceServer objects, but that we don't extend the lifetime of that shared BackendManager longer than what is required by the InferenceServer objects.

nv-kmcgill53: Thanks for the explanation!

nv-kmcgill53 (approved):

GuanLuo (approved):

PR #60, opened 2022-04-15
Move model_config to common

kpedro88:
Addresses https://github.com/triton-inference-server/server/issues/2759: many of the functions in `model_config.h` are generically useful even on the client side and only depend on the generated protobuf headers. Therefore, it is useful to have them more broadly accessible via the common repo.

A few functions that have non-trivial dependence on the server API or on `constants.h` are kept in the core repo, moved to `model_config_utils`. One definition from `constants.h` is moved to the common repo.

The associated changes in the `model_config.h` header location and the namespace are propagated throughout the core repo.

I checked that these changes compiled successfully by following https://github.com/triton-inference-server/server/blob/main/docs/build.md to get the Docker container and install all needed system dependencies, then built the core repo following https://github.com/triton-inference-server/core#build-the-triton-core-library.

This PR goes with https://github.com/triton-inference-server/common/pull/48.

PR #71, opened 2022-05-09
Move model_config to common

jbkyang-nvi:
Creating new PR because @deadeyegoodwin is out for a bit and we want it to get it into 22.05. 
[Old PR](https://github.com/triton-inference-server/core/pull/60)
Addresses https://github.com/triton-inference-server/server/issues/2759: 
many of the functions in model_config.h are generically useful even on the client side and only 
depend on the generated protobuf headers. Therefore, it is useful to have them more broadly 
accessible via the common repo.

A few functions that have non-trivial dependence on the server API or on constants.h are kept in
 the core repo, moved to model_config_utils. One definition from constants.h is moved to the common repo.

The associated changes in the model_config.h header location and the namespace are propagated
 throughout the core repo.

This PR goes with https://github.com/triton-inference-server/common/pull/48.


kpedro88 (changes requested):
I didn't notice this the other day, but I think 9123adc10bcc70925e971ba8ea4539bd1c9199cd did the wrong thing. `persistent_backend_manager.h/cc` now show up in this PR as totally new files because they were removed in #63. I think they should also be removed here...

kpedro88 (dismissed):

GuanLuo (dismissed):
Check and update copyright year if needed

kpedro88:
@GuanLuo all files touched in this PR should now have copyright year 2022.

GuanLuo (dismissed):

GuanLuo (approved):

PR #73, opened 2022-05-12
Fix metric lifecycle

GuanLuo:


src/metric_family.h
```diff
@@ -45,8 +48,13 @@ class MetricFamily {
   void* Family() const { return family_; }
   TRITONSERVER_MetricKind Kind() const { return kind_; }
 
+  void* Add(std::map<std::string, std::string> label_map);
+  void Remove(void* metric);
+
  private:
   void* family_;
+  std::mutex mtx_;
+  std::unordered_map<void*, size_t> metric_ref_cnt_;
```
rmccorm4, line 57: Can you add a comment right here or in `.cc` with your ref_cnt description since prometheus returns the same ptr for dupe labels?

rmccorm4 (changes requested):

src/metric_family.h
rmccorm4: Otherwise, LGTM

rmccorm4 (approved):

tanmayv25 (approved):

PR #77, opened 2022-05-17
Return warmup request error during model loading

GuanLuo:


src/backend_model_instance.cc
```diff
@@ -547,36 +561,62 @@ TritonModelInstance::Initialize()
...
+        response_complete(sample.requests_.size());
+
+    std::vector<TRITONBACKEND_Request*> triton_requests;
+    for (size_t i = 0; i < sample.requests_.size(); ++i) {
+      auto& request = sample.requests_[i];
+      request->SetReleaseCallback(WarmupRequestComplete, &request_complete[i]);
+      response_complete[i].second = &response_errors;
+      request->SetResponseCallback(
+          &warmup_allocator, nullptr, WarmupResponseComplete,
+          &response_complete[i]);
```
rmccorm4, line 582: Can you write a comment here summarizing this process? 

> ... set response callback for each warmup request and pass response_complete userp so we can check for errors from each async response and exit early if any warmup request fails ...

> ... use promises to signal when each response is complete async ... etc.

This part of the code is new to me so these kinds of summaries are very helpful when reading it later.

Otherwise, the WIP LGTM so far

rmccorm4 (changes requested):

Tabrizian (commented):
Do we want to add some testing for decoupled models using warmup feature too?

GuanLuo:
@tanmayv25 Do you think if this [commit](https://github.com/triton-inference-server/core/pull/77/commits/3c1893102080d6b4eced389e713789603b08e252) makes sense? I am thinking about warmup for sequence model where the batch will be formed by batch-1 requests in regular inference. This is the same before the commit, after the commit the control inputs are "concatenated" to a single input from the backend's perspective.

I think it is okay though as the backend doesn't examine the control input of the requests anyway, it treats them like regular input: put them into single buffer and send to the model.

src/backend_model_instance.cc
```diff
@@ -409,107 +418,106 @@ TritonModelInstance::GenerateWarmupData()
...
+
+      const char* allocated_ptr = nullptr;
+      switch (input_meta.second.input_data_type_case()) {
+        case inference::ModelWarmup_Input::InputDataTypeCase::kZeroData:
+          allocated_ptr = zero_buffer;
+          break;
+        case inference::ModelWarmup_Input::InputDataTypeCase::kRandomData: {
+          if (input_meta.second.data_type() ==
+              inference::DataType::TYPE_STRING) {
             allocated_ptr = zero_buffer;
```
rmccorm4, line 449: Why is a string treated differently?

GuanLuo: Added comment

rmccorm4 (commented):
LGTM, though I can't confidently comment on correctness of this comment: https://github.com/triton-inference-server/core/pull/77#issuecomment-1129238067

I'll leave that to @tanmayv25

tanmayv25:
Some of the backends may be using control inputs to identify/manage state within the backend. I would prefer if we don't merge the control inputs.

GuanLuo:
Reverted single request change

rmccorm4 (approved):

tanmayv25 (approved):

PR #79, opened 2022-05-17
Default-max-batch-size changed to 4

nv-kmcgill53:
Default-max-batch-size is now 2 as a follow up to the findings found in DLIS-3571

### DIFF
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -91,7 +91,7 @@ struct TRITONBACKEND_ModelInstance;
 ///   }
 ///
 #define TRITONBACKEND_API_VERSION_MAJOR 1
-#define TRITONBACKEND_API_VERSION_MINOR 9
+#define TRITONBACKEND_API_VERSION_MINOR 10
 
 /// Get the TRITONBACKEND API version supported by Triton. This value
 /// can be compared against the TRITONBACKEND_API_VERSION_MAJOR and
@@ -974,12 +974,12 @@ TRITONBACKEND_ModelAutoCompleteConfig(
     TRITONBACKEND_Model* model, bool* auto_complete_config);
 
 /// Set the model configuration in Triton server. Only the inputs, outputs,
-/// and max batch size can be changed. Any other changes to the model
-/// configuration will be ignored by Triton. This function can only be called
-/// from TRITONBACKEND_ModelInitialize, calling in any other context will result
-/// in an error being returned. The function does not take ownership of the
-/// message object and so the caller should call TRITONSERVER_MessageDelete to
-/// release the object once the function returns.
+/// max batch size, and dynamic batcher can be changed. Any other changes to the
+/// model configuration will be ignored by Triton. This function can only be
+/// called from TRITONBACKEND_ModelInitialize, calling in any other context will
+/// result in an error being returned. The function does not take ownership of
+/// the message object and so the caller should call TRITONSERVER_MessageDelete
+/// to release the object once the function returns.
 ///
 /// \param model The model.
 /// \param config_version The format version of the model configuration.
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -974,12 +974,13 @@ TRITONBACKEND_ModelAutoCompleteConfig(
     TRITONBACKEND_Model* model, bool* auto_complete_config);
 
 /// Set the model configuration in Triton server. Only the inputs, outputs,
-/// max batch size, and scheduling choice can be changed. Any other changes to the
-/// model configuration will be ignored by Triton. This function can only be
-/// called from TRITONBACKEND_ModelInitialize, calling in any other context will
-/// result in an error being returned. The function does not take ownership of
-/// the message object and so the caller should call TRITONSERVER_MessageDelete
-/// to release the object once the function returns.
+/// max batch size, and scheduling choice can be changed. A caveat being
+/// scheduling choice can only be changed if none is previously set. Any other
+/// changes to the model configuration will be ignored by Triton. This function
+/// can only be called from TRITONBACKEND_ModelInitialize, calling in any other
+/// context will result in an error being returned. The function does not take
+/// ownership of the message object and so the caller should call
+/// TRITONSERVER_MessageDelete to release the object once the function returns.
 ///
 /// \param model The model.
 /// \param config_version The format version of the model configuration.
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -974,7 +974,7 @@ TRITONBACKEND_ModelAutoCompleteConfig(
     TRITONBACKEND_Model* model, bool* auto_complete_config);
 
 /// Set the model configuration in Triton server. Only the inputs, outputs,
-/// max batch size, and dynamic batcher can be changed. Any other changes to the
+/// max batch size, and scheduling choice can be changed. Any other changes to the
 /// model configuration will be ignored by Triton. This function can only be
 /// called from TRITONBACKEND_ModelInitialize, calling in any other context will
 /// result in an error being returned. The function does not take ownership of
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -330,6 +330,9 @@ TritonModel::UpdateModelConfig(
   auto outputs_config = config.mutable_output();
   *outputs_config = updated_config.output();
 
+  auto dynamic_batching_config = config.mutable_dynamic_batching();
+  *dynamic_batching_config = updated_config.dynamic_batching();
+
   RETURN_IF_ERROR(SetModelConfig(config));
   return Status::Success;
 }
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -330,7 +330,7 @@ TritonModel::UpdateModelConfig(
   auto outputs_config = config.mutable_output();
   *outputs_config = updated_config.output();
 
-  if (!config.scheduling_choice_case() ) {
+  if (!config.scheduling_choice_case()) {
     if (updated_config.has_dynamic_batching()) {
       auto dynamic_batching_config = config.mutable_dynamic_batching();
       *dynamic_batching_config = updated_config.dynamic_batching();
@@ -344,11 +344,17 @@ TritonModel::UpdateModelConfig(
       // updated_config is empty
       config.clear_scheduling_choice();
     }
-  } else if (config.scheduling_choice_case() != updated_config.scheduling_choice_case()){
-    return Status(triton::common::Error::Code::INTERNAL,
-      (std::string("Failed to overwrite scheduling choice, ") + std::to_string(updated_config.scheduling_choice_case()) + 
-      std::string(", from updated config to config, ") + std::to_string(config.scheduling_choice_case())).c_str());
-  } // else do nothing
+  } else if (
+      config.scheduling_choice_case() !=
+      updated_config.scheduling_choice_case()) {
+    return Status(
+        triton::common::Error::Code::INTERNAL,
+        (std::string("Failed to overwrite scheduling choice, ") +
+         std::to_string(updated_config.scheduling_choice_case()) +
+         std::string(", from updated config to config, ") +
+         std::to_string(config.scheduling_choice_case()))
+            .c_str());
+  }  // else do nothing
 
   RETURN_IF_ERROR(SetModelConfig(config));
   return Status::Success;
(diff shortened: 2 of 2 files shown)


tanmayv25 (approved):

krishung5 (approved):

Tabrizian (approved):

Tabrizian:
@nv-kmcgill53 I briefly looked at the ticket but I couldn't find the summary report of Model Analyzer. Does max_batch_size of four lead to decreased performance for some models?

nv-kmcgill53:
Yes, from what I observed, the batch size of 4 lead to decreased performance. I reported it in the comment section of the ticket and there is a file attached with the model summaries of the models I looked at.

GuanLuo:
I am looking at the report from DLIS-3571, I notice that it only capture the case when the request concurrency is 1? Any batch size won't take effect if there is only 1 request in any given moment, and I would expect the concurrency will be larger in real world setting

nv-kmcgill53:
@GuanLuo I will investigate these models with a larger concurrency. Do you know of a typical value our customers use when testing this?

Tabrizian:
@nv-kmcgill53 I would suggest running Model Analyzer using the default arguments and it will automatically sweep various max_batch_size and model instance values with appropriate concurrency values. We can make the final judgement based on those results.

tanmayv25:
@GuanLuo @Tabrizian  This is a default value of max_batch_size which will be populated by auto-complete logic. This is not expected to be the most performant but should definitely not cause regressions on other workloads. It is possible that the workload for this model is indeed a single request at a time say every minute.
For this reason, we don't have this value as a constant MACRO, but a a CLI option. If the user want to override this default value, then they can do so using --backend-config track. That is if they expect a lot of requests.
This is where the feature request of AutoTuner comes into picture. With this default, we are not claiming that it will be the most performant config.

tanmayv25:
This being said I don't understand why mbs 4 is less performant that mbs 2 for a single request concurrency and no queue wait time. :thinking:

nv-kmcgill53:
Letting the model analyzer run on default settings against a savedmodel_f32_f32_f32 model produced a preferred max batch size of 64 with 3 GPU instances. I ran this on an A6000 which, for all intents and purposes, is GPU processor bound. I agree with @tanmayv25 that everyone's mileage will vary depending on what hardware they are running and that this value is a reasonable default for ease of use but is no in do way claiming to be the most performant.

nv-kmcgill53:
After an offline discussion about the default-max-batch-size, we have decided to revert the change from 2 back to 4. This will be reflected in the next commit.

include/triton/core/tritonbackend.h
```diff
@@ -91,7 +91,7 @@ struct TRITONBACKEND_ModelInstance;
 ///   }
 ///
 #define TRITONBACKEND_API_VERSION_MAJOR 1
-#define TRITONBACKEND_API_VERSION_MINOR 9
+#define TRITONBACKEND_API_VERSION_MINOR 10
```
GuanLuo, line 94: Don't need to change version as there is no API changes

nv-kmcgill53: The API hasn't changed but the functionality has. This lets the user know that they may see different behavior when using this version.

src/backend_model.cc
```diff
@@ -330,6 +330,9 @@ TritonModel::UpdateModelConfig(
   auto outputs_config = config.mutable_output();
   *outputs_config = updated_config.output();
 
+  auto dynamic_batching_config = config.mutable_dynamic_batching();
```
GuanLuo, line 333: Need to be careful here, what if `sequence_batching` has been set for the model from the beginning, I think this will clear that because scheduler choice is "oneof" field. We probably should fail the config update, and on the backend side, the auto-complete should check if scheduler choice has been set and don't overwrite it.

nv-kmcgill53: Updated the method to have the ability to update all batching types when the current config is different from the updated config.

src/backend_model.cc
```diff
@@ -330,6 +330,22 @@ TritonModel::UpdateModelConfig(
...
+      auto sequence_batching_config = config.mutable_sequence_batching();
+      *sequence_batching_config = updated_config.sequence_batching();
+    } else if (updated_config.has_ensemble_scheduling()) {
+      auto ensemble_scheduling_config = config.mutable_ensemble_scheduling();
+      *ensemble_scheduling_config = updated_config.ensemble_scheduling();
+    } else {
+      // updated_config is empty
+      config.clear_scheduling_choice();
+    }
+  }
```
GuanLuo, line 347: The current check seems to be always overwriting the original scheduling choice if the updated config is different? Do we want to have more restriction regarding that? i.e. if the original config specifies sequence / ensemble scheduling, we probably don't want to overwrite it. Perhaps only update the scheduling choice if original config has no scheduling set, or has dynamic batching set?

nv-kmcgill53: `UpdateModelConfig` is only used by [`TRITONBACKEND_ModelSetConfig`](https://github.com/triton-inference-server/core/blob/1c60840e05eafcc3e0a7fa4d29a996fd7f76f1aa/src/backend_model.cc#L509-L520). `TRITONBACKEND_ModelSetConfig` is only used in the Autocomplete sections of the backends. As we have written the documentation in the header file, the incoming updated config should be overwriting the current config. Otherwise, we are cherry picking updates under certain conditions and that makes it unclear what this function is doing. 

Usage of this method as I see it is, a change to the model configuration happened and the calling code wants to save this new data; this method should do that. Let me know if my understanding is wrong here.

GuanLuo: > Usage of this method as I see it is, a change to the model configuration happened and the calling code wants to save this new data; this method should do that

I think this should be conditional. Although the TF/ORT backend change has handled it (not filling scheduling option if it has been set), the danger comes from when backend implementer (/ user) tries to have auto-complete support that always set dynamic batching when possible, without realizing that the model may have been configured as a sequence model. In this case the model will end up being served without using sequence batcher.

nv-kmcgill53: Updated the method to only update the scheduling choice if none is already set. Updated the API in `tritonbackend.h` to reflect this change.

src/backend_model.cc
```diff
@@ -330,6 +330,22 @@ TritonModel::UpdateModelConfig(
   auto outputs_config = config.mutable_output();
   *outputs_config = updated_config.output();
 
+  if (!config.scheduling_choice_case() ) {
```
Tabrizian, line 333: Do we need an else if clause to fail the auto-complete if both the original and updated config has a scheduling setting?

nv-kmcgill53: I don't believe we should fail here. It's up to each backend how they want to autocomplete the scheduling choice. The else-if in this case being the user has provided a scheduling choice in their model config and autocomplete believing it should be something different. 

An example might be, the user sets the sequence batching in their model config but autocomplete wants to set it to dynamic batching. Guan's point above is that we don't know if  autocomplete checked this scheduling choice before it sets it. Thus, only set the scheduling choice if none has been set and do nothing otherwise (prioritizing the user's config). 

Let me know if this was incoherent and needs more explanation.

Tabrizian: I was wondering may be we should remove the condition `!config.scheduling_choice_case()` altogether. If I'm understanding correctly, it looks like we will be silently not setting what the backend has provided to us as an auto-completed model configuration for scheduling section.

Similar to how we don't do any error checking for inputs/outputs, I think we should trust that the backend developer has properly auto-filled the scheduling section and we will use whatever setting they have provided as the final configuration. All the conflict detection between the original and the auto-completed model configuration should be handled in the backend side (similar to inputs and outputs).

nv-kmcgill53: This [idea was rejected](https://github.com/triton-inference-server/core/pull/79#discussion_r881008560) by @GuanLuo above with the following reason: 
> the danger comes from when backend implementer (/ user) tries to have auto-complete support that always set dynamic batching when possible, without realizing that the model may have been configured as a sequence model.

Tabrizian: I find it a bit strange that here we are neither returning an error nor setting what the backend has specified. I think maybe we can wait until Guan is back to make the final call.

GuanLuo: I think we can return error here if the backend tries to overwrite the scheduling choice

nv-kmcgill53: I may need a bit of clarification here. From this thread I understand the following:
- We will only set the scheduling choice if none is already set
- If the scheduling choice in `config` is different from the scheduling choice in `updated_config`, we will return an Error, presumably INTERNAL, with an appropriate error message.
- If the scheduling choice in `config` is the same as the scheduling choice in `updated_config` then skip this update.

Is this the correct understanding?

GuanLuo: Yes, that is my understanding as well, although I would use `INVALID_ARGS` instead of `INTERNAL` because the error returned in this case is intentional.

CC @tanmayv25 to see if below make sense: My thought process is that comparing to I/O where the backend will know better and can provide "correct" config by examining the model file, scheduling choice is different in the sense that if user specify it in model config, then the backend shouldn't overwrite it as the user "know better" on how the model should be served

GuanLuo (approved):

src/backend_model.cc
tanmayv25: I agree with current changes. We must through an error if the backend developer sets a scheduling_choice incorrectly.
The autocomplete logic in  backend must be mindful of handling whatever scheduling choice is specified in config.pbtxt and not try to change it. 
We have those checks in TF and ORT backends before we try adding dynamic_batching scheduling choice.

tanmayv25: I think it should still be INTERNAL because the backend developer has not done a good job of writing a robust auto-complete logic.

src/backend_model.cc
```diff
@@ -330,6 +330,32 @@ TritonModel::UpdateModelConfig(
...
+    } else {
+      // updated_config is empty
+      config.clear_scheduling_choice();
+    }
+  } else if (
+      config.scheduling_choice_case() !=
+      updated_config.scheduling_choice_case()) {
+    return Status(
+        triton::common::Error::Code::INTERNAL,
+        (std::string("Failed to overwrite scheduling choice, ") +
```
tanmayv25, line 352: Isn't the relation inverse? I would find this to be more clear.
```
Can not update scheduling choice from `config.scheduling_choice_case()` to `updated_config.scheduling_choice_case()` when auto-completing. 
```

src/backend_model.cc
```diff
@@ -330,6 +330,32 @@ TritonModel::UpdateModelConfig(
...
+      *dynamic_batching_config = updated_config.dynamic_batching();
+    } else if (updated_config.has_sequence_batching()) {
+      auto sequence_batching_config = config.mutable_sequence_batching();
+      *sequence_batching_config = updated_config.sequence_batching();
+    } else if (updated_config.has_ensemble_scheduling()) {
+      auto ensemble_scheduling_config = config.mutable_ensemble_scheduling();
+      *ensemble_scheduling_config = updated_config.ensemble_scheduling();
+    } else {
+      // updated_config is empty
+      config.clear_scheduling_choice();
```
tanmayv25, line 345: Don't clear the scheduling choice. Do nothing here.

src/backend_model.cc
nv-kmcgill53: Fixed to have more readable error message.

src/backend_model.cc
nv-kmcgill53: Took out clearing of scheduling choice and left comment defining the "do nothing" default behavior

src/backend_model.cc
```diff
@@ -330,6 +330,29 @@ TritonModel::UpdateModelConfig(
...
+      auto ensemble_scheduling_config = config.mutable_ensemble_scheduling();
+      *ensemble_scheduling_config = updated_config.ensemble_scheduling();
+    }  // else do nothing
+  } else if (
+      config.scheduling_choice_case() !=
+      updated_config.scheduling_choice_case()) {
+    return Status(
+        triton::common::Error::Code::INTERNAL,
+        (std::string("Cannot update scheduling choice from ") +
+         std::to_string(config.scheduling_choice_case()) + std::string(" to ") +
```
GuanLuo, line 350: I think protobuf provides method that returns readable string of the scheduling choice

GuanLuo: You are right, I was thinking about `enum` when I typed this, and I couldn't find the equivalent for `oneof` neither, resolving this discussion

GuanLuo (approved):

tanmayv25 (approved):

nv-kmcgill53:
Rebased on top of main

tanmayv25 (approved):

PR #82, opened 2022-06-03
Default CPU Instance to 2 

nv-kmcgill53:
Defaults instance groups of KIND_CPU to a count of 2 rather than 1 be default. Pytorch is excluded from this list as it doesn't benefit from this higher instance count. Other backends which need to be excluded can be added to the list.

### DIFF
--- a/src/model_config_utils.cc
+++ b/src/model_config_utils.cc
@@ -768,7 +768,7 @@ GetNormalizedModelConfig(
 
       // Count
       if (group.count() < 1) {
-        group.set_count(1);
+        RETURN_IF_ERROR(SetDefaultInstanceCount(&group, config->backend()));
       }
 
       // GPUs
@@ -782,6 +782,27 @@ GetNormalizedModelConfig(
   }
 
   return Status::Success;
+}
+
+Status
+SetDefaultInstanceCount(
+    ::inference::ModelInstanceGroup* group, std::string& backend)
+{
+  if (group->kind() == inference::ModelInstanceGroup::KIND_CPU) {
+    // Some backends don't perform well with multiple
+    // model instances such as pytorch so we want to
+    // default to a different value. Add other similar backends
+    // to this check.
+    if (backend.compare(kPyTorchBackend) == 0) {
+      group->set_count(1);
+    } else {
+      group->set_count(2)
+    }
+  } else {
+    group->set_count(1);
+  }
+
+  return nullptr;  // Success
 }
 
 Status
--- a/src/model_config_utils.cc
+++ b/src/model_config_utils.cc
@@ -791,12 +791,11 @@ SetDefaultInstanceCount(
   const int default_cpu_instance_count = 2;
   if (group->kind() == inference::ModelInstanceGroup::KIND_CPU) {
     // Backends opt into the default_cpu_instance_count since
-    // some backends don't perform well/have high overhead
+    // some backends (pytorch, OpenVINO) don't perform well/have high overhead
     // when using multiple instances.
-    bool use_default_cpu_instance_count =
-        (backend.compare(kTensorRTBackend) == 0) ||
-        (backend.compare(kTensorFlowBackend) == 0) ||
-        (backend.compare(kOnnxRuntimeBackend) == 0);
+    bool use_default_cpu_instance_count = (backend == kTensorRTBackend) ||
+                                          (backend == kTensorFlowBackend) ||
+                                          (backend == kOnnxRuntimeBackend);
     if (use_default_cpu_instance_count) {
       group->set_count(default_cpu_instance_count);
     } else {
--- a/src/model_config_utils.cc
+++ b/src/model_config_utils.cc
@@ -786,23 +786,20 @@ GetNormalizedModelConfig(
 
 Status
 SetDefaultInstanceCount(
-    ::inference::ModelInstanceGroup* group, const std::string& backend)
+    inference::ModelInstanceGroup* group, const std::string& backend)
 {
+  group->set_count(1);
+
+  // Backends opt into the default_cpu_instance_count since
+  // some backends (pytorch, OpenVINO) don't perform well/have high overhead
+  // when using multiple instances.
   const int default_cpu_instance_count = 2;
-  if (group->kind() == inference::ModelInstanceGroup::KIND_CPU) {
-    // Backends opt into the default_cpu_instance_count since
-    // some backends (pytorch, OpenVINO) don't perform well/have high overhead
-    // when using multiple instances.
-    bool use_default_cpu_instance_count = (backend == kTensorRTBackend) ||
-                                          (backend == kTensorFlowBackend) ||
-                                          (backend == kOnnxRuntimeBackend);
-    if (use_default_cpu_instance_count) {
-      group->set_count(default_cpu_instance_count);
-    } else {
-      group->set_count(1);
-    }
-  } else {
-    group->set_count(1);
+  bool use_default_cpu_instance_count = (backend == kTensorRTBackend) ||
+                                        (backend == kTensorFlowBackend) ||
+                                        (backend == kOnnxRuntimeBackend);
+  if (group->kind() == inference::ModelInstanceGroup::KIND_CPU &&
+      use_default_cpu_instance_count) {
+    group->set_count(default_cpu_instance_count);
   }
 
   return Status::Success;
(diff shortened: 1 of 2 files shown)


src/model_config_utils.cc
```diff
@@ -784,6 +784,28 @@ GetNormalizedModelConfig(
   return Status::Success;
 }
 
+Status
+SetDefaultInstanceCount(
+    ::inference::ModelInstanceGroup* group, const std::string& backend)
+{
+  const int default_cpu_instance_count = 2;
+  if (group->kind() == inference::ModelInstanceGroup::KIND_CPU) {
+    // Some backends don't perform well with multiple
```
Tabrizian, line 793: I think it might be better to only set the number of instances to two for backends we know than turn it on for all the backends. For Python backend, if the model is using Pytorch library, it would add ~600MB of GPU memory for each new model instance. Also, I think it is best to not add backend-specific code in the core repository. Perhaps we can add a backend API so that the backend can indicate the instance kind and instance count that it prefers if the user has not specified anything?

Tabrizian, line 793: I think it might be better to only set the number of instances to two for backends we know than turn it on for all the backends. For Python backend, if the model is using Pytorch library, it would add 600MB of GPU memory for each new model instance. Also, I think it is best to not add backend-specific code in the core repository. Perhaps we can add a backend API so that the backend can indicate the instance kind and instance count that it prefers if the user has not specified anything?

nv-kmcgill53: For now I have changed the list to opt in rather than opt out to show what it will look like.

Tabrizian: Thanks for the response. I just wanted to bring-up this change since it keeps coming up. This backend API change will both fix Python backend + jetson issue and remove the backend specific changes from core. I understand that it is easier to hardcode backend names for now. Just something to consider in the future.

Tabrizian (dismissed):
@nv-kmcgill53 We should also make sure that this change does not affect the nightly performance benchmarks that we collect for each backend. Need to make sure that all the benchmark models use instance count of 1 for the cases they require to do so.

nv-kmcgill53:
That's a good catch. I will make sure of that.

src/model_config_utils.cc
```diff
@@ -784,6 +784,31 @@ GetNormalizedModelConfig(
...
 }
 
+Status
+SetDefaultInstanceCount(
+    ::inference::ModelInstanceGroup* group, const std::string& backend)
+{
+  const int default_cpu_instance_count = 2;
+  if (group->kind() == inference::ModelInstanceGroup::KIND_CPU) {
+    // Backends opt into the default_cpu_instance_count since
+    // some backends don't perform well/have high overhead
```
GuanLuo, line 794: Can you add some examples? Looks like they are OpenVINO and Torch backend?

src/model_config_utils.cc
```diff
@@ -784,6 +784,31 @@ GetNormalizedModelConfig(
...
+    // when using multiple instances.
+    bool use_default_cpu_instance_count =
+        (backend.compare(kTensorRTBackend) == 0) ||
+        (backend.compare(kTensorFlowBackend) == 0) ||
+        (backend.compare(kOnnxRuntimeBackend) == 0);
```
GuanLuo, line 799: For `std::string`, you can use `==` directly

GuanLuo (approved):

src/model_config_utils.cc
nv-kmcgill53: Updated comment to use Pytorch and OpenVINO as examples of backends which don't perform well with multiple instances.

src/model_config_utils.cc
nv-kmcgill53: using overloaded operator instead of `compare()`

src/model_config_utils.cc
```diff
@@ -784,6 +784,30 @@ GetNormalizedModelConfig(
...
+                                          (backend == kTensorFlowBackend) ||
+                                          (backend == kOnnxRuntimeBackend);
+    if (use_default_cpu_instance_count) {
+      group->set_count(default_cpu_instance_count);
+    } else {
+      group->set_count(1);
+    }
+  } else {
+    group->set_count(1);
+  }
```
rmccorm4, line 806: optional: can this code be cleaned up at all by just setting group count to 1 at the beginning, and setting to 2 `if use_default_cpu_instance_count` ?

src/model_config_utils.h
```diff
@@ -86,6 +86,14 @@ Status GetNormalizedModelConfig(
     const std::string& model_name, const std::string& path,
     const double min_compute_capability, inference::ModelConfig* config);
 
+/// Auto-complete the instance count based on instance kind and backend name.
+/// For KIND_CPU the default instance count is 2 except for the pytorch backend.
+/// \param group The instance group to set the count for.
+/// \param backned The backend name to check against.
```
rmccorm4, line 92: backned -> backend

src/model_config_utils.h
```diff
@@ -86,6 +86,14 @@ Status GetNormalizedModelConfig(
     const std::string& model_name, const std::string& path,
     const double min_compute_capability, inference::ModelConfig* config);
 
+/// Auto-complete the instance count based on instance kind and backend name.
+/// For KIND_CPU the default instance count is 2 except for the pytorch backend.
```
rmccorm4, line 90: currently this says (except for the pytorch backend).

maybe either 
1. add full list of unsupported backends (pytorch, openvino, ...) 
2. or the converse (default is 2 for TRT/TF/ORT)
3. or just say something more generic (default instance count is 2 for certain backends that benefit from it) so this documentation makes sense even as we change the backends that default to 2 over time

src/model_config_utils.h
```diff
@@ -86,6 +86,14 @@ Status GetNormalizedModelConfig(
...
+/// \param group The instance group to set the count for.
+/// \param backned The backend name to check against.
+/// \return The error status.
+Status SetDefaultInstanceCount(
+    ::inference::ModelInstanceGroup* group, const std::string& backend);
```
rmccorm4, line 95: any specific reason for the leading `::` before `::inference` ? This isn't done in the rest of the file, so just curious.

rmccorm4 (changes requested):

src/model_config_utils.h
nv-kmcgill53: >For KIND_CPU the default instance count is 2 except for the pytorch backend.

Would you be ok with me taking this sentence out entirely? After the changes which have been made, I don't think this explanation makes much sense on the API

rmccorm4: I think it's OK to remove here if we document something about this in server docs: https://github.com/triton-inference-server/server/pull/4466#discussion_r897251607

nv-kmcgill53: Since the first sentence states that it auto-complete's on the kind and the backend, I'll update the docs (linked) to be more explicit about this feature.

src/model_config_utils.h
nv-kmcgill53: I saw it being done in other files so I thought this was the norm. Fixed.

src/model_config_utils.h
nv-kmcgill53: fixed

src/model_config_utils.cc
nv-kmcgill53: You are right, it can be cleaned up. Updated.

rmccorm4 (approved):

tanmayv25:
It looks fine for now. But we must think of how to configure these settings from backend.

src/model_config_utils.cc
```diff
@@ -784,6 +784,27 @@ GetNormalizedModelConfig(
...
+SetDefaultInstanceCount(
+    inference::ModelInstanceGroup* group, const std::string& backend)
+{
+  group->set_count(1);
+
+  // Backends opt into the default_cpu_instance_count since
+  // some backends (pytorch, OpenVINO) don't perform well/have high overhead
+  // when using multiple instances.
+  const int default_cpu_instance_count = 2;
+  bool use_default_cpu_instance_count = (backend == kTensorRTBackend) ||
```
tanmayv25, line 797: TensorRT backend runs only on GPU and can not run on CPU.
So, KIND can not be KIND_CPU for TensorRT.
It is fine removing that from these conditions.

tanmayv25 (changes requested):

src/model_config_utils.cc
nv-kmcgill53: Updated to take out TensorRT from the opt in list.

tanmayv25 (approved):

GuanLuo (approved):

PR #83, opened 2022-06-07
Use thread pool to load models

rmccorm4:
Fixes DLIS-1833 and removes need for 100ms sleep in model load thread due to [glibc bug](https://sourceware.org/bugzilla/show_bug.cgi?id=19329).

Corresponding server PR: https://github.com/triton-inference-server/server/pull/4474

### DIFF
--- a/src/model_repository_manager.cc
+++ b/src/model_repository_manager.cc
@@ -28,6 +28,8 @@
 #include "model_repository_manager.h"
 
 #include <algorithm>
+#include <boost/asio/post.hpp>
+#include <boost/asio/thread_pool.hpp>
 #include <deque>
 #include <future>
 #include <stdexcept>
@@ -39,10 +41,6 @@
 #include "model_config_utils.h"
 #include "repo_agent.h"
 #include "triton/common/logging.h"
-
-// Thread Pool
-#include <boost/asio/post.hpp>
-#include <boost/asio/thread_pool.hpp>
 
 #ifdef TRITON_ENABLE_GPU
 #include <cuda_runtime_api.h>
--- a/src/model_repository_manager.cc
+++ b/src/model_repository_manager.cc
@@ -557,6 +557,7 @@ class ModelRepositoryManager::ModelLifeCycle {
         cmdline_config_map_(backend_cmdline_config_map),
         host_policy_map_(host_policy_map)
   {
+    LOG_INFO << "Creating load_pool_ with thread_count: " << model_load_thread_count;
     load_pool_.reset(new boost::asio::thread_pool(model_load_thread_count));
   }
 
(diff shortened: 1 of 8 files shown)


src/model_repository_manager.cc
```diff
@@ -40,6 +40,10 @@
 #include "repo_agent.h"
 #include "triton/common/logging.h"
 
+// Thread Pool
+#include <boost/asio/post.hpp>
+#include <boost/asio/thread_pool.hpp>
```
nv-kmcgill53, line 45: A question I always ask when seeing new dependencies is: What do we get from this library which is worth the linking rather than writing the thread pool ourselves? It looks like we only need basic functionality out of this lib so I would argue it would be better to write the thread pool ourselves and reduce the code/dependency overhead in this case. 

CC: @tanmayv25 @GuanLuo

rmccorm4: Going off our use of boost for performant implementations in a few places, and similar [use of boost thread_pool in Python Backend](https://github.com/triton-inference-server/python_backend/blob/73fdcdad66f79064e5fb04b66fa30af375b6d0fa/src/python_be.h#L271), I thought this would be OK here rather than re-implementing something similar.

nv-kmcgill53 (dismissed):

src/model_repository_manager.cc
```diff
@@ -1191,18 +1203,10 @@ ModelRepositoryManager::ModelLifeCycle::Load(
...
-        std::this_thread::sleep_for(std::chrono::milliseconds(100));
-        std::thread worker(
-            &ModelRepositoryManager::ModelLifeCycle::CreateModel, this,
-            model_name, version, model_info);
-        worker.detach();
-      }
+      // Load model asynchronously via thread pool
+      boost::asio::post(*load_pool_, [this, model_name, version, model_info]() {
+        CreateModel(model_name, version, model_info);
+      });
```
Tabrizian, line 1209: I was wondering do we need to make sure that all the threads have exited before destructing this class? Especially since the callback is capturing this.

Tabrizian (changes requested):

src/model_repository_manager.cc
Tabrizian: @GuanLuo @tanmayv25 what do you think?

GuanLuo: I think the thread pool destructor will join and stop the working thread, however, we should destruct the thread pool as the first step of stopping model lifecycle as the callback can modify model lifecycle members (updating model status)

src/model_repository_manager.cc
```diff
@@ -546,11 +554,14 @@ class ModelRepositoryManager::ModelLifeCycle {
   ModelLifeCycle(
       const double min_compute_capability, InferenceServer* server,
       const triton::common::BackendCmdlineConfigMap& backend_cmdline_config_map,
-      const triton::common::HostPolicyCmdlineConfigMap& host_policy_map)
+      const triton::common::HostPolicyCmdlineConfigMap& host_policy_map,
+      const unsigned int model_load_thread_count)
```
GuanLuo, line 558: It may make sense to use a `option` struct to group the configurations. Can be done in separate change / ticket

rmccorm4:
See new PR using custom thread pool implementation instead of adding boost dependency: https://github.com/triton-inference-server/core/pull/86

src/model_repository_manager.cc
rmccorm4: Added DLIS-3866

PR #85, opened 2022-06-07
Prefix Logs with Request ID

dyastremsky:
Prefix logs with the request ID, where possible.

Related changes:
Backend: https://github.com/triton-inference-server/backend/pull/60
Server: https://github.com/triton-inference-server/server/pull/4484

### DIFF
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -91,7 +91,7 @@ struct TRITONBACKEND_ModelInstance;
 ///   }
 ///
 #define TRITONBACKEND_API_VERSION_MAJOR 1
-#define TRITONBACKEND_API_VERSION_MINOR 10
+#define TRITONBACKEND_API_VERSION_MINOR 11
 
 /// Get the TRITONBACKEND API version supported by Triton. This value
 /// can be compared against the TRITONBACKEND_API_VERSION_MAJOR and
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -372,6 +372,14 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_OutputBufferAttributes(
 TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_RequestId(
     TRITONBACKEND_Request* request, const char** id);
 
+/// Get the ID of the request as a string formatted for logging.
+///
+/// \param request The inference request.
+/// \param formatted_id Returns the formatted string of the ID.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_RequestIdString(
+    TRITONBACKEND_Request* request, const char** formatted_id);
+
 /// Get the correlation ID of the request if it is an unsigned integer.
 /// Zero indicates that the request does not have a correlation ID.
 /// Returns failure if correlation ID for given request is not an unsigned
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -91,7 +91,7 @@ struct TRITONBACKEND_ModelInstance;
 ///   }
 ///
 #define TRITONBACKEND_API_VERSION_MAJOR 1
-#define TRITONBACKEND_API_VERSION_MINOR 11
+#define TRITONBACKEND_API_VERSION_MINOR 10
 
 /// Get the TRITONBACKEND API version supported by Triton. This value
 /// can be compared against the TRITONBACKEND_API_VERSION_MAJOR and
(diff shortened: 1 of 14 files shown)


include/triton/core/tritonbackend.h
```diff
@@ -372,6 +372,14 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_OutputBufferAttributes(
 TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_RequestId(
     TRITONBACKEND_Request* request, const char** id);
 
+/// Get the ID of the request as a string formatted for logging.
+///
+/// \param request The inference request.
+/// \param formatted_id Returns the formatted string of the ID.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_RequestIdString(
+    TRITONBACKEND_Request* request, const char** formatted_id);
```
GuanLuo, line 381: Why not reusing `TRITONBACKEND_RequestId` above?

src/infer_request.cc
```diff
@@ -154,17 +154,17 @@ InferenceRequest::TraceInputTensors(
...
+                        ": " + msg + ": tensor: " + name +
+                        ": fail to get data buffer: " + status.Message());
         return status;
       }
 
       if (buffer_size != byte_size) {
         LOG_STATUS_ERROR(
             status,
-            std::string(TRITONSERVER_InferenceTraceActivityString(activity)) +
+            IdString() + TRITONSERVER_InferenceTraceActivityString(activity) +
```
GuanLuo, line 167: `IdString()` is too specific, if we are going to log more request info, the name will need to be changed. Can be something like `LogRequest()`

include/triton/core/tritonbackend.h
GuanLuo: I would like to think about it more before adding new public-facing API. For now, we can provide common utility to format the request ID returned from existing APIs.

dyastremsky: What do you mean by providing a common utility? Repositories other than core seem to use TRITONBACKEND_Request or TRITONSERVER_Request objects rather than InferenceRequest objects directly.

I'm happy to use an alternative approach. It seems like the two main options would be:

1. Create a function in the common repo just for formatting the request ID.
2. Replicate the formatting function in each repository we want to use it in (backend, server, and any backends we decide to add logging to). Right now, the formatting is happening in one place, so any changes to how we print request IDs or similar can be done by changing IdString() in InferenceRequest. This could could create redundancy and different formatting over time.

I think the current change avoids issues with those two approaches, but I get the desire to avoid updating the API for a logging ticket. Take your time on thinking about it, I can change the approach as needed.

rmccorm4: I think it would be better to have an internal helper (can be in `common` repo if the helper is used across multiple repos server/core/backend/etc.) for formatting the request string if needed, rather than exposing another public C API method we would need to then maintain for backward compatibility and such, as Guan suggests.

Users that want to do this kind of logging with C API in custom backend themselves can just make use of the existing `TRITONBACKEND_RequestId` to get the ID when logging. We don't need to provide that API for them.

dyastremsky: Thank you both for your comments and insights. I removed the API changes. Upon some more thought, we're primarily using this in `core` (where the API calls are unnecessary) and `backends` (where we have a common utility) for now. Hopefully, this and the related PRs look cleaner now.

GuanLuo (approved):

dyastremsky:
@rmccorm4 @GuanLuo This is now passing CI. The changes in this repo were: reverting the API version change, adding a default unknown request ID value for LogRequest(), and updating the query test to expect an unknown request ID. Please see last two commits.

(@rmccorm4 For the last one, it doesn't appear there's an easy gunit way to do "AssertIn" or substring without using other frameworks/libraries, so this may be the best use of time.)

src/test/query_test.cc
```diff
@@ -298,7 +298,8 @@ TEST_F(QueryTest, NoQueryFn)
   ASSERT_TRUE(err != nullptr) << "Expect error";
   EXPECT_EQ(TRITONSERVER_ErrorCode(err), TRITONSERVER_ERROR_UNAVAILABLE);
   EXPECT_STREQ(
-      TRITONSERVER_ErrorMessage(err), "Output properties are not available");
+      TRITONSERVER_ErrorMessage(err),
+      "[request id: <id_unknown>]Output properties are not available");
```
GuanLuo, line 302: add space `<id_unknown>] Output`

GuanLuo (commented):
LGTM, need minor formatting fix

src/test/query_test.cc
```diff
@@ -298,7 +298,8 @@ TEST_F(QueryTest, NoQueryFn)
   ASSERT_TRUE(err != nullptr) << "Expect error";
   EXPECT_EQ(TRITONSERVER_ErrorCode(err), TRITONSERVER_ERROR_UNAVAILABLE);
   EXPECT_STREQ(
-      TRITONSERVER_ErrorMessage(err), "Output properties are not available");
+      TRITONSERVER_ErrorMessage(err),
+      "[request id: <id_unknown>] Output properties are not available");
```
rmccorm4, line 302: To your comment about gtest not supporting substring easily, I think you can do something like this:
```
using ::testing::HasSubstr;  // I think you'll need this up top or somewhere similar

EXPECT_THAT(
  TRITONSERVER_ErrorMessage(err), 
  HasSubstr("Output properties are not available")
);
```

src/test/query_test.cc
dyastremsky: Updated. No longer includes the [request id]... match. Also added a space after the closing bracket in LogRequest().

GuanLuo (approved):

GuanLuo (approved):

rmccorm4 (approved):

PR #86, opened 2022-06-08
Use thread pool to load models (no boost)

rmccorm4:
Fixes DLIS-1833 / DLIS-3780 and removes need for 100ms sleep in model load thread due to [glibc bug](https://sourceware.org/bugzilla/show_bug.cgi?id=19329).

[common] Adds custom ThreadPool implementation in common: https://github.com/triton-inference-server/common/pull/60
[core] Use thread pool for loading models in model_repository_manager.cc: https://github.com/triton-inference-server/core/pull/86
[server] Expose num threads to server CLI: https://github.com/triton-inference-server/server/pull/4500

### DIFF
--- a/src/tritonserver.cc
+++ b/src/tritonserver.cc
@@ -269,6 +269,9 @@ class TritonServerOptions {
     buffer_manager_thread_count_ = c;
   }
 
+  unsigned int ModelLoadThreadCount() const { return model_load_thread_count_; }
+  void SetModelLoadThreadCount(unsigned int c) { model_load_thread_count_ = c; }
+
   bool Metrics() const { return metrics_; }
   void SetMetrics(bool b) { metrics_ = b; }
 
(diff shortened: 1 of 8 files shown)


src/tritonserver.cc
```diff
@@ -1199,6 +1205,16 @@ TRITONSERVER_ServerOptionsSetBufferManagerThreadCount(
   return nullptr;  // Success
 }
 
+TRITONAPI_DECLSPEC TRITONSERVER_Error*
+TRITONSERVER_ServerOptionsSetModelLoadThreadCount(
+    TRITONSERVER_ServerOptions* options, unsigned int thread_count)
+{
+  TritonServerOptions* loptions =
+      reinterpret_cast<TritonServerOptions*>(options);
+  loptions->SetModelLoadThreadCount(thread_count);
```
nv-kmcgill53, line 1214: Question: Since this is a user provided variable, should we set an upper bound or is that being too opinionated? The user could potentially set thread count to -1 and lock up the system by generating a large number of threads. What would be a reasonable max here, I'm thinking `2 * std::thread::hardware_concurrency()` or `3 * std::thread::hardware_concurrency()`

rmccorm4: Personally, I think it's opinionated to limit it, it's up to the user to pick their desired value if they choose to set it at all. If they don't set it, they will get a reasonable default.

nv-kmcgill53 (approved):

GuanLuo (approved):

tanmayv25 (approved):

PR #90, opened 2022-06-29
Normalize the updated model configuration received from backend

tanmayv25:
Running complete normalizing for future-proofing auto-complete. In case other, features gets added to the auto-complete logic on backend side.

I didn't move all the auto-complete logic from core to backend utilities because the changes will not be backwards compatible with our suggestion. We don't require users to essentially use our backend utilities. Some server-side auto-complete is definitely needed. 
https://github.com/triton-inference-server/backend/blob/main/include/triton/backend/backend_model.h#L40-L41

### DIFF
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -981,9 +981,9 @@ TRITONBACKEND_ModelAutoCompleteConfig(
 /// context will result in an error being returned. Additionally, Triton server
 /// can add some of the missing fields in the provided config with this call.
 /// The backend must get the complete configuration again by using
-/// TRITONBACKEND_ModelConfig. The function does not take ownership of the
-/// message object and so the caller should call TRITONSERVER_MessageDelete to
-/// release the object once the function returns.
+/// TRITONBACKEND_ModelConfig. TRITONBACKEND_ModelSetConfig does not take
+/// ownership of the message object and so the caller should call
+/// TRITONSERVER_MessageDelete to release the object once the function returns.
 ///
 /// \param model The model.
 /// \param config_version The format version of the model configuration.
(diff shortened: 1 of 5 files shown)


src/model_config_utils.h
```diff
@@ -105,6 +98,21 @@ Status AutoCompleteBackendFields(
     const std::string& model_name, const std::string& model_path,
     inference::ModelConfig* config);
 
+/// Detects and adds missing fields in the model configuration.
+/// \param config The model configuration
+/// \param min_compute_capability The minimum support CUDA compute
```
rmccorm4, line 103: support -> supported

src/backend_model.cc
```diff
@@ -353,6 +353,10 @@ TritonModel::UpdateModelConfig(
             .c_str());
   }  // else do nothing
 
+  // The updated model config may have added some
+  // fields that may need normalizing.
```
rmccorm4, line 357: remove second `may`, don't need `may` twice:
> The updated model config may have added some fields that need normalizing.

rmccorm4 (changes requested):

nv-kmcgill53 (approved):

rmccorm4 (approved):

include/triton/core/tritonbackend.h
```diff
@@ -978,7 +978,10 @@ TRITONBACKEND_ModelAutoCompleteConfig(
 /// scheduling choice can only be changed if none is previously set. Any other
 /// changes to the model configuration will be ignored by Triton. This function
 /// can only be called from TRITONBACKEND_ModelInitialize, calling in any other
-/// context will result in an error being returned. The function does not take
+/// context will result in an error being returned. Additionally, Triton server
+/// can add some of the missing fields in the provided config with this call.
```
Tabrizian, line 982: Is this the new behavior or it existed before too and we are just trying to better document it?

Tabrizian (approved):

include/triton/core/tritonbackend.h
tanmayv25: It is a new behavior added by enabling dynamic batching as part of auto-complete.

PR #92, opened 2022-07-01
Move ModelLifeCycle class to separate file. Cleanup model lifecycle logic

GuanLuo:


src/model_lifecycle.cc
```diff
@@ -0,0 +1,761 @@
...
+        }
+      }
+    }
+
+    // Load model asynchronously via thread pool
+    load_pool_->Enqueue([this, model_name, version, model_info, OnComplete,
+                         load_tracker]() {
+      CreateModel(model_name, version, model_info);
+      OnLoadComplete(model_name, version, model_info, OnComplete, load_tracker);
+    });
```
rmccorm4, line 508: Is it possible the unique_ptr around `model_info` doesn't get moved here and goes out of scope, invalidating the raw pointer `model_info` passed to CreateModel/OnLoadComplete ?

i.e. What if `linfo->state_` isn't LOADING or UNLOADING, it won't get moved to background models and may go out of scope?

src/model_lifecycle.cc
```diff
@@ -0,0 +1,761 @@
...
+          // Use recursive mutex as this deleter is likely to to be called
+          // within ModelLifeCycle class where the same mutex is being hold.
+          // However, mutex acquisition is needed here for the case where
+          // the model is requested to be unloaded while there are inflight
+          // requests, then the deleter will be called from the inference thread
+          {
+            std::lock_guard<std::recursive_mutex> lock(model_info->mtx_);
+            model_info->state_ = ModelReadyState::UNAVAILABLE;
+            model_info->state_reason_ = "unloaded";
+          }
```
rmccorm4, line 599: I don't see anything wrong here yet, but this comment stands out to me in regards to the bug you're seeing:

> then the deleter will be called from the inference thread

Maybe there's an issue when deleter is called from inference thread.

src/model_lifecycle.cc
```diff
@@ -0,0 +1,761 @@
...
+
+          // Check if the model info is in background, if so, remove from the
+          // map
+          std::lock_guard<std::recursive_mutex> lk(this->map_mtx_);
```
rmccorm4, line 603: Would there be any improvement, or ease of debugging if we had separate mutexes instead of 1 shared mutex for both models and background_models ?

src/model_lifecycle.cc
```diff
@@ -0,0 +1,761 @@
...
+
+  return Status::Success;
+}
+
+void
+ModelLifeCycle::CreateModel(
+    const std::string& model_name, const int64_t version, ModelInfo* model_info)
+{
+  LOG_VERBOSE(2) << "CreateModel() '" << model_name << "' version " << version;
+  const auto& model_config = model_info->model_config_;
```
rmccorm4, line 519: There used to be a [lock](https://github.com/triton-inference-server/core/blob/6a984dd747df4b0cc1782f63169a16ef9eb1ea7c/src/model_repository_manager.cc#L1259-L1263) on the model_info->mtx here before getting the model_config - is this no longer necessary?

src/model_lifecycle.h
```diff
@@ -0,0 +1,283 @@
...
+      const std::string& model_name, const int64_t version,
+      ModelInfo* model_info, std::function<void(Status)> OnComplete,
+      std::shared_ptr<LoadTracker> load_tracker);
+
+  const double min_compute_capability_;
+
+  // Mutex for 'map_' and 'background_models_'
+  std::recursive_mutex map_mtx_;
+
+  using VersionMap = std::map<int64_t, std::unique_ptr<ModelInfo>>;
```
rmccorm4, line 269: Why did VersionMap [used to map](https://github.com/triton-inference-server/core/blob/6a984dd747df4b0cc1782f63169a16ef9eb1ea7c/src/model_repository_manager.cc#L589-L591) to a `std::pair<ModelInfo*, ModelInfo*>` and now it just maps to a single `ModelInfo*`?

Was this how background models was handled previously?

src/model_lifecycle.cc
```diff
@@ -0,0 +1,761 @@
...
+      for (auto& version_info : it->second) {
+        auto& mi = version_info.second;
+        std::lock_guard<std::recursive_mutex> info_lk(mi->mtx_);
+        if ((mi->state_ == ModelReadyState::READY) &&
+            (mi->last_update_ns_ < load_tracker->last_update_ns_)) {
+          if ((mi->agent_model_list_ != nullptr) &&
+              (mi->agent_model_list_->LastActionType() ==
+               TRITONREPOAGENT_ACTION_LOAD_COMPLETE)) {
+            auto status = mi->agent_model_list_->InvokeAgentModels(
+                TRITONREPOAGENT_ACTION_UNLOAD);
```
rmccorm4, line 695: Can you explain this part? 

1. General explanation
2. What is the meaning of LastActionType == TRITONREPOAGENT_ACTION_LOAD_COMPLETE
3. Does this unload all previously loaded versions of a model? This sounds like if a new load fails, it will unload the already loaded/running model which is counterintuitive to the fallback logic we discussed for failed loads?

rmccorm4 (commented):
@GuanLuo Added general comments trying to find the bug with concurrent loads/inference

src/model_lifecycle.cc
GuanLuo: If it is not `LOADING / UNLOADING`, then it won't be referenced in the future (no callback pending), so we can let it go out of scope and clear it now.

src/model_lifecycle.cc
GuanLuo: The comment should be updated.. In `ModelDeleter` you can see that the callback is being invoked in a detached thread.

src/model_lifecycle.cc
GuanLuo: I didn't state this explicitly, in the current implementation, [some members](https://github.com/triton-inference-server/core/pull/92/files#diff-5df42a57c62d9f5cf01af9061da771a5e827f41e746e8c777ce021de3a14e5d7R207-R209) in `ModelInfo` will remain unchanged, so we don't really need a lock around those members. In previous implementation, this is not the case because we may reuse the `ModelInfo`

src/model_lifecycle.cc
```diff
@@ -0,0 +1,760 @@
...
+                                    agent_model_list]() mutable {
+          LOG_VERBOSE(2) << "OnDestroy callback() '" << model_name
+                         << "' version " << version;
+          LOG_INFO << "successfully unloaded '" << model_name << "' version "
+                   << version;
+          // Use recursive mutex as this deleter is likely to to be called
+          // within ModelLifeCycle class where the same mutex is being hold.
+          // However, mutex acquisition is needed here for the case where
+          // the model is requested to be unloaded while there are inflight
+          // requests, then the deleter will be called from the inference thread
```
rmccorm4, line 593: Update this comment:
1. not using recursive mutex anymore
2. > The comment should be updated.. In ModelDeleter you can see that the callback is being invoked in a detached thread.

src/model_lifecycle.cc
```diff
@@ -0,0 +1,760 @@
...
+          }
+
+          mi->state_ = ModelReadyState::UNLOADING;
+          mi->state_reason_.clear();
+          mi->agent_model_list_.reset();
+          mi->model_.reset();
+        }
+      }
+
+      // Mark current versions ready and track info in frontground
```
rmccorm4, line 709: find/replace all "frontground" -> "foreground"

src/model_lifecycle.cc
```diff
@@ -0,0 +1,760 @@
...
+    VersionStateMap version_map;
+
+    for (auto& version_model : model_version.second) {
+      std::lock_guard<std::mutex> lock(version_model.second->mtx_);
+      if (strict_readiness &&
+          version_model.second->state_ != ModelReadyState::READY) {
+        continue;
+      }
+
+      // At lease one version is live (ready / loading / unloading)
```
rmccorm4, line 207:
```suggestion
      // At least one version is live (ready / loading / unloading)
```

src/model_lifecycle.cc
```diff
@@ -0,0 +1,760 @@
...
+              << "Agent model returns error on TRITONREPOAGENT_ACTION_UNLOAD: "
+              << status.AsString();
+        }
+      }
+
+      // unload
+      model_info->state_ = ModelReadyState::UNLOADING;
+      model_info->state_reason_.clear();
+      model_info->agent_model_list_.reset();
+      model_info->model_.reset();
```
rmccorm4, line 423: Can you add a `ModelInfo::Reset()` or `Unload()` or similar that does all of these actions, and call it throughout the code? This will be easier to maintain if more things need to be done on unload. Seems to happen in a enough places to be worth adding a helper:
- https://github.com/triton-inference-server/core/blob/9676048c25581f58f25a32ecfcbde4ca72f65010/src/model_lifecycle.cc#L420
- https://github.com/triton-inference-server/core/blob/9676048c25581f58f25a32ecfcbde4ca72f65010/src/model_lifecycle.cc#L667
- https://github.com/triton-inference-server/core/blob/9676048c25581f58f25a32ecfcbde4ca72f65010/src/model_lifecycle.cc#L702

src/model_lifecycle.cc
rmccorm4: Can we make the code something like:
```
// Only need to invoke UNLOAD action if ...
// Otherwise ...
const bool is_first_unload = 
  (mi->agent_model_list_ != nullptr) && 
  (mi->agent_model_list_->LastActionType() == TRITONREPOAGENT_ACTION_LOAD_COMPLETE)

if (is_first_unload) {
  auto status = mi->agent_model_list_->InvokeAgentModels(TRITONREPOAGENT_ACTION_UNLOAD);
}
```

I find this code hard to read/understand at a glance without being very familiar with the implementation

src/model_repository_manager.h
```diff
@@ -39,42 +41,15 @@ namespace triton { namespace core {
 class InferenceServer;
 class Model;
 
+// [FIXME] should have separated load / unload functions for clarity
+enum ActionType { NO_ACTION, LOAD, UNLOAD };
```
tanmayv25, line 45: Do we still need NO_ACTION? When would this be needed?

src/model_lifecycle.cc
GuanLuo: I don't think it would be helpful for this case, even their names seem unrelated, they are often manipulated in group to complete one operation. Also, having multiple mutexes also means to hold multiple locks in those operations and then the order of acquisition may become the problem again.

src/model_repository_manager.h
GuanLuo: We don't need it, I just didn't modify this enum as it will be removed completely if model repository manager is going to be refactored.

rmccorm4:
Great refactor 👍

PR #94, opened 2022-07-08
Add per-response statistics

Tabrizian:
Example output:

```JSON
{
  "model_stats": [
    {
      "name": "square_int32",
      "version": "1",
      "last_inference": 0,
      "inference_count": 0,
      "execution_count": 0,
      "inference_stats": {
        "success": {
          "count": 0,
          "ns": 0
        },
        "fail": {
          "count": 0,
          "ns": 0
        },
        "queue": {
          "count": 0,
          "ns": 0
        },
        "compute_input": {
          "count": 0,
          "ns": 0
        },
        "compute_infer": {
          "count": 0,
          "ns": 0
        },
        "compute_output": {
          "count": 0,
          "ns": 0
        },
        "cache_hit": {
          "count": 0,
          "ns": 0
        },
        "cache_miss": {
          "count": 0,
          "ns": 0
        }
      },
      "batch_stats": [],
      "response_stats": [
        {
          "responses": [
            {
              "index": 0,
              "success": {
                "count": 3758,
                "ns": 104645144
              },
              "fail": {
                "count": 0,
                "ns": 0
              },
              "compute_infer": {
                "count": 3758,
                "ns": 312936
              },
              "compute_output": {
                "count": 3758,
                "ns": 83222303
              }
            },
            {
              "index": 1,
              "success": {
                "count": 3758,
                "ns": 71950386
              },
              "fail": {
                "count": 0,
                "ns": 0
              },
              "compute_infer": {
                "count": 3758,
                "ns": 284120
              },
              "compute_output": {
                "count": 3758,
                "ns": 59387380
              }
            }
          ]
        },
        {
          "responses": [
            {
              "index": 0,
              "success": {
                "count": 3725,
                "ns": 111852810
              },
              "fail": {
                "count": 0,
                "ns": 0
              },
              "compute_infer": {
                "count": 3725,
                "ns": 321325
              },
              "compute_output": {
                "count": 3725,
                "ns": 85285353
              }
            },
            {
              "index": 1,
              "success": {
                "count": 3725,
                "ns": 79317289
              },
              "fail": {
                "count": 0,
                "ns": 0
              },
              "compute_infer": {
                "count": 3725,
                "ns": 284731
              },
              "compute_output": {
                "count": 3725,
                "ns": 64023205
              }
            },
            {
              "index": 2,
              "success": {
                "count": 3725,
                "ns": 65396800
              },
              "fail": {
                "count": 0,
                "ns": 0
              },
              "compute_infer": {
                "count": 3725,
                "ns": 274910
              },
              "compute_output": {
                "count": 3725,
                "ns": 53976564
              }
            }
          ]
        }
      ]
    }
  ]
}
```


src/infer_response.h
```diff
@@ -118,10 +119,18 @@ class InferenceResponseFactory {
...
       response_delegator_;
 
+
 #ifdef TRITON_ENABLE_TRACING
   // Inference trace associated with this response.
   std::shared_ptr<InferenceTraceProxy> trace_;
 #endif  // TRITON_ENABLE_TRACING
+
+  // The internal unique ID of the request
+  uint64_t request_id_;
```
rmccorm4, line 129: How does `request_id_` differ from `id_` ?
```
  // The ID of the corresponding request that should be included in
  // every response.
  std::string id_;
```

Maybe the comments should be updated to make them more clearly distinct

tanmayv25: The id_ is a property that can be optionally provided by the user. request_id_ is a unique identification that triton attaches to the request to make this feature work for cases where user doesnot provide id_. 
I agree it is little confusing within this header. 
@Tabrizian  We can elaborate this in L102-L103 about the nature of id_ field.

src/infer_response.h
```diff
@@ -338,10 +356,22 @@ class InferenceResponse {
...
   std::shared_ptr<InferenceTraceProxy> trace_;
 #endif  // TRITON_ENABLE_TRACING
+
+  // Representing the response index.
+  uint64_t response_idx_;
+
+  // Representing the request id that the response was created from.
+  uint64_t request_id_;
+
+  uint64_t response_start_;
```
tanmayv25, line 374: For sense of completion, add a descriptive comment here.

src/infer_response.h
```diff
@@ -255,6 +265,14 @@ class InferenceResponse {
   // The response outputs.
   const std::deque<Output>& Outputs() const { return outputs_; }
 
+  // The response index.
+  uint64_t ResponseIdx() const { return response_idx_; }
+
+  // The unique request ID stored in triton.
+  uint64_t RequestUniqueId() const { return request_id_; }
+
+  uint64_t ResponseStartNs() const { return response_start_; }
```
tanmayv25, line 274: Add a descriptive comment here for uniformity.

src/common.h
```diff
@@ -0,0 +1,44 @@
...
+// OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#pragma once
+
+#include <atomic>
+#include <mutex>
+
+namespace triton { namespace core {
+class CountedObject {
```
tanmayv25, line 32: Why is this in a common.h? Why not counted_object.h?
Also, why do we need this class? Why would a simple atomic variable not work for us? All the class is doing is getting a unique id by incrementing the last. The returned instance_id can be stored within the Response/Request class objects.

src/backend_model_instance.cc
```diff
@@ -962,5 +962,24 @@ TRITONBACKEND_ModelInstanceReportBatchStatistics(
...
+    TRITONBACKEND_ModelInstance* instance, TRITONBACKEND_Response* response,
+    const bool success, const uint64_t response_start,
+    const uint64_t compute_output_start, const uint64_t response_end,
+    const uint32_t response_flags)
+{
+#ifdef TRITON_ENABLE_STATS
+  TritonModelInstance* ti = reinterpret_cast<TritonModelInstance*>(instance);
+  InferenceResponse* ir = reinterpret_cast<InferenceResponse*>(response);
+
+  INFER_STATS_DECL_TIMESTAMP(response_end_ns);
```
tanmayv25, line 976: `response_end` and `response_end_ns` is confusing. Do we expect a lot of difference between the timestamps currently being captured here?
It is kind of user-dependent by the fact when exactly they call TRITONBACKEND_ModelInstanceReportResponseStatistics, but do we care to capture `response_end_ns`?

src/infer_response.h
```diff
@@ -118,10 +119,18 @@ class InferenceResponseFactory {
...
   // Inference trace associated with this response.
   std::shared_ptr<InferenceTraceProxy> trace_;
 #endif  // TRITON_ENABLE_TRACING
+
+  // The internal unique ID of the request
+  uint64_t request_id_;
+
+  // Response index
+  uint64_t total_response_idx_;
+  std::mutex mu_;
```
tanmayv25, line 133: Should comment here what does this mutex protects.

src/backend_model_instance.cc
tanmayv25: I think compute_output_end should be treated as response_end.

tanmayv25 (changes requested):

src/infer_response.h
kthui: Updated.

include/triton/core/tritonbackend.h
```diff
@@ -1254,6 +1254,55 @@ TRITONBACKEND_ModelInstanceReportBatchStatistics(
     const uint64_t exec_start_ns, const uint64_t compute_start_ns,
     const uint64_t compute_end_ns, const uint64_t exec_end_ns);
 
+/// Record statistics for an inference response.
+///
+/// Set 'success' true to indicate that the inference response completed
```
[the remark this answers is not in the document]
kthui: Updated.

src/infer_response.h
kthui: Updated.

src/infer_response.h
kthui: Updated.

src/common.h
kthui: Updated.

src/backend_model_instance.cc
kthui: Updated.

src/infer_response.h
kthui: Updated, to use atomic variable.

kthui:
Rebased.
Changes since last review: 
```
$ git diff d7f8e411231431d58a3c5ed5008cb96263807826 07c66fe128bbb2c36c19c5115348a17e8c7a8add
diff --git a/include/triton/core/tritonbackend.h b/include/triton/core/tritonbackend.h
index cff61da..5c2deef 100644
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -1257,7 +1257,7 @@ TRITONBACKEND_ModelInstanceReportBatchStatistics(
 
 /// Record statistics for an inference response.
 ///
-/// Set 'success' true to indicate that the inference response completed
+/// Set 'success' to true to indicate that the inference response completed
 /// successfully. In this case all timestamps should be non-zero values reported
 /// in nanoseconds and should be collected using
 /// std::chrono::steady_clock::now().time_since_epoch() or the equivalent.  Set
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 4e22fb7..5fb4062 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -106,7 +106,6 @@ set(
   backend_model.cc
   backend_model_instance.cc
   buffer_attributes.cc
-  common.cc
   cuda_utils.cc
   dynamic_batch_scheduler.cc
   ensemble_scheduler.cc
@@ -150,7 +149,6 @@ set(
   backend_model.h
   backend_model_instance.h
   buffer_attributes.h
-  common.h
   constants.h
   cuda_utils.h
   dynamic_batch_scheduler.h
diff --git a/src/backend_model_instance.cc b/src/backend_model_instance.cc
index 35a7b7e..2eb0f53 100644
--- a/src/backend_model_instance.cc
+++ b/src/backend_model_instance.cc
@@ -966,7 +966,7 @@ TRITONAPI_DECLSPEC TRITONSERVER_Error*
 TRITONBACKEND_ModelInstanceReportResponseStatistics(
     TRITONBACKEND_ModelInstance* instance, TRITONBACKEND_Response* response,
     const bool success, const uint64_t response_start,
-    const uint64_t compute_output_start, const uint64_t response_end,
+    const uint64_t compute_output_start, const uint64_t compute_output_end,
     const uint32_t response_flags)
 {
 #ifdef TRITON_ENABLE_STATS
@@ -976,7 +976,7 @@ TRITONBACKEND_ModelInstanceReportResponseStatistics(
   INFER_STATS_DECL_TIMESTAMP(response_end_ns);
   ti->Model()->MutableStatsAggregator()->UpdateResponse(
       ti->MetricReporter(), ir, response_start, compute_output_start,
-      response_end, response_end_ns, response_flags, success);
+      compute_output_end, response_end_ns, response_flags, success);
 #endif  // TRITON_ENABLE_STATS
   return nullptr;
 }
diff --git a/src/common.cc b/src/common.cc
deleted file mode 100644
index 82b7bc9..0000000
--- a/src/common.cc
+++ /dev/null
@@ -1,39 +0,0 @@
-// Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-//
-// Redistribution and use in source and binary forms, with or without
-// modification, are permitted provided that the following conditions
-// are met:
-//  * Redistributions of source code must retain the above copyright
-//    notice, this list of conditions and the following disclaimer.
-//  * Redistributions in binary form must reproduce the above copyright
-//    notice, this list of conditions and the following disclaimer in the
-//    documentation and/or other materials provided with the distribution.
-//  * Neither the name of NVIDIA CORPORATION nor the names of its
-//    contributors may be used to endorse or promote products derived
-//    from this software without specific prior written permission.
-//
-// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
-// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
-// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
-// PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
-// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
-// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
-// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
-// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
-// OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
-// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
-// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-
-#include "common.h"
-
-namespace triton { namespace core {
-
-uint64_t CountedObject::instance_count_ = 0;
-
-CountedObject::CountedObject()
-{
-  std::lock_guard<std::mutex> guard(mu_);
-  instance_index_ = CountedObject::instance_count_;
-  instance_count_++;
-}
-}};  // namespace triton::core
diff --git a/src/common.h b/src/common.h
deleted file mode 100644
index 32e9fc5..0000000
--- a/src/common.h
+++ /dev/null
@@ -1,44 +0,0 @@
-// Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-//
-// Redistribution and use in source and binary forms, with or without
-// modification, are permitted provided that the following conditions
-// are met:
-//  * Redistributions of source code must retain the above copyright
-//    notice, this list of conditions and the following disclaimer.
-//  * Redistributions in binary form must reproduce the above copyright
-//    notice, this list of conditions and the following disclaimer in the
-//    documentation and/or other materials provided with the distribution.
-//  * Neither the name of NVIDIA CORPORATION nor the names of its
-//    contributors may be used to endorse or promote products derived
-//    from this software without specific prior written permission.
-//
-// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
-// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
-// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
-// PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
-// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
-// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
-// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
-// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
-// OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
-// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
-// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-#pragma once
-
-#include <atomic>
-#include <mutex>
-
-namespace triton { namespace core {
-class CountedObject {
- public:
-  CountedObject();
-
- protected:
-  uint64_t InstanceIndex() { return instance_index_; }
-
- private:
-  static uint64_t instance_count_;
-  uint64_t instance_index_;
-  std::mutex mu_;
-};
-}};  // namespace triton::core
diff --git a/src/infer_request.h b/src/infer_request.h
index b43daa4..d5f551b 100644
--- a/src/infer_request.h
+++ b/src/infer_request.h
@@ -31,7 +31,6 @@
 #include <unordered_map>
 #include <vector>
 #include "buffer_attributes.h"
-#include "common.h"
 #include "infer_response.h"
 #include "infer_stats.h"
 #include "infer_trace.h"
@@ -48,6 +47,9 @@ class Model;
 class InferenceServer;
 class MetricModelReporter;
 
+// Count the number of InferenceRequest instance created
+static std::atomic<uint64_t> inference_request_count_(0);
+
 //
 // An inference request. A request can be used multiple times for
 // inference but before each inference run, PrepareForInference() must
@@ -56,7 +58,7 @@ class MetricModelReporter;
 // valid. Preparing involves removing/resetting any state left over
 // from the previous inference.
 //
-class InferenceRequest : CountedObject {
+class InferenceRequest {
  public:
   // Input tensor
   class Input {
@@ -284,6 +286,7 @@ class InferenceRequest : CountedObject {
         requested_model_version_(requested_model_version), flags_(0),
         correlation_id_(0), batch_size_(0), timeout_us_(0), collect_stats_(true)
   {
+    inference_request_count_++;
     SetPriority(0);
   }
 
@@ -484,7 +487,7 @@ class InferenceRequest : CountedObject {
   {
     response_factory_.reset(new InferenceResponseFactory(
         model_shared_, id_, allocator, alloc_userp, response_fn, response_userp,
-        response_delegator_, InstanceIndex()));
+        response_delegator_, inference_request_count_));
     return Status::Success;
   }
 
diff --git a/src/infer_response.cc b/src/infer_response.cc
index 063fd90..91cfc16 100644
--- a/src/infer_response.cc
+++ b/src/infer_response.cc
@@ -40,12 +40,7 @@ Status
 InferenceResponseFactory::CreateResponse(
     std::unique_ptr<InferenceResponse>* response)
 {
-  uint64_t response_index;
-  {
-    std::lock_guard<std::mutex> guard{mu_};
-    response_index = total_response_idx_;
-    total_response_idx_++;
-  }
+  uint64_t response_index = total_response_idx_++;
   response->reset(new InferenceResponse(
       model_, id_, allocator_, alloc_userp_, response_fn_, response_userp_,
       response_delegator_, response_index, request_id_));
diff --git a/src/infer_response.h b/src/infer_response.h
index f9da7e2..43c1687 100644
--- a/src/infer_response.h
+++ b/src/infer_response.h
@@ -25,6 +25,7 @@
 #pragma once
 
+#include <atomic>
 #include <deque>
 #include <functional>
 #include <string>
@@ -99,8 +100,8 @@ class InferenceResponseFactory {
   // to handle the nullptr case.
   std::shared_ptr<Model> model_;
 
-  // The ID of the corresponding request that should be included in
-  // every response.
+  // The ID of the corresponding request that should be included in every
+  // response. This is a property that can be optionally provided by the user.
   std::string id_;
 
   // The response allocator and user pointer. The 'allocator_' is a
@@ -125,12 +126,12 @@ class InferenceResponseFactory {
   std::shared_ptr<InferenceTraceProxy> trace_;
 
-  // The internal unique ID of the request
+  // The internal unique ID of the request. This is a unique identification
+  // that triton attaches to each request in case user does not provide id_.
   uint64_t request_id_;
 
   // Response index
-  uint64_t total_response_idx_;
-  std::mutex mu_;
+  std::atomic<uint64_t> total_response_idx_;
 };
 
 //
@@ -270,7 +271,7 @@ class InferenceResponse {
 
   // The unique request ID stored in triton.
   uint64_t RequestUniqueId() const { return request_id_; }
-
+  // The timestamp in nanoseconds when the response started.
   uint64_t ResponseStartNs() const { return response_start_; }
 
   // Add an output to the response. If 'output' is non-null
@@ -370,7 +371,7 @@ class InferenceResponse {
 
   // Representing the request id that the response was created from.
   uint64_t request_id_;
-
+  // Timestamp in nanoseconds when the response started.
   uint64_t response_start_;
 };
 
diff --git a/src/test/CMakeLists.txt b/src/test/CMakeLists.txt
index 2b9c848..6c65ed4 100644
--- a/src/test/CMakeLists.txt
+++ b/src/test/CMakeLists.txt
@@ -94,8 +94,6 @@ if(${TRITON_ENABLE_GPU})
     ../status.cc
     ../status.h
     ../constants.h
-    ../common.h
-    ../common.cc
     ${MEMORY_SRCS}
     ${CUDA_MEMORY_MANAGER_SRCS}
     ${PINNED_MEMORY_MANAGER_SRCS}
$ 
```

src/infer_response.cc
```diff
@@ -40,12 +40,7 @@ Status
 InferenceResponseFactory::CreateResponse(
     std::unique_ptr<InferenceResponse>* response)
 {
-  uint64_t response_index;
-  {
-    std::lock_guard<std::mutex> guard{mu_};
-    response_index = total_response_idx_;
-    total_response_idx_++;
-  }
+  uint64_t response_index = total_response_idx_++;
```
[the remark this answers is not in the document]
kthui: Yes. I switched the total_response_idx_ to std::atomic<uint64_t> variable.
Please correct me if I am wrong, from the [c++ doc](https://en.cppreference.com/w/cpp/atomic/atomic/operator_arith), `uint64_t response_index = total_response_idx_++;` should atomically post-increment the variable and assign to "response_index". Then, it should do exactly the same thing as using a mutex in a block.

tanmayv25 (approved):

include/triton/core/tritonbackend.h
```diff
@@ -1255,6 +1255,67 @@ TRITONBACKEND_ModelInstanceReportBatchStatistics(
...
+/// Record statistics for an inference with no response.
+///
+/// This API is used with models that return no response. If a model returns
+/// one or more responses, the
+/// TRITONBACKEND_ModelInstanceReportResponseStatistics should be used instead.
+///
+/// \param instance The model instance.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONBACKEND_DECLSPEC TRITONSERVER_Error*
+TRITONBACKEND_ModelInstanceReportNoResponseStatistics(
```
Tabrizian, line 1317: @tanmayv25 / @kthui What if the response is using only flags but it is not the final response? It looks like the API doesn't permit reporting statistics for this scenario? This interface should be changed to:

```
TRITONBACKEND_ModelInstanceReportResponseFactoryStatistics(
    TRITONBACKEND_ModelInstance* instance, TRITONBACKEND_ResponseFactory* response_factory,
    const bool success, const uint64_t response_start,
    const uint64_t compute_output_start, const uint64_t response_end,
    const uint32_t response_flags);
```

kthui: @Tabrizian, please help me understand. Do you mean a scenario which a model can dynamically choose how many response(s) to return?

Given a model that will return at most 3 responses, per request.
```
| res 1 | -> | res 2 | -> | res 3 |
```
In the normal case, all three responses are computed and returned.
```
| res 1 | -> | res 2 | -> | res 3 |
   ✅           ✅          ✅
```
For some requests, the model does not return the second response.
```
| res 1 | -> | res 2 | -> | res 3 |
   ✅           ❓          ✅
```
Are you saying we should report statistics (i.e. compute time) for the second response? (where there is not a response object)

-----
If this is correct, maybe we can simply add a new metric to the response statistics, from `success`, `fail`, `compute_infer` and `compute_output` to `success`, `fail`, *`no_output`*, `compute_infer` and `compute_output`?

For example, if the model returns the following three responses:
```
| res 1 | -> | res 2 | -> | res 3 |
   ✅           ❓          ❌
```
Then, `res 1` is `success`, `res 2` is `no_output` and `res 3` is `fail`, and `compute_infer` and `compute_output` are calculated.

The newly added `response_stats` can be simplified into, with the above example: 
```
"response_stats": [
        {
                "index": 0,
                "success": {
                        "count": 1,
                        "ns": 10
                },
                "fail": {
                        "count": 0,
                        "ns": 0
                },
                "no_output": {
                        "count": 0,
                        "ns": 0
                },
                "compute_infer": {
                        "count": 1,
                        "ns": 4
                },
                "compute_output": {
                        "count": 1,
                        "ns": 5
                }
        }, 
        {
                "index": 1,
                "success": {
                        "count": 0,
                        "ns": 0
                },
                "fail": {
                        "count": 0,
                        "ns": 0
                },
                "no_output": {
                        "count": 1,
                        "ns": 8
                },
                "compute_infer": {
                        "count": 1,
                        "ns": 4
                },
                "compute_output": {
                        "count": 0,
                        "ns": 0
                }
        },
        {
                "index": 2,
                "success": {
                        "count": 0,
                        "ns": 10
                },
                "fail": {
                        "count": 1,
                        "ns": 10
                },
                "no_output": {
                        "count": 0,
                        "ns": 0
                },
                "compute_infer": {
                        "count": 1,
                        "ns": 4
                },
                "compute_output": {
                        "count": 1,
                        "ns": 5
                }
        }
]
```
By sending more requests, the respective `count` will increase correspondingly and the `ns` will update.
-----
CC @tanmayv25

src/infer_response.cc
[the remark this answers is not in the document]
Tabrizian: I see. I didn't know that the assignment is also atomic.

include/triton/core/tritonbackend.h
Tabrizian: > Are you saying we should report statistics (i.e. compute time) for the second response? (where there is not a response object)

Right, I'm saying that it looks like there is no way to report statistics for the second response in your example. I think the current backend API is too specific and should be extended to cover this scenario.

> If this is correct, maybe we can simply add a new metric to the response statistics, from success, fail, compute_infer and compute_output to success, fail, no_output, compute_infer and compute_output?

How about changing the HTTP/GRPC statistics output to something like below:

```
"response_stats": [
        {
                "index": 2,
                 response: {
                  "success": {
                          "count": 0,
                          "ns": 10
                  },
                  "fail": {
                          "count": 1,
                          "ns": 10
                  },
              
                  "compute_infer": {
                          "count": 1,
                          "ns": 4
                  },
                  "compute_output": {
                          "count": 1,
                          "ns": 5
                  },
                  "flag": "TRITONSERVER_RESPONSE_COMPLETE_FINAL"
                 },
                 response: {
                  "success": {
                          "count": 0,
                          "ns": 10
                  },
                  "fail": {
                          "count": 1,
                          "ns": 10
                  },
              
                  "compute_infer": {
                          "count": 1,
                          "ns": 4
                  },
                  "compute_output": {
                          "count": 1,
                          "ns": 5
                  },
                  "flag": "0"
                 }
        }
]
```

This would allow the API to be more extensible in case we add new flags to the response object.

kthui: Can you elaborate your through on the flag object a bit more? For example, what are some possible flags to be added in the future. It is because currently the flag is only used to detect if the provided response is the last one for this request.

There might be another scenario to be considered:
```
| res 1 | -> | res 2 | -> | res 3 |
   ✅           ❓          ✅
```
and
```
| res 1 | -> | res 2 | -> | res 3 |
   ✅           ✅          ❓
```
Both requests generate two responses, but the API cannot distinguish between them. Then, we might want the user to provide a statistics for `res 2` and `res 3` regardless whether a response is returned. Do you mean using the flag object to distinguish it?
If so, do we still need an array of `response_stats`, like `[2 responses grouped here], [3 responses grouped here]`? It is because there will always be 3 responses in the above scenario.

Tabrizian: I see. Looks like flag doesn't resolve the issue. How about renaming "flag" to "is_flag_only" which indicates whether a response is only flags and this will be set to true for responses reported using `TRITONBACKEND_ModelInstanceReportResponseFactoryStatistics`? For other response statistics this will be set to false?

kthui: I was thinking maybe we can simplify the response statistics to a table.

For example, the following responses are generated:
```
request 1: 
| res 1 | -> | res 2 | -> | res 3 |
   ✅           ❓          ✅

request 2:
| res 1 | -> | res 2 | -> | res 3 |
   ✅           ✅          ✅

request 3:
| res 1 | -> | res 2 | -> | res 3 |
   ✅           ✅          ❓

request 4:
| res 1 | -> | res 2 | -> | res 3 |
   ✅           ✅          ❌
```

We can aggregate by the `res #` for each request.

For example, before any request, the response statistics:
|  | Success | Fail | No output | Compute infer | Compute output |
| --- | --- | --- | --- | --- | --- |

After request 1:
|  | Success | Fail | No output | Compute infer | Compute output |
| --- | --- | --- | --- | --- | --- |
| res 1 | 1 | 0 | 0 | 1 | 1 |
| res 2 | 0 | 0 | 1 | 1 | 0 |
| res 3 | 1 | 0 | 0 | 1 | 1 |

After request 2:
|  | Success | Fail | No output | Compute infer | Compute output |
| --- | --- | --- | --- | --- | --- |
| res 1 | 2 | 0 | 0 | 2 | 2 |
| res 2 | 1 | 0 | 1 | 2 | 1 |
| res 3 | 2 | 0 | 0 | 2 | 2 |

After request 3:
|  | Success | Fail | No output | Compute infer | Compute output |
| --- | --- | --- | --- | --- | --- |
| res 1 | 3 | 0 | 0 | 3 | 3 |
| res 2 | 2 | 0 | 1 | 3 | 2 |
| res 3 | 2 | 0 | 1 | 3 | 2 |

After request 4:
|  | Success | Fail | No output | Compute infer | Compute output |
| --- | --- | --- | --- | --- | --- |
| res 1 | 4 | 0 | 0 | 4 | 4 |
| res 2 | 3 | 0 | 1 | 4 | 3 |
| res 3 | 2 | 1 | 1 | 3 | 2 |

Note: only count is shown above, ns average should be included as well.

The backend implementation must report statistics for each possible response under the appropriate category. The core will aggregate the reports into response statistics, available for query.

I think this should make the response statistics much simpler to understand, but I do not know if I missed any reportable metrics that is covered by the current design.

tanmayv25: I like the idea being proposed by @kthui. It hides the complexity of flags from the reported statistics. The onus is on backend developer to properly send the compute times for each response. The intent of statistics report should be how much work the backend expended on these request/response processing. 
I think we can remove `TRITONBACKEND_ModelInstanceReportNoResponseStatistics`. And ask users to properly set response_start, compute_output_starts and response_end timestamps in the `TRITONBACKEND_ModelInstanceReportResponseStatistics` itself for empty responses?

kthui: > I think we can remove `TRITONBACKEND_ModelInstanceReportNoResponseStatistics`. And ask users to properly set response_start, compute_output_starts and response_end timestamps in the `TRITONBACKEND_ModelInstanceReportResponseStatistics` itself for empty responses?

Yes. For grouping, we only need the `res #` for each call to `TRITONBACKEND_ModelInstanceReportResponseStatistics`.

I think we can either figure it out from the `request` objects' unique id and with the order of calling, or simply request the backend to provide the `res #` with each call.

kthui: @Tabrizian what do you think about the change away from flag?

Tabrizian: > I think this should make the response statistics much simpler to understand, but I do not know if I missed any reportable metrics that is covered by the current design.

The very first proposal was similar to what you describe here but the problem is that it is not possible find the answer to the questions below:

1. What is the average request time for requests that generate 4 responses?
2. How long does the third response take in requests that generate 4 responses?

These questions can be answered in the current design.

> I think we can remove TRITONBACKEND_ModelInstanceReportNoResponseStatistics. And ask users to properly set response_start, compute_output_starts and response_end timestamps in the TRITONBACKEND_ModelInstanceReportResponseStatistics itself for empty responses?

The idea of `TRITONBACKEND_ModelInstanceReportResponseFactoryStatistics` is similar to what you described here. It is exactly like the other API except it accepts a response factory object instead of a response object.

> For grouping, we only need the res # for each call to TRITONBACKEND_ModelInstanceReportResponseStatistics.

I think we should try to hide the response number if we can. It would make the API easier to use.

kthui: I see. An outer dimension to the table is needed to track statistics for requests that generate different number of responses. Then, it is very similar to the design you proposed.

GuanLuo: Jumping into the discussion without knowing the whole picture here:
```
| res 1 | -> | res 2 | -> | res 3 |
   ✅           ❓          ✅
```
What does this case even mean here? Why we want to add index on to the responses? Shouldn't this collapse to the case that the model generates 2 responses for the request? From the client's perspective, it only receives 2 responses at the end.

Note that "flag" is not a response, it is utility for Triton to manipulate the sequence of responses (now the only use case is where the response sequence end). Response is the way to convey messages, and if user does want to communicate data other than tensors, they should be able to describe it as one of the Triton model output. In this sense, res 2 will still be in the form of Triton response and then the example will be case the the model generates 3 responses for the request. On the client side they will read and identify that res 2 is some special response.

Tabrizian: > What does this case even mean here? Why we want to add index on to the responses? Shouldn't this collapse to the case that the model generates 2 responses for the request? From the client's perspective, it only receives 2 responses at the end.

IMO, the way that response to client relationship is currently handled is just one implementation of the Triton Server API. I also don't know exactly what this means or if there are use-cases for it but the API allows this. If customers are using this right now, they can't report statistics for this scenario.

 > Why we want to add index on to the responses?
 
 It is a way to distinguish the responses (e.g.. the first response took this time).

PR #96, opened 2022-07-13
Fix possible segfault from invalid infer shape input

kthui:
When a model config has `max_batch_size > 0` and `dims = [1]`, the server will copy the `shape` object without the outer most dimension from the `shape` during infer requests. The original copy process did not check if there are at least two elements in the `shape` object. Thus, if the user sets `shape = [1]`, the server can run into a segmentation fault by reading the second element from the `shape` object, which does not exist. This PR adds the check.

### DIFF
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -881,8 +881,14 @@ InferenceRequest::Normalize()
                 "'");
       }
 
-      input.MutableShape()->assign(
-          input.OriginalShape().begin() + 1, input.OriginalShape().end());
+      // When skipping the first dimension, make sure there is a second
+      // dimension
+      if (input.OriginalShape().size() > 1) {
+        input.MutableShape()->assign(
+            input.OriginalShape().begin() + 1, input.OriginalShape().end());
+      } else {
+        input.MutableShape()->clear();
+      }
     }
   }
 
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -863,11 +863,12 @@ InferenceRequest::Normalize()
         continue;
       }
 
-      if (input.OriginalShape().size() == 0) {
+      // Make sure shape has batch dimension and at least 1d shape for model
+      if (input.OriginalShape().size() < 2) {
         return Status(
             Status::Code::INVALID_ARG,
             LogRequest() + "input '" + input.Name() +
-                "' has no shape but model requires batch dimension for '" +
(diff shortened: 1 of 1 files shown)


src/infer_request.cc
```diff
@@ -881,8 +881,14 @@ InferenceRequest::Normalize()
                 "'");
       }
 
-      input.MutableShape()->assign(
-          input.OriginalShape().begin() + 1, input.OriginalShape().end());
+      // When skipping the first dimension, make sure there is a second
+      // dimension
+      if (input.OriginalShape().size() > 1) {
```
GuanLuo, line 886: Do you actually observe segfault without this change? I would imagine the iterator (`xxx.begin() + 1`)will be out of range only if the size is 0. If size is 1, then `begin() + 1` should equivalent to `end()` and `assign()` won't do anything?

GuanLuo: Perhaps changing the check in line 866 to `if (input.OriginalShape().size() < 2) {`? At this point we should assume the input has at least two dimensions (input config shape which is > 0, + batch dim), although the error message needs to be changed to reflect that

kthui: Sounds good, updated.
What do you think of the updated error message?

rmccorm4:
@kthui what happened to this PR? Is it still needed?

kthui:
> @kthui what happened to this PR? Is it still needed?

Hi @GuanLuo, do you think the presence of batch dimension is checked somewhere else? If so, then this added check is not necessary and we can close this PR.

kthui:
@GuanLuo @tanmayv25, do you think we need this PR. If this PR doesn't fix a possible segfault scenario, we can close it.

GuanLuo:
Do you have a test case that will fail (segfault) without this change?

kthui:
>Do you have a test case that will fail (segfault) without this change?

No. This change is only based on theory.

tanmayv25:
Does your change handle the case where we use reshape to support a single batch size dimension?
Read more here: https://github.com/triton-inference-server/server/blob/main/docs/user_guide/model_configuration.md#reshape

Was the CI run successful with this change? If you can come up with a test case that gets fixed with this change then we can merge this change.

kthui:
DLIS-4317

GuanLuo:
Closing this PR until there is proof that this is indeed broken, I will be surprised that the assign implementation does not perform range check before accessing element

PR #98, opened 2022-07-15
Core repo changes for new logging extension

fpetrini15:


src/tritonserver.cc
```diff
@@ -1215,6 +1215,20 @@ TRITONSERVER_ServerOptionsSetModelLoadThreadCount(
   return nullptr;  // Success
 }
 
+TRITONAPI_DECLSPEC TRITONSERVER_Error*
+TRITONSERVER_ServerOptionsSetLogOutFile(
+    TRITONSERVER_ServerOptions* options, const char* file)
+{
+#ifdef TRITON_ENABLE_LOGGING
+  std::string out_file(file);
```
Tabrizian, line 1223: Do we need to add a check here to make sure that the log file is not empty and return an error if it is empty?

fpetrini15: I believe that should be handled here: https://github.com/triton-inference-server/common/pull/66/commits/efd0df9d0a2fac31918673b462cc3635f386ba37#diff-6215e044cfaca75f9bf9a1e1156e1cb21ce7085706b68b5d1b771ace1fc8f8fcR57-R67. Is this sufficient?

GuanLuo (approved):

src/tritonserver.cc
Tabrizian: I was wondering does setting the file to an empty string mean that we will be redirecting the logs to output? Do we need to document this behavior?

fpetrini15: No, empty string is the default. If the user enters the empty string, triton will redirect logging facilities to the console. It is documented here: https://github.com/triton-inference-server/server/pull/4640/files#diff-b2f94db5cc9d55373c593406216c0c9815a3bf8d10275fa60d76006346c988a3R71-R72

However, I can specifically mention the behavior of the empty string if you think it would be best?

src/tritonserver.cc
```diff
@@ -1215,6 +1215,20 @@ TRITONSERVER_ServerOptionsSetModelLoadThreadCount(
   return nullptr;  // Success
 }
 
+TRITONAPI_DECLSPEC TRITONSERVER_Error*
+TRITONSERVER_ServerOptionsSetLogFile(
+    TRITONSERVER_ServerOptions* options, const char* file)
+{
+#ifdef TRITON_ENABLE_LOGGING
+  std::string out_file(file);
+  LOG_SET_OUT_FILE(out_file);
```
GuanLuo, line 1224: Should check and return error if fail to open the file

include/triton/core/tritonserver.h
```diff
@@ -1761,6 +1761,14 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error*
 TRITONSERVER_ServerOptionsSetModelLoadThreadCount(
     TRITONSERVER_ServerOptions* options, unsigned int thread_count);
 
+/// Provide a log output file.
+///
+/// \param options The server options object.
```
Tabrizian, line 1766: It would be great to describe the behavior for empty string here too.

src/tritonserver.cc
Tabrizian: https://github.com/triton-inference-server/core/pull/98/files#r934647392

src/tritonserver.cc
fpetrini15: As discussed, an error cannot be returned. Will return bool instead.

GuanLuo: Can you point me to the discussion?

include/triton/core/tritonserver.h
```diff
@@ -1761,6 +1761,17 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error*
...
+/// \param file a string defining the file where the log outputs will be saved.
+/// An "empty" string for the file name will cause triton to direct logging
```
GuanLuo, line 1768: don't double quote empty, also check `nullptr` and log to console if `nullptr` is provided.

include/triton/core/tritonserver.h
```diff
@@ -1761,6 +1761,17 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error*
...
+/// facilities to the console (default behavior)
```
GuanLuo, line 1769: remove "(default behavior)"

include/triton/core/tritonserver.h
```diff
@@ -1761,6 +1761,17 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error*
...
+/// \return On success, this function will return on empty string. On failure,
+/// it will return a string containing the error message.
```
GuanLuo, line 1771:
```suggestion
/// \return a TRITONSERVER_Error indicating success or failure.
```

GuanLuo (approved):

PR #101, opened 2022-07-22
Core changes for Python BE Logging

fpetrini15:
User should be able to specify what level of verbosity their log is, not just 1.

### DIFF
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -27,7 +27,6 @@
 
 /// \file
 
-#include <string>
 #include <stdbool.h>
 #include <stddef.h>
 #include <stdint.h>
@@ -1768,7 +1767,7 @@ TRITONSERVER_ServerOptionsSetModelLoadThreadCount(
 /// \param file a string defining the file where the log outputs will be saved.
 /// \return a TRITONSERVER_Error indicating success or failure.
 TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerOptionsSetLogOutFile(
-    TRITONSERVER_ServerOptions* options, std::string file);
+    TRITONSERVER_ServerOptions* options, const char* file);
 
 /// Enable or disable info level logging.
 ///
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -27,6 +27,7 @@
 
 /// \file
 
+#include <string>
 #include <stdbool.h>
 #include <stddef.h>
 #include <stdint.h>
@@ -1761,7 +1762,7 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error*
 TRITONSERVER_ServerOptionsSetModelLoadThreadCount(
     TRITONSERVER_ServerOptions* options, unsigned int thread_count);
 
-/// Enable or disable info level logging.
+/// Provide a log output file.
 ///
 /// \param options The server options object.
 /// \param file a string defining the file where the log outputs will be saved.
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -1764,6 +1764,14 @@ TRITONSERVER_ServerOptionsSetModelLoadThreadCount(
 /// Enable or disable info level logging.
 ///
 /// \param options The server options object.
+/// \param file a string defining the file where the log outputs will be saved.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerOptionsSetLogOutFile(
+    TRITONSERVER_ServerOptions* options, std::string file);
+
+/// Enable or disable info level logging.
+///
+/// \param options The server options object.
 /// \param log True to enable info logging, false to disable.
 /// \return a TRITONSERVER_Error indicating success or failure.
 TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerOptionsSetLogInfo(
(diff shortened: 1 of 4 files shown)


src/tritonserver.cc
```diff
@@ -646,7 +646,7 @@ TRITONSERVER_LogIsEnabled(TRITONSERVER_LogLevel level)
 TRITONAPI_DECLSPEC TRITONSERVER_Error*
 TRITONSERVER_LogMessage(
```
dyastremsky, line 647: You're going to want to be really careful here. If you're changing API functions, you have to update the minor version. But this seems like a breaking change (i.e. would need to be a major version update, which you want to avoid for a small change), given that it changes how the API is called even with a default value. For example, in stubs. It may make more sense to add a separate function.

src/tritonserver.cc
```diff
@@ -646,7 +646,7 @@ TRITONSERVER_LogIsEnabled(TRITONSERVER_LogLevel level)
...
     TRITONSERVER_LogLevel level, const char* filename, const int line,
-    const char* msg)
+    const char* msg, const uint32_t verbosity)
```
dyastremsky, line 649: You're going to want to be really careful here. If you're changing API functions, you have to update the minor version. But this seems like a breaking change (i.e. would need to be a major version update, which you want to avoid for a small change), given that it changes how the API is called even with a default value. For example, in stubs. It may make more sense to add a separate function.

dyastremsky (changes requested):

src/tritonserver.cc
fpetrini15: I also don't like the use of a default value here and first was going to add a separate function. However, I felt that it wouldn't be the best fix. Down the line I figured it would add function clutter when the original function itself would be more robust if it took in a verbosity parameter since it has taken the responsibility of logging verbose messages. This way we could keep logging centralized in a single, main function. I figured the default parameter would add time for us to either specify this parameter wherever it is called from or extract verbose logging responsibility from this function and place it into a new one, depending on the direction we want to go.

fpetrini15: Ultimately, I believe, for now, APIs could be called in the exact same way, but I would have to test. Eventually they would need to have an additional parameter specified, unless we want to go the direction of extracting/creating a new verbose logging method.

dyastremsky: Logic makes sense, just make sure it passes the CI pipeline (including running some related tests like one of the perf_analyzer ones, as I believe they require manual trigger). That should hopefully cover most cases. CC: @GuanLuo for whether the minor version needs to be incremented and adding a new default parameter to an API call seems non-breaking.

Tabrizian: As David mentioned, you cannot add arguments to the current API functions since this will break the C ABI compatibility.
There is an option to set the logging level to verbose but it is global: https://github.com/triton-inference-server/core/blob/28b02988b65ce2a0092def59306de04c0a2c607f/src/tritonserver.cc#L1268-L1269

Also, this is a C-API which means that many of the functionalities we rely on for C++ such default values are not available.

include/triton/core/tritonserver.h
```diff
@@ -1761,6 +1761,14 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error*
 TRITONSERVER_ServerOptionsSetModelLoadThreadCount(
     TRITONSERVER_ServerOptions* options, unsigned int thread_count);
 
+/// Provide a log output file.
+///
+/// \param options The server options object.
+/// \param file a string defining the file where the log outputs will be saved.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ServerOptionsSetLogOutFile(
```
Tabrizian, line 1769: `TRITONSERVER_ServerOptionsSetLogOutFile` -> `TRITONSERVER_ServerOptionsSetLogFile`

include/triton/core/tritonserver.h
```diff
@@ -286,7 +286,7 @@ TRITONSERVER_DECLSPEC bool TRITONSERVER_LogIsEnabled(
 /// \return a TRITONSERVER_Error indicating success or failure.
 TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_LogMessage(
     TRITONSERVER_LogLevel level, const char* filename, const int line,
-    const char* msg);
+    const char* msg, const uint32_t verbosity=1);
```
Tabrizian, line 289: Remove the verbosity based on the discussions below.

Tabrizian (changes requested):

src/tritonserver.cc
fpetrini15: Okay, I will definitely not add an extra parameter to that function then. However, the issue that remains is that it contains the following line: `LOG_VERBOSE_FL(1, filename, line) << msg;`. This means that if the user attempts to log a verbose level 3 message and global setting is level 1, it will print anyways, which the user will not expect. We can use the global function like you mention, but then we need to keep track of the previous value and pop it back. To remedy this more simply, we can create a new function which passes the requested verbosity level, but now we have some functional clutter. Is this an acceptable amount of clutter?

GuanLuo: I agree with Iman and David that this is actually changing the API which can break backward compatibility. I believe default argument is a compiler step and thus if a backend is compiled before will not work as it calls the API with one-less argument. You will have to add new API if the backend really want to log verbose message with different level.

I incline to not add new API until it is explicitly be requested by user, which implies that the Python backend shouldn't expose verbose level for now. It won't be too much future work if the backend ends up needing verbose level as new API must be added for backward compatibility reason. And specifically for Python backend, the interface can be extended easily as we just add new default argument which shouldn't have backward compatibility issue. @Tabrizian does the assessment on Python make sense to you?

include/triton/core/tritonserver.h
GuanLuo: I don't think this change belongs to this PR, should remove.

src/tritonserver.cc
Tabrizian: I agree with Guan. We can remove `verbosity` argument from the Python API until there is an explicit request for it. The current Python API has room for this improvement.

fpetrini15:
Closing PR on the basis of the discussion above. Verbosity will not be exposed in the python backend until requested.

PR #103, opened 2022-07-25
Add cloud storage credential management

kthui:
Adds the ability to load cloud storage credentials aggregated into a single JSON file. The [server PR](https://github.com/triton-inference-server/server/pull/4695) provides some documentations on how the file can be used.

### DIFF
--- a/src/filesystem.cc
+++ b/src/filesystem.cc
@@ -70,9 +70,11 @@
 #include <stdlib.h>
 #include <sys/stat.h>
 #include <sys/types.h>
+
 #include <cerrno>
 #include <fstream>
 #include <mutex>
+
 #include "constants.h"
 #include "status.h"
 #include "triton/common/logging.h"
@@ -511,20 +513,20 @@ class GCSFileSystem final : public FileSystem {
       google::cloud::StatusOr<gcs::ObjectMetadata>* metadata);
 
   google::cloud::StatusOr<gcs::Client> client_;
-  static std::mutex env_mu;  // protect GOOGLE_APPLICATION_CREDENTIALS env var
+  static std::mutex env_mu_;  // protect GOOGLE_APPLICATION_CREDENTIALS env var
 };
-std::mutex GCSFileSystem::env_mu;
+std::mutex GCSFileSystem::env_mu_;
 
 GCSFileSystem::GCSFileSystem()
 {
-  std::lock_guard<std::mutex> lock(env_mu);
+  std::lock_guard<std::mutex> lock(env_mu_);
   client_ = gcs::Client::CreateDefaultClient();
 }
 
 GCSFileSystem::GCSFileSystem(const std::string& credential_path)
 {
   std::string env_name = "GOOGLE_APPLICATION_CREDENTIALS";
-  std::lock_guard<std::mutex> lock(env_mu);
+  std::lock_guard<std::mutex> lock(env_mu_);
 #ifdef _WIN32
   _putenv((env_name + "=" + credential_path).c_str());
 #else
@@ -917,7 +919,9 @@ const std::string AS_URL_PATTERN = "as://([^/]+)/([^/?]+)(?:/([^?]*))?(\\?.*)?";
 class ASFileSystem final : public FileSystem {
  public:
   ASFileSystem(const std::string& s3_path);
-  ASFileSystem(const std::string& s3_path, const char* account_str, const char* account_key);
+  ASFileSystem(
+      const std::string& s3_path, const char* account_str,
+      const char* account_key);
   Status CheckClient();
 
   Status FileExists(const std::string& path, bool* exists) override;
@@ -972,12 +976,17 @@ ASFileSystem::ParsePath(
   return Status::Success;
 }
 
-ASFileSystem::ASFileSystem(const std::string& path) : ASFileSystem(path, std::getenv("AZURE_STORAGE_ACCOUNT"), std::getenv("AZURE_STORAGE_KEY")) {}
-
-ASFileSystem::ASFileSystem(const std::string& path, const char* account_str, const char* account_key) : as_regex_(AS_URL_PATTERN)
+ASFileSystem::ASFileSystem(const std::string& path)
+    : ASFileSystem(
+          path, std::getenv("AZURE_STORAGE_ACCOUNT"),
+          std::getenv("AZURE_STORAGE_KEY"))
 {
-  //const char* account_str = std::getenv("AZURE_STORAGE_ACCOUNT");
-  //const char* account_key = std::getenv("AZURE_STORAGE_KEY");
+}
+
+ASFileSystem::ASFileSystem(
+    const std::string& path, const char* account_str, const char* account_key)
+    : as_regex_(AS_URL_PATTERN)
+{
   std::shared_ptr<as::storage_account> account = nullptr;
   std::string host_name, container, blob_path, query;
   if (RE2::FullMatch(
@@ -1318,7 +1327,9 @@ namespace s3 = Aws::S3;
 class S3FileSystem final : public FileSystem {
  public:
   S3FileSystem(const std::string& s3_path);
-  S3FileSystem(const std::string& s3_path, const char* secret_key, const char* key_id, const char* region, const char* session_token);
+  S3FileSystem(
+      const std::string& s3_path, const char* secret_key, const char* key_id,
+      const char* region, const char* session_token);
   ~S3FileSystem();
 
   Status CheckClient(const std::string& s3_path);
@@ -1350,14 +1361,14 @@ class S3FileSystem final : public FileSystem {
       const std::string& path, std::string* bucket, std::string* object);
   Status CleanPath(const std::string& s3_path, std::string* clean_path);
   const Aws::SDKOptions options_;
-  s3::S3Client* client_;  // must NOT init until Aws::InitAPI is called
+  std::unique_ptr<s3::S3Client> client_;  // init after Aws::InitAPI is called
   re2::RE2 s3_regex_;
   // enforce order: Aws::InitAPI -> Aws::ShutdownAPI -> Aws::InitAPI ...
-  static std::mutex init_mu;
-  static size_t init_count;
+  static std::mutex init_mu_;
+  static size_t init_count_;
 };
-std::mutex S3FileSystem::init_mu;
-size_t S3FileSystem::init_count = 0;
+std::mutex S3FileSystem::init_mu_;
+size_t S3FileSystem::init_count_ = 0;
 
 Status
 S3FileSystem::ParsePath(
@@ -1458,20 +1469,27 @@ S3FileSystem::CleanPath(const std::string& s3_path, std::string* clean_path)
   return Status::Success;
 }
 
-S3FileSystem::S3FileSystem(const std::string& s3_path) : S3FileSystem(s3_path, std::getenv("AWS_SECRET_ACCESS_KEY"), std::getenv("AWS_ACCESS_KEY_ID"), std::getenv("AWS_DEFAULT_REGION"), std::getenv("AWS_SESSION_TOKEN")) {}
+S3FileSystem::S3FileSystem(const std::string& s3_path)
+    : S3FileSystem(
+          s3_path, std::getenv("AWS_SECRET_ACCESS_KEY"),
+          std::getenv("AWS_ACCESS_KEY_ID"), std::getenv("AWS_DEFAULT_REGION"),
+          std::getenv("AWS_SESSION_TOKEN"))
+{
+}
 
 S3FileSystem::S3FileSystem(
-    const std::string& s3_path, const char* secret_key, const char* key_id, const char* region, const char* session_token)
+    const std::string& s3_path, const char* secret_key, const char* key_id,
+    const char* region, const char* session_token)
     : s3_regex_(
           "s3://(http://|https://|)([0-9a-zA-Z\\-.]+):([0-9]+)/"
           "([0-9a-z.\\-]+)(((/[0-9a-zA-Z.\\-_]+)*)?)")
 {
   // init aws API if not already
-  init_mu.lock();
-  if (!init_count++) {
+  init_mu_.lock();
+  if (!init_count_++) {
     Aws::InitAPI(options_);
   }
-  init_mu.unlock();
+  init_mu_.unlock();
 
   Aws::Client::ClientConfiguration config;
   Aws::Auth::AWSCredentials credentials;
@@ -1510,27 +1528,25 @@ S3FileSystem::S3FileSystem(
   }
 
   if ((secret_key != NULL) && (key_id != NULL)) {
-    client_ = new s3::S3Client(
+    client_.reset(new s3::S3Client(
         credentials, config,
         Aws::Client::AWSAuthV4Signer::PayloadSigningPolicy::Never,
-        /*useVirtualAdressing*/ false);
+        /*useVirtualAdressing*/ false));
 
   } else {
-    client_ = new s3::S3Client(
+    client_.reset(new s3::S3Client(
         config, Aws::Client::AWSAuthV4Signer::PayloadSigningPolicy::Never,
-        /*useVirtualAdressing*/ false);
+        /*useVirtualAdressing*/ false));
   }
 }
 
 S3FileSystem::~S3FileSystem()
 {
-  delete client_;
   // shutdown aws API if no more active instance
-  init_mu.lock();
-  if (!--init_count) {
+  std::lock_guard<std::mutex> lock(init_mu_);
+  if (!--init_count_) {
     Aws::ShutdownAPI(options_);
   }
-  init_mu.unlock();
 }
 
 Status
@@ -1963,77 +1979,67 @@ class FileSystemManager {
   // create a FileSystem from the given type
   Status GetFileSystem(FileSystemType type, FileSystem** file_system);
 
-  FileSystemManager();
-  ~FileSystemManager();
-
  private:
-  static Status LoadCredentialFromFile();
-  static Status LoadCredential();
-  template<class FileSystemCredential> static Status GetLongestMatchingCredential(const std::vector<std::pair<std::string, FileSystemCredential>>& credentials, const std::string& file_system_prefix, const std::string& path, FileSystemCredential& credential);
+  Status ReturnErrorOrReload(
+      const Status& load_status, const Status& error_status,
+      const std::string& path, FileSystem** file_system);
 
-  static std::recursive_mutex mu_;  // for protecting concurrent access into static variables
-  static bool is_credential_cached;
+  static Status LoadCredential(bool flush_cache = false);
+
+  template <class FileSystemCredential>
+  static void SortCredential(
+      std::vector<std::pair<std::string, FileSystemCredential>>& credential);
+  template <class FileSystemCredential>
+  static Status GetLongestMatchingCredential(
+      const std::vector<std::pair<std::string, FileSystemCredential>>&
+          credentials,
+      const std::string& file_system_prefix, const std::string& path,
+      FileSystemCredential& credential);
+
+  // for protecting concurrent access into static variables
+  static std::mutex mu_;
+  static bool is_credential_cached_;
   // credentials should be sorted in descending name length order
 #ifdef TRITON_ENABLE_GCS
-  static std::vector<std::pair<std::string, std::string>> gs_credential;
-  GCSFileSystem* gcs_fs;
+  static std::vector<std::pair<std::string, std::string>> gs_credential_;
+  std::unique_ptr<GCSFileSystem> gcs_fs_;
 #endif  // TRITON_ENABLE_GCS
 #ifdef TRITON_ENABLE_S3
-  static std::vector<std::pair<std::string, std::tuple<std::string, std::string, std::string, std::string>>> s3_credential;
-  S3FileSystem* s3_fs;
+  static std::vector<std::pair<
+      std::string,
+      std::tuple<std::string, std::string, std::string, std::string>>>
+      s3_credential_;
+  std::unique_ptr<S3FileSystem> s3_fs_;
 #endif  // TRITON_ENABLE_S3
 #ifdef TRITON_ENABLE_AZURE_STORAGE
-  static std::vector<std::pair<std::string, std::pair<std::string, std::string>>> as_credential;
-  ASFileSystem* as_fs;
+  static std::vector<
+      std::pair<std::string, std::pair<std::string, std::string>>>
+      as_credential_;
+  std::unique_ptr<ASFileSystem> as_fs_;
 #endif  // TRITON_ENABLE_AZURE_STORAGE
-  LocalFileSystem* local_fs;
+  std::unique_ptr<LocalFileSystem> local_fs_;
 };
 // initialize FileSystemManager static variables
-std::recursive_mutex FileSystemManager::mu_;
-bool FileSystemManager::is_credential_cached = false;
+std::mutex FileSystemManager::mu_;
+bool FileSystemManager::is_credential_cached_ = false;
 #ifdef TRITON_ENABLE_GCS
-std::vector<std::pair<std::string, std::string>> FileSystemManager::gs_credential;
+std::vector<std::pair<std::string, std::string>>
+    FileSystemManager::gs_credential_;
 #endif  // TRITON_ENABLE_GCS
 #ifdef TRITON_ENABLE_S3
-std::vector<std::pair<std::string, std::tuple<std::string, std::string, std::string, std::string>>> FileSystemManager::s3_credential;
+std::vector<std::pair<
+    std::string,
+    std::tuple<std::string, std::string, std::string, std::string>>>
+    FileSystemManager::s3_credential_;
 #endif  // TRITON_ENABLE_S3
 #ifdef TRITON_ENABLE_AZURE_STORAGE
-std::vector<std::pair<std::string, std::pair<std::string, std::string>>> FileSystemManager::as_credential;
+std::vector<std::pair<std::string, std::pair<std::string, std::string>>>
+    FileSystemManager::as_credential_;
 #endif  // TRITON_ENABLE_AZURE_STORAGE
 
-FileSystemManager::FileSystemManager()
-{
-#ifdef TRITON_ENABLE_GCS
-  gcs_fs = NULL; 
-#endif  // TRITON_ENABLE_GCS
-#ifdef TRITON_ENABLE_S3
-  s3_fs = NULL;
-#endif  // TRITON_ENABLE_S3
-#ifdef TRITON_ENABLE_AZURE_STORAGE
-  as_fs = NULL;
-#endif  // TRITON_ENABLE_AZURE_STORAGE
-  local_fs = NULL;
-}
-
-FileSystemManager::~FileSystemManager()
-{
-#ifdef TRITON_ENABLE_GCS
-  if (gcs_fs != NULL)
-    delete gcs_fs;
-#endif  // TRITON_ENABLE_GCS
-#ifdef TRITON_ENABLE_S3
-  if (s3_fs != NULL)
-    delete s3_fs;
-#endif  // TRITON_ENABLE_S3
-#ifdef TRITON_ENABLE_AZURE_STORAGE
-  if (as_fs != NULL)
-    delete as_fs;
-#endif  // TRITON_ENABLE_AZURE_STORAGE
-  if (local_fs != NULL)
-    delete local_fs;
-}
-
-Status FileSystemManager::GetFileSystem(const std::string& path, FileSystem** file_system)
+Status
+FileSystemManager::GetFileSystem(
+    const std::string& path, FileSystem** file_system)
 {
   // Check if this is a GCS path (gs://$BUCKET_NAME)
   if (!path.empty() && !path.rfind("gs://", 0)) {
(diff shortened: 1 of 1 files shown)


src/filesystem.cc
```diff
@@ -1898,8 +1970,115 @@ S3FileSystem::DeleteDirectory(const std::string& path)
...
+  static void SortCredential(
+      std::vector<std::pair<std::string, FileSystemCredential>>& credential);
+  template <class FileSystemCredential>
+  static Status GetLongestMatchingCredential(
+      const std::vector<std::pair<std::string, FileSystemCredential>>&
+          credentials,
+      const std::string& file_system_prefix, const std::string& path,
+      FileSystemCredential& credential);
+
+  static std::recursive_mutex
```
rmccorm4, line 2005: Can this be done with a `std::mutex` instead?

src/filesystem.cc
```diff
@@ -1474,21 +1528,39 @@ S3FileSystem::S3FileSystem(
...
         /*useVirtualAdressing*/ false);
   }
 }
 
 S3FileSystem::~S3FileSystem()
 {
-  Aws::ShutdownAPI(options_);
+  delete client_;
+  // shutdown aws API if no more active instance
+  init_mu.lock();
```
rmccorm4, line 1547: Can use `std::lock_guard` instead

src/filesystem.cc
```diff
@@ -1941,37 +2166,244 @@ GetFileSystem(const std::string& path, FileSystem** file_system)
...
+      case FileSystemType::GCS:
+        return GetFileSystem("gs://", file_system);
+      case FileSystemType::S3:
+        return GetFileSystem("s3://", file_system);
+      case FileSystemType::AS:
+        return GetFileSystem("as://", file_system);
+      default:
+        return Status(
+            Status::Code::UNSUPPORTED,
+            "The requested filesteam cannot be accessed by type");
```
rmccorm4, line 2222:
```suggestion
            "Unsupported filesystem type");
```

src/filesystem.cc
```diff
@@ -1941,37 +2166,244 @@ GetFileSystem(const std::string& path, FileSystem** file_system)
...
+    // with env var, only LOCAL and GCS are not path-dependent
+    switch (type) {
+      case FileSystemType::LOCAL:
+        return GetFileSystem("", file_system);
+      case FileSystemType::GCS:
+        return GetFileSystem("gs://", file_system);
+      default:
+        return Status(
+            Status::Code::UNSUPPORTED,
+            "The requested filesteam cannot be accessed by type");
```
rmccorm4, line 2234:
```suggestion
            "Unsupported filesystem type");
```

src/filesystem.cc
```diff
@@ -1941,37 +2166,244 @@ GetFileSystem(const std::string& path, FileSystem** file_system)
...
+    }
+  } else if (status_.Message() == "Use legacy credential") {
+    // with env var, only LOCAL and GCS are not path-dependent
+    switch (type) {
+      case FileSystemType::LOCAL:
+        return GetFileSystem("", file_system);
+      case FileSystemType::GCS:
+        return GetFileSystem("gs://", file_system);
+      default:
```
rmccorm4, line 2231: Maybe return specific error messages here on S3/Azure for clearer errors

src/filesystem.cc
```diff
@@ -1941,37 +2166,244 @@ GetFileSystem(const std::string& path, FileSystem** file_system)
...
+  }
+  return status_;
+}
+
+Status
+FileSystemManager::ReturnErrorOrReload(
+    const Status& load_status, const Status& error_status,
+    const std::string& path, FileSystem** file_system)
+{
+  if (load_status.Message() == "Cached") {
```
rmccorm4, line 2245: Maybe make use of `Status::Code::ALREADY_EXISTS` so this is less dependent on specific message wording if changed?

src/filesystem.cc
kthui: Yes. Updated.

src/filesystem.cc
kthui: `LoadCredential` will call `LoadCredentialFromFile` which both hold the same mutex. It might be easier to use `recursive_mutex` to avoid the same thread from locking on the same mutex twice. 

In terms of making the locking mechanism simple, I think it depends on whether we want to protect the `bool is_credential_cached` variable. Because if credential is not cached, then it is loaded from disk. By protecting this check, we prevents a possible duplicate load from disk from concurrent calls into `LoadCredential` or `LoadCredentialFromDisk`, and force concurrent calls to use cached results. On the other hand, if credential is cached, it terminates right away and release the lock, which should not be any significant reduction in parallism. I am not not sure if there is an easier way to protect this corner case?

src/filesystem.cc
kthui: Thanks for catching this. Updated.

src/filesystem.cc
kthui: Yes, good suggestion. Updated.

src/filesystem.cc
kthui: Updated.

src/filesystem.cc
kthui: Yes. Updated.

src/filesystem.cc
rmccorm4: It looks like LoadCredentialFromFile is only called directly [here](https://github.com/triton-inference-server/core/blob/84429c6f2cca2181e9a92e3240f54a65bc662b62/src/filesystem.cc#L2248), so maybe you could tweak the logic so `LoadCredential` is the only function ever called, and it calls `LoadCredentialFromFile` internally with the lock already held?

Or something along these lines, there can be a documented expectation that the lock must be held when calling this method, etc.

src/filesystem.cc
```diff
@@ -1898,8 +1970,115 @@ S3FileSystem::DeleteDirectory(const std::string& path)
...
+  template <class FileSystemCredential>
+  static Status GetLongestMatchingCredential(
+      const std::vector<std::pair<std::string, FileSystemCredential>>&
+          credentials,
+      const std::string& file_system_prefix, const std::string& path,
+      FileSystemCredential& credential);
+
+  static std::recursive_mutex
+      mu_;  // for protecting concurrent access into static variables
+  static bool is_credential_cached;
```
rmccorm4, line 2007: Add an `_` underscore after all class member variable names.

src/filesystem.cc
kthui: The `LoadCredential` will only return status SUCCESS, ALREADY_EXISTS or UNAVAILABLE. Except SUCCESS, all dependencies do not return those code, so the code is changed to not depends on any message wording.

src/filesystem.cc
```diff
@@ -70,12 +70,21 @@
...
+#include <mutex>
+
 #include "constants.h"
 #include "status.h"
 #include "triton/common/logging.h"
 
+#define TRITONJSON_STATUSTYPE triton::core::Status
+#define TRITONJSON_STATUSRETURN(M) \
+  return triton::core::Status(triton::core::Status::Code::INTERNAL, (M))
+#define TRITONJSON_STATUSSUCCESS triton::core::Status::Success
```
rmccorm4, line 85: Are these being used?

src/filesystem.cc
```diff
@@ -2242,7 +2250,7 @@ FileSystemManager::ReturnErrorOrReload(
     const Status& load_status, const Status& error_status,
     const std::string& path, FileSystem** file_system)
 {
-  if (load_status.Message() == "Cached") {
+  if (load_status.ErrorCode() == Status::Code::ALREADY_EXISTS) {
```
rmccorm4, line 2253: I think these should be `StatusCode()` instead of `ErrorCode()` everywhere

src/filesystem.cc
```diff
@@ -1941,37 +2167,251 @@ GetFileSystem(const std::string& path, FileSystem** file_system)
...
-  switch (type) {
-    case FileSystemType::LOCAL:
-      return GetFileSystem("", file_system);
-    case FileSystemType::GCS:
-      return GetFileSystem("gs://", file_system);
-    default:
-      return Status(
-          Status::Code::UNSUPPORTED,
-          "The requested filesteam can not be accessed by type");
+  const Status& status_ = LoadCredential();
```
rmccorm4, line 2209: don't use underscore for local variable, use underscores for class variables

```suggestion
  const Status& status = LoadCredential();
```

kthui: Yes. Corrected.

src/filesystem.cc
kthui: Yes. Updated.

src/filesystem.cc
kthui: Corrected.

src/filesystem.cc
kthui: Yes, removing them will cause compiler error. This is just how other files import `triton_json.h`.

src/filesystem.cc
kthui: I see what you means, grouped them into a single `LoadCredential`.

src/filesystem.cc
```diff
@@ -2001,78 +2000,77 @@ class FileSystemManager {
...
 #ifdef TRITON_ENABLE_S3
-  s3_fs = NULL;
+  s3_fs_ = NULL;
 #endif  // TRITON_ENABLE_S3
 #ifdef TRITON_ENABLE_AZURE_STORAGE
-  as_fs = NULL;
+  as_fs_ = NULL;
 #endif  // TRITON_ENABLE_AZURE_STORAGE
-  local_fs = NULL;
+  local_fs_ = NULL;
```
rmccorm4, line 2055: Can these be smart pointers to be cleaned up automatically?

src/filesystem.cc
```diff
@@ -2001,78 +2000,77 @@ class FileSystemManager {
...
 #ifdef TRITON_ENABLE_AZURE_STORAGE
-  if (as_fs != NULL)
-    delete as_fs;
+  if (as_fs_ != NULL)
+    delete as_fs_;
 #endif  // TRITON_ENABLE_AZURE_STORAGE
-  if (local_fs != NULL)
-    delete local_fs;
+  if (local_fs_ != NULL)
+    delete local_fs_;
```
rmccorm4, line 2073: Can these be smart pointers instead to be cleaned up automatically?

src/filesystem.cc
rmccorm4, line 2055: Use `nullptr`

src/filesystem.cc
```diff
@@ -2126,33 +2124,33 @@ FileSystemManager::GetFileSystem(
...
       if (!match_status.IsOk()) {
         return ReturnErrorOrReload(
             load_status, match_status, path, file_system);
       }
-      s3_fs = new S3FileSystem(
+      s3_fs_ = new S3FileSystem(
           path, std::get<0>(credential_).c_str(),
           std::get<1>(credential_).c_str(), std::get<2>(credential_).c_str(),
           std::get<3>(credential_).size() ? std::get<3>(credential_).c_str()
                                           : NULL);
```
rmccorm4, line 2142: `nullptr` over `NULL`

src/filesystem.cc
```diff
@@ -2242,144 +2240,138 @@ FileSystemManager::GetFileSystem(FileSystemType type, FileSystem** file_system)
...
+  std::lock_guard<std::mutex> lock(mu_);
 
-    triton::common::TritonJson::Value cred_json;
+  // check if credential is already cached
+  if (is_credential_cached_ && !flush_cache) {
+    return Status(Status::Code::ALREADY_EXISTS, "Cached");
+  }
+
+  const char* file_path_c_str = std::getenv("TRITON_CLOUD_CREDENTIAL_PATH");
+  if (file_path_c_str != NULL) {
```
rmccorm4, line 2274: `nullptr` over `NULL`

src/filesystem.cc
kthui: Yes. Switched to `unique_ptr`.

src/filesystem.cc
kthui: Switched to `unique_ptr`, no longer applicable.

src/filesystem.cc
kthui: Updated.

src/filesystem.cc
kthui: Updated.

src/filesystem.cc
```diff
@@ -1898,8 +1969,113 @@ S3FileSystem::DeleteDirectory(const std::string& path)
...
+  if (s3_fs_ != NULL)
+    delete s3_fs_;
+#endif  // TRITON_ENABLE_S3
+#ifdef TRITON_ENABLE_AZURE_STORAGE
+  if (as_fs_ != NULL)
+    delete as_fs_;
+#endif  // TRITON_ENABLE_AZURE_STORAGE
+  if (local_fs_ != NULL)
+    delete local_fs_;
+}
```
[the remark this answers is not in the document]
kthui: Updated. The file system will be cached.

src/filesystem.cc
```diff
@@ -1898,8 +1969,113 @@ S3FileSystem::DeleteDirectory(const std::string& path)
...
+#ifdef TRITON_ENABLE_S3
+std::vector<std::pair<
+    std::string,
+    std::tuple<std::string, std::string, std::string, std::string>>>
+    FileSystemManager::s3_credential_;
+#endif  // TRITON_ENABLE_S3
+#ifdef TRITON_ENABLE_AZURE_STORAGE
+std::vector<std::pair<std::string, std::pair<std::string, std::string>>>
+    FileSystemManager::as_credential_;
+#endif  // TRITON_ENABLE_AZURE_STORAGE
```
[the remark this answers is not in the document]
kthui: Updated. There will only be one file system manager object.

kthui:
There is a greater change in architecture. Previously, the file system manager only handles credentials from file, and let each file system to either initialize from env var or from provided credential.

Currently, the file system manager will handle loading from both credential file and env var, and forms a single credential cache. Therefore, each file system will only initialize from the provided credential, from credential manager. Each credential in cache may also cache its respective file system, specifically, to be cached when the file system is requested.

src/filesystem.cc
```diff
@@ -948,28 +969,29 @@ ASFileSystem::ParsePath(
...
       }
     } else {
-      account_name = std::string(account_str);
+      account_name = account_str;
     }
 
     std::shared_ptr<as::storage_credential> cred;
-    if (account_key != NULL) {
+    if (account_key.size()) {
       // Shared Key
```
rmccorm4, line 995: 1. I would double check this part, not sure which logic is intended.
```cpp
// Before - these 2 conditions were opposites
if (account_str == NULL)
...
if (account_key != NULL)
...

// After - two conditions are the same?
if (account_str.size())
...
if (account_key.size())
...
```
2. Use `string.empty()` instead of `string.size()` so it's a bit more clear for the reader what the condition checks

src/filesystem.cc
```diff
@@ -1291,10 +1313,13 @@ namespace s3 = Aws::S3;
 
 class S3FileSystem : public FileSystem {
  public:
-  S3FileSystem(const Aws::SDKOptions& options, const std::string& s3_path);
+  S3FileSystem(
+      const std::string& s3_path, const std::string& secret_key,
+      const std::string& key_id, const std::string& region,
+      const std::string& session_token, const std::string& profile_name);
```
rmccorm4, line 1319: Maybe encapsulate in a simple `S3Options` struct?

src/filesystem.cc
```diff
@@ -1473,22 +1507,46 @@ S3FileSystem::S3FileSystem(
     }
   }
 
-  if ((secret_key != NULL) && (key_id != NULL)) {
-    client_ = s3::S3Client(
+  if (secret_key.size() && key_id.size()) {
```
rmccorm4, line 1510: `!empty()` instead of `size()`, apply everywhere applicable

src/filesystem.cc
```diff
@@ -1473,22 +1507,46 @@ S3FileSystem::S3FileSystem(
...
+  // FIXME: Calling Aws::ShutdownAPI will result in segmentation fault.
+  //        Currently, the Aws::initAPI will be called at most once at the
+  //        constructor and zero calls to Aws::ShutdownAPI, which is not a
+  //        memory leak and resources will be reclaimed when triton stops.
+  /*
+  std::lock_guard<std::mutex> lock(init_mu_);
+  if (!--init_count_) {
+    Aws::ShutdownAPI(options_);
+  }
+  */
```
rmccorm4, line 1537: remove if no longer needed

src/filesystem.cc
```diff
@@ -1473,22 +1507,46 @@ S3FileSystem::S3FileSystem(
...
+}
+
+Status
+S3FileSystem::CheckClient(const std::string& s3_path)
+{
+  bool b;
+  if (!IsDirectory(s3_path, &b).IsOk()) {
```
rmccorm4, line 1544: Do you need to check the value of `b`? At a glance it looks like it's possible to be false even if status is success: https://github.com/triton-inference-server/core/blob/162a4fd7e039bc5904de021922b48b8aa6657bea/src/filesystem.cc#L1568 ?

src/filesystem.cc
```diff
@@ -1941,39 +2110,283 @@ GetFileSystem(const std::string& path, FileSystem** file_system)
...
+          s3_cred_profile_json.AsString(&s3_cred_profile);
+        s3_cache_.push_back(std::make_tuple(
+            s3_cred_name,
+            std::make_tuple(
+                s3_cred_secret_key, s3_cred_key_id, s3_cred_region,
+                s3_cred_session_token, s3_cred_profile),
+            std::shared_ptr<S3FileSystem>()));
+      }
+      SortCache(s3_cache_);
+    }
```
rmccorm4, line 2265: Can these be moved into internal helpers like `LoadGCSCredentials`, `LoadS3Crendentials`, etc. for brevity?

src/filesystem.cc
```diff
@@ -893,7 +912,9 @@ const std::string AS_URL_PATTERN = "as://([^/]+)/([^/?]+)(?:/([^?]*))?(\\?.*)?";
 
 class ASFileSystem : public FileSystem {
  public:
-  ASFileSystem(const std::string& s3_path);
+  ASFileSystem(
+      const std::string& path, const std::string& account_str,
+      const std::string& account_key);
```
rmccorm4, line 917: Maybe encapsulate in a simple `AzureStorageOptions` struct?

src/filesystem.cc
```diff
@@ -503,10 +513,19 @@ class GCSFileSystem : public FileSystem {
...
+GCSFileSystem::GCSFileSystem(const std::string& credential_path)
 {
+  static std::string env_name = "GOOGLE_APPLICATION_CREDENTIALS";
+  std::lock_guard<std::mutex> lock(env_mu_);
+#ifdef _WIN32
+  _putenv((env_name + "=" + credential_path).c_str());
+#else
+  setenv(env_name.c_str(), credential_path.c_str(), 1);
+#endif
   client_ = gcs::Client::CreateDefaultClient();
```
GuanLuo, line 529: There is no way to pass in the credential directly?

src/filesystem.cc
```diff
@@ -948,28 +969,29 @@ ASFileSystem::ParsePath(
...
-  const char* account_str = std::getenv("AZURE_STORAGE_ACCOUNT");
-  const char* account_key = std::getenv("AZURE_STORAGE_KEY");
   std::shared_ptr<as::storage_account> account = nullptr;
   std::string host_name, container, blob_path, query;
   if (RE2::FullMatch(
           path, as_regex_, &host_name, &container, &blob_path, &query)) {
     size_t pos = host_name.rfind(".blob.core.windows.net");
     std::string account_name;
-    if (account_str == NULL) {
+    if (account_str.size()) {
```
GuanLuo, line 983: Using size() seems wrong, evaluate to "true" if the string has content
```suggestion
    if (account_str.empty()) {
```

src/filesystem.cc
```diff
@@ -948,28 +969,29 @@ ASFileSystem::ParsePath(
...
         account_name = host_name;
       }
     } else {
-      account_name = std::string(account_str);
+      account_name = account_str;
     }
 
     std::shared_ptr<as::storage_credential> cred;
-    if (account_key != NULL) {
+    if (account_key.size()) {
```
GuanLuo, line 994:
```suggestion
    if (!account_key.empty()) {
```

src/filesystem.cc
```diff
@@ -1322,10 +1347,15 @@ class S3FileSystem : public FileSystem {
   Status ParsePath(
       const std::string& path, std::string* bucket, std::string* object);
   Status CleanPath(const std::string& s3_path, std::string* clean_path);
-  Aws::SDKOptions options_;
-  s3::S3Client client_;
+  const Aws::SDKOptions options_;
+  s3::S3Client* client_;  // init after Aws::InitAPI is called
   re2::RE2 s3_regex_;
+  // enforce order: Aws::InitAPI -> Aws::ShutdownAPI -> Aws::InitAPI ...
```
GuanLuo, line 1353: Elaborate on why

src/filesystem.cc
kthui: My bad. Thanks for catching. Corrected.

src/filesystem.cc
```diff
@@ -1427,32 +1457,36 @@ S3FileSystem::CleanPath(const std::string& s3_path, std::string* clean_path)
...
           "s3://(http://|https://|)([0-9a-zA-Z\\-.]+):([0-9]+)/"
           "([0-9a-z.\\-]+)(((/[0-9a-zA-Z.\\-_]+)*)?)")
 {
+  // init aws api if not already
+  init_mu_.lock();
+  if (!init_count_++) {
+    Aws::InitAPI(options_);
+  }
+  init_mu_.unlock();
+
```
GuanLuo, line 1473:
```suggestion
{
  // init aws api if not already
  std::lock_guard<std::mutex> lk(init_mu_);
  if (!init_count_++) {
    Aws::InitAPI(options_);
  }
}
```

src/filesystem.cc
kthui: Updated.

src/filesystem.cc
```diff
@@ -1473,22 +1507,46 @@ S3FileSystem::S3FileSystem(
...
   }
 }
 
 S3FileSystem::~S3FileSystem()
 {
-  Aws::ShutdownAPI(options_);
+  // should delete all client_ before shutdown
+  delete client_;
+  // shutdown aws api if no more S3 instances
+  // FIXME: Calling Aws::ShutdownAPI will result in segmentation fault.
```
GuanLuo, line 1528: Bug in AWS client? Is there an issue to track that?

src/filesystem.cc
GuanLuo: I think we should revert to  `std::call_once`, the Init / Shutdown seems to be a one-time call in respect to the process? Should isolate it from possible multiple S3 clients for different credentials

src/filesystem.cc
kthui: Yes. Moved to smart pointer for `client_`, and no more destructors.

src/filesystem.cc
kthui: Because this is only to check if the client can connect with S3, it do not matter whether it is a directory or not. Changed the code to check with S3 directly.

src/filesystem.cc
kthui: Updated.

src/filesystem.cc
kthui: Yes. Updated.

src/filesystem.cc
kthui: Yes. Updated.

src/filesystem.cc
kthui: Updated.

src/filesystem.cc
kthui: Yes. Updated.
The mapping from previous char NULL -> empty string, since credential cannot be empty.

src/filesystem.cc
kthui: Updated.

src/filesystem.cc
kthui: Because AWS requires its S3 client library to be called exactly like: 
```
#include <aws/core/Aws.h>
int main(int argc, char** argv)
{
   Aws::SDKOptions options;
   Aws::InitAPI(options);
   {
      // make your SDK calls here.
   }
   Aws::ShutdownAPI(options);
   return 0;
}
```
[source](https://docs.aws.amazon.com/sdk-for-cpp/v1/developer-guide/basic-use.html)

The `InitAPI` and `ShutdownAPI` are static. They should not be double called and should follow the order.

src/filesystem.cc
kthui: Yes. Updated.

src/filesystem.cc
kthui: The closest one was in 2019 ([link](https://github.com/aws/aws-sdk-cpp/issues/1263)), but was dismissed because of the calling order (`InitAPI`-> do things -> `ShutdownAPI`).

I suspect this is because the `ShutdownAPI` is called when the program is shutting down, and the order might be slightly different from [the required ordering](https://github.com/triton-inference-server/core/pull/103#discussion_r933497563).

src/filesystem.cc
kthui: Just found the way to pass credential via argument. Thanks for checking. Updated.

src/filesystem.cc
```diff
@@ -891,10 +935,40 @@ GCSFileSystem::DeleteDirectory(const std::string& path)
...
+  account_key_ = to_str(account_key);
+}
+
+ASCredential::ASCredential(triton::common::TritonJson::Value& cred_json)
+{
+  triton::common::TritonJson::Value account_str_json, account_key_json;
+  if (cred_json.Find("account_str", &account_str_json))
+    account_str_json.AsString(&account_str_);
+  if (cred_json.Find("account_key", &account_key_json))
+    account_key_json.AsString(&account_key_);
```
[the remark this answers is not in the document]
kthui: If the JSON contains syntax error, the JSON parser will catch it before reaching this construction steps, and prints for example `error: creating server: Internal - failed to parse the request JSON buffer: Missing a comma or '}' after an object member. at 279`.

If an expected JSON object is of a different type, the JSON parser will not crash and skip forward. Then, credentials cannot be matched and `error: creating server: Not found - Cannot match credential for path  s3://s3-bucket-41498862/models1` is printed for example.

If the JSON is missing some parts of a credential, this is tricky because the Azure construction allows either/both `account_str` and `account_key` to be empty, and will use "default" if it is empty. If the user misses a key or sets it to empty, it might still be valid. The same goes for the S3 construction.

If the credential is invalid, the `check_client()` will catch it, for example with [`log-verbose=1` and S3 `"region": 0`] prints the following:

```
I0802 06:04:24.262804 33401 filesystem.cc:2335] Using credential    for path  s3://s3-bucket-41498862/models1
I0802 06:04:28.612361 33401 metrics.cc:650] Collecting metrics for GPU 0: NVIDIA TITAN V
I0802 06:04:28.612790 33401 tritonserver.cc:2176] ...(skip the table printout)

I0802 06:04:28.612829 33401 server.cc:259] No server context available. Exiting immediately.
error: creating server: Internal - Unable to create S3 filesystem client. Check account credentials.
+ exit 1
```
Basically, accessing the S3 path `s3://s3-bucket-41498862/models1` is using the default/empty S3 credential and gets account credential error, which should signals for checking for the default S3 credential. The only possible issue is the user must remember to set `log-verbose` to print which credential is used for which path.

kthui:
Rebased.

GuanLuo (dismissed):

src/filesystem.cc
```diff
@@ -505,16 +540,17 @@ class GCSFileSystem : public FileSystem {
   google::cloud::StatusOr<gcs::Client> client_;
 };
 
-GCSFileSystem::GCSFileSystem()
+GCSFileSystem::GCSFileSystem(const GCSCredential& gs_cred)
 {
-  client_ = gcs::Client::CreateDefaultClient();
+  auto creds = gcs::oauth2::CreateServiceAccountCredentialsFromJsonFilePath(gs_cred.path_);
+  if (creds) {
```
rmccorm4, line 546: what happens for invalid creds?

rmccorm4 (approved):

rmccorm4 (approved):
Just need some new tests in server PR

kthui:
Rebased.

src/filesystem.cc
kthui: The `FileSystemManager` will always call `CheckClient` before returning the file system. If the check fails, it reloads the credential and retry. If the check also fails after reload, the error from `CheckClient` is returned and the server gracefully terminates. It will print something like "Unable to create GCS client. Check account credentials."

kthui:
Rebased.

rmccorm4:
What changed since approvals? Hard to tell with force pushes.

kthui:
No change, rebased only.
```
$ git diff e9023eb2316a13397a4cc10b1fb2dafc631cf8ad de6dcd6a9a2bc6192e52e8432a08f30599caab82 --name-only
include/triton/core/tritonserver.h
src/backend_config.cc
src/backend_config.h
src/backend_model.cc
src/backend_model_instance.cc
src/backend_model_instance.h
src/cuda_utils.cc
src/cuda_utils.h
src/model_config_utils.cc
src/pinned_memory_manager.cc
src/response_cache.cc
src/test/response_cache_test.cc
src/tritonserver.cc
src/tritonserver_stub.cc
```
Notice none of the file change involves `filesystem.cc`

rmccorm4:
For future reference, I think you should be able to rebase without force pushing and it won't discard approvals

rmccorm4 (approved):

GuanLuo (approved):

PR #105, opened 2022-07-28
Add GPU memory limit check during mode loading

GuanLuo:


src/backend_config.h
```diff
@@ -68,4 +68,10 @@ Status BackendConfigurationSpecializeBackendName(
 Status BackendConfigurationBackendLibraryName(
     const std::string& backend_name, std::string* libname);
 
+/// Get GPU memory limit for model loading in fraction
```
rmccorm4, line 71:
```suggestion
/// Get GPU memory limit fraction for model loading
```

src/backend_model_instance.cc
```diff
@@ -235,6 +238,29 @@ TritonModelInstance::CreateInstances(
...
+        // there is sufficient GPU memory for other use.
+        if (kind == TRITONSERVER_INSTANCEGROUPKIND_GPU) {
+          size_t free, total;
+          double memory_limit;
+          RETURN_IF_ERROR(GetDeviceMemoryInfo(id, &free, &total));
+          RETURN_IF_ERROR(BackendConfigurationModelLoadGpuFraction(
+              backend_cmdline_config_map, id, &memory_limit));
+          const size_t allow = total * memory_limit;
+          const size_t used = total - free;
+          if (used > allow) {
```
rmccorm4, line 254: If I'm reading this correctly, say my GPU is currently using 60% of it's memory for various cuda processes or Triton inferences, etc.

If my `--model-load-gpu-limit 0:0.3"` limit is 30%, won't this fail even though there is 40% gpu available? I would think as long as model loading doesn't exceed 30%, this is OK. But this reads like if memory usage is > 30% then I can't load a model even if 30% is available.

GuanLuo: the limit is in respect to the total GPU memory usage, so 0.3 means we reject any loading on the GPU if memory usage > 30% regardless of the reason of allocation. Relative measurement may not work well with the issue to be solved, like the scenario you purposed, it can be configured that Triton will still load the model and hit 99% of memory usage to fail future inferences. But I think this check should be deferred to after the instance is created, using after-measurement seems to be a little bit more robust

Tabrizian: I was wondering do we load model instances in parallel? If yes, does this mean that for example, if the GPU limit is set to 99% and model A, and model B check the GPU memory usage before loading and the GPU memory is 90% but after loading both models require 10% more GPU memory which leads to overshooting since both of them will be loaded.

It looks like using this feature, requires a knowledge about the memory usage of the models beforehand. Is my understanding correct? For example, if the largest model is going to consume 70% of memory, you need to set the limit to 30% if you want to avoid OOM?

GuanLuo: > I was wondering do we load model instances in parallel? If yes, does this mean that for example, if the GPU limit is set to 99% and we are loading model A, and model B. They check the GPU memory usage before loading and the GPU memory is 90% but after loading each model require 10% more GPU memory which leads to overshooting since both of them will be loaded.

Currently we load instances of the same model one by one, but we do load different models in parallel. I don't think we can oversubscribe GPU memory (maybe we can with UVM?), I would think either or both models will receive error from the framework due to OOM. Now that I moved the check to after instance creation, both models should failed because hitting the check means the instance has created and memory usage will be 90+10 > 99.

> Also, it looks like using this feature requires a knowledge about the memory usage of the models beforehand. Is my understanding correct? For example, if the largest model is going to consume 70% of memory, you need to set the limit to 30% if you want to avoid OOM?

Yes if "memory usage" refers to usage other than loading the model (i.e. extra allocation for inference). Then your example is correct, if the model will require 70% on inference, then setting limit to 30% should help the inference to go through.

src/cuda_utils.cc
```diff
@@ -41,6 +41,42 @@ MemcpyHost(void* args)
...
+    }
+  }
+
+  if (cuerr == cudaSuccess) {
+    cuerr = cudaMemGetInfo(free, total);
+  }
+
+  if (overridden) {
+    cudaSetDevice(current_device);
+  }
```
rmccorm4, line 68: Should set `cuerr = cudaSetDevice` here and check

src/cuda_utils.cc
```diff
@@ -41,6 +41,42 @@ MemcpyHost(void* args)
...
+
+  if (cuerr != cudaSuccess) {
+    return Status(
+        Status::Code::INTERNAL,
+        (std::string("unable to get memory info for device ") +
+         std::to_string(device_id) + ": " + cudaGetErrorString(cuerr)));
+  }
```
rmccorm4, line 75: should move this after `cudaMemGetInfo` block to avoid overwriting `cuerr` with `cuerr = cudaSetDevice()`

src/backend_model_instance.cc
Tabrizian: >Now that I moved the check to after instance creation, both models should failed because hitting the check means the instance has created and memory usage will be 90+10 > 99.

I think my example didn't fully capture this. Let's change the number to 89 instead of 90. Just to make sure, is the main idea for this flag to prevent loading too many models that might result in failure when performing inference? To me it feels like we need to have a mutex or serialize model loadings to make sure we are reading the correct numbers for GPU memory usage.

src/cuda_utils.cc
GuanLuo: There is nothing we an do if can't revert the override, just handle it passively here.

src/backend_model_instance.cc
GuanLuo: You are correct that we are not getting the accurate measurement here, the usage obtained here is "over-estimated" for a given model instance when parallel loading is ongoing. There is no good way to get accurate instance memory usage except for serializing model loading which is not desired, so we will pessimistically reject more model loads when the usage is on the edge (i.e. your example can fit one more but we will likely reject both). I wouldn't worry about this too much as it is an edge case and user can work around it by loading the model one by one via load API (when usage is approaching the limit)

src/cuda_utils.h
```diff
@@ -50,6 +50,13 @@ namespace triton { namespace core {
 using cudaStream_t = void*;
 #endif  // !TRITON_ENABLE_GPU
 
+/// Get the memory info for the specified device.
+/// \param device_id The device ID.
+/// \param free Return free memory in bytes.
+/// \param total Return total memory in bytes.
+/// \return The error status. A non-OK status means not all pairs are enabled
```
rmccorm4, line 57:
```suggestion
/// \return The error status. A non-OK status indicates failure to get memory info for the specified device.
```

rmccorm4 (commented):
LGTM except the one copy/paste comment

rmccorm4 (approved):

PR #109, opened 2022-08-03
Add backend attribute for instance group preference

GuanLuo:


GuanLuo:
@nv-kmcgill53 for additional feedback. Look and feel for backend attribute. The idea is to define an extensible interface that Triton can query the backend to make better decision, i.e. whether to set instance kind to GPU if not specified?

Another use case that I want it to also address is when the config is modified, Triton can use this new API to let backend check the modified config (complete config or a representation of the changes) and the backend can inform back to Triton whether the change requires a hard model reload, or can be applied to the current model instances directly.

include/triton/core/tritonbackend.h
```diff
@@ -1359,6 +1360,86 @@ TRITONBACKEND_ISPEC TRITONSERVER_Error* TRITONBACKEND_ModelInstanceExecute(
...
+///
+/// \param backend_attributes The backend_attributes object.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_BackendAttributeDelete(
+    TRITONBACKEND_BackendAttribute* backend_attributes);
+
+///
+/// TRITONBACKEND_ATTRIBUTE_PREFERRED_INSTANCE_GROUP
+///
+/// Set the preferred instance group of the backend. This function should be
```
tanmayv25, line 1421: > This function should be called for TRITONBACKEND_ATTRIBUTE_PREFERRED_INSTANCE_GROUP.

This statement is confusing. How about "This function adds a backend attribute with TRITONBACKEND_ATTRIBUTE_PREFERRED_INSTANCE_GROUP as its kind"?

How about renaming TRITONBACKEND_BackendAttributeSetPreferredInstanceGroup to TRITONBACKEND_BackendAttributeAddPreferredInstanceGroup ? 
This looks more accurate in what it is doing.

include/triton/core/tritonbackend.h
```diff
@@ -1359,6 +1360,86 @@ TRITONBACKEND_ISPEC TRITONSERVER_Error* TRITONBACKEND_ModelInstanceExecute(
...
+/// \param in_params The array of input parameters as a complement of the
+/// attribute inquiry.
+/// \param in_count The number of input parameters.
+/// \param backend_attributes The backend attribute. See
+/// TRITONBACKEND_BackendAttribute on how it should be set for different 'type'.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONBACKEND_ISPEC TRITONSERVER_Error* TRITONBACKEND_GetBackendAttribute(
+    TRITONBACKEND_Backend* backend, TRITONBACKEND_Model* model,
+    const TRITONBACKEND_AttributeType type,
+    const TRITONSERVER_Parameter** in_params, const uint64_t in_count,
```
tanmayv25, line 1394: As an API reader, I am little confused how 
```
    const TRITONSERVER_InstanceGroupKind kind, const uint64_t count,
    const uint64_t* device_ids, const uint64_t id_count)
```
will get mapped to `TRITONSERVER_Parameter** in_params, const uint64_t in_count` ? 
How would multiple `TRITONBACKEND_BackendAttributeSetPreferredInstanceGroup` calls add these parameters? 
Don't we care too much about this because it is expected to be used on the triton core side?

include/triton/core/tritonbackend.h
```diff
@@ -1359,6 +1360,86 @@ TRITONBACKEND_ISPEC TRITONSERVER_Error* TRITONBACKEND_ModelInstanceExecute(
...
+/// \param backend The backend.
+/// \param model The model. 'nullptr' may be provided if no model is associated.
+/// \param type The attribute to query about.
+/// \param in_params The array of input parameters as a complement of the
+/// attribute inquiry.
+/// \param in_count The number of input parameters.
+/// \param backend_attributes The backend attribute. See
+/// TRITONBACKEND_BackendAttribute on how it should be set for different 'type'.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONBACKEND_ISPEC TRITONSERVER_Error* TRITONBACKEND_GetBackendAttribute(
```
Tabrizian, line 1391: I was wondering do you think there will be any benefits if we add an API to set this callback instead of adding it like `TRITONBACKEND_Initialize/TRITONBACKEND_Finalize`?

include/triton/core/tritonbackend.h
```diff
@@ -1359,6 +1360,86 @@ TRITONBACKEND_ISPEC TRITONSERVER_Error* TRITONBACKEND_ModelInstanceExecute(
...
+    TRITONBACKEND_Backend* backend, TRITONBACKEND_Model* model,
```
Tabrizian, line 1392: Perhaps we can break this down further to `TRITONBACKEND_GetBackendAttribute` and  `TRITONBACKEND_GetModelAttribute`? I think we should also be more clear on when we are going to call this callback function. For the two use cases that we currently have in mind, Triton will call `TRITONBACKEND_GetBackendAttribute` once after the `TRIITONBACKEND_Initialize` and another time after `TRITONBACKEND_ModelInitialize` for each model.

include/triton/core/tritonbackend.h
```diff
@@ -1359,6 +1360,86 @@ TRITONBACKEND_ISPEC TRITONSERVER_Error* TRITONBACKEND_ModelInstanceExecute(
...
+/// \param backend_attributes The backend attributes object.
+/// \param kind The kind of the instance group.
+/// \param count The number of instances per device. Triton default will be used
+/// if 0 is provided.
+/// \param device_ids The devices where instances should be available. Triton
+/// default will be used if 'nullptr' is provided.
+/// \param id_count The number of devices in 'device_ids'.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error*
+TRITONBACKEND_BackendAttributeSetPreferredInstanceGroup(
```
Tabrizian, line 1438: I was wondering maybe this should be a model attribute instead of a backend attribute? For example, a backend may decide that the best instance group for this model is CPU and the best instance group for another model is GPU?

Another complete different way of approaching this problem would be to allow setting instance_group settings in the auto-complete API. This way we wouldn't need a new API and the backend can indicate the instance group preference as a part of auto-completing the model configuration.

include/triton/core/tritonbackend.h
GuanLuo: `in_xxx` was designed to for future extensions where some the query for attribute may attach some additional information to deliver the context. This will not be needed for preferred instance group, but if it is needed for other attribute, we will point that out.

GuanLuo: The "answer" will always be returned in 'backend_attribute' argument.

include/triton/core/tritonbackend.h
```diff
@@ -1359,6 +1360,86 @@ TRITONBACKEND_ISPEC TRITONSERVER_Error* TRITONBACKEND_ModelInstanceExecute(
     const uint32_t request_count);
 
 
+/// [BETA] Below is an experimental feature and the interface is subject to
```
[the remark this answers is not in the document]
GuanLuo: I will mark it BETA even if it is merged, I would want to see if it can be extended for other use before fixing the interface

include/triton/core/tritonbackend.h
GuanLuo: I think "adding an API to set this callback" will then require the backend to connect the dots, i.e. if the backend does provide backend attribute, the control will need to reach the backend before they can register the callback. In the current way, Triton can access the "callback" when loading the shared library.

include/triton/core/tritonbackend.h
GuanLuo: The decision on not modifying current auto-complete API behavior is that new API will be more clear on what is changed. The other thing that I was thinking is that if it is done within auto-complete API, the backend needs to check available devices to make the correct decision instead of setting a set of preferred settings and let Triton core to match them.

tanmayv25 (dismissed):

include/triton/core/tritonbackend.h
Tabrizian: @GuanLuo can you reply to this comment? I think using the same "TRITONBACKEND_BackendAttribute" struct for both model attributes and backend attributes can be confusing.

GuanLuo: I dropped the reference to the model at this point. It could be useful to inquiry attributes at model level, but I haven't sorted out the timing of calling such function. For this particular case (preferred instance group), I think we need to inquiry and set the field in model config before auto-complete (so before model init)

Tabrizian (approved):

src/model_config_utils.cc
```diff
@@ -727,66 +708,136 @@ NormalizeModelConfig(
...
-      // For KIND_AUTO... if there are no GPUs or if any of the listed
-      // 'gpu's are not present, then use KIND_CPU.
-      if (group.kind() == inference::ModelInstanceGroup::KIND_AUTO) {
+  // Make sure there is at least one instance_group.
+  if (config->instance_group().empty()) {
+    inference::ModelInstanceGroup* group = config->add_instance_group();
+    group->set_name(config->name());
+
+    for (const auto& pg : preferred_groups) {
+      group->set_kind(pg.kind());
```
Tabrizian, line 744: Do we need to have a multi-GPU test for this to make sure that the instances get correctly populated?

include/triton/core/tritonbackend.h
```diff
@@ -1359,6 +1360,54 @@ TRITONBACKEND_ISPEC TRITONSERVER_Error* TRITONBACKEND_ModelInstanceExecute(
...
```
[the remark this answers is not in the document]
GuanLuo: Good question, I will need to think more about it. I was thinking that "beta" may give us flexibility if the current API need to be changed in the future. But the current form doesn't seem to provide such flexibility because it is the same as other APIs in essence. The proper way may be to isolate BETA APIs from this header so the backends can optionally include the BETA header if they decide to use the experimental APIs with the cost of possible backend incompatible issues.

Now we may just drop the mentioning of "beta" and add this API like the others if it looks future-proof / extensible. And try out different practice once we have come up with a rollout strategy.

src/model_config_utils.cc
```diff
@@ -727,66 +708,136 @@ NormalizeModelConfig(
...
 
-        if (group.kind() == inference::ModelInstanceGroup::KIND_AUTO) {
-          group.set_kind(inference::ModelInstanceGroup::KIND_GPU);
+    // For KIND_AUTO... if there are no GPUs or if any of the listed
+    // 'gpu's are not present, then use KIND_CPU.
+    if (group.kind() == inference::ModelInstanceGroup::KIND_AUTO) {
+      if (supported_gpus.empty()) {
+        group.set_kind(inference::ModelInstanceGroup::KIND_CPU);
+      } else {
+        for (const int32_t gid : group.gpus()) {
```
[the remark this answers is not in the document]
GuanLuo: Yes, the KIND_AUTO description in model config states that the kind will only be set to GPU if all specified devices are available.

src/model_config_utils.cc
GuanLuo: Can you elaborate?

Tabrizian: I meant to add a multi-GPU test that tries the different combinations of the preferred instance group and model instance group if you think it make sense.

Tabrizian (approved):

tanmayv25 (approved):

PR #112, opened 2022-08-04
Fix Windows Compilation

nv-kmcgill53:
This fixes the following error:
>C:\tmp\tritonbuild\tritonserver\build\_deps\repo-core-src\src\tritonserver.cc(1390,5): error C2361: initialization of 'key_prefix' is skipped by 'default' label [C:\tmp\tritonbuild\tritonserver\build\_deps\repo-core-build\triton-core\triton-core.vcxproj] [C:\tmp\tritonbuild\tritonserver\build\_deps\repo-core-build\triton-core.vcxproj]

Link to the MSDN doc for C2361: https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2361?view=msvc-170

### DIFF
--- a/src/tritonserver.cc
+++ b/src/tritonserver.cc
@@ -1384,9 +1384,11 @@ TRITONSERVER_ServerOptionsSetModelLoadDeviceLimit(
       reinterpret_cast<TritonServerOptions*>(options);
   switch (kind) {
     case TRITONSERVER_INSTANCEGROUPKIND_GPU:
-      static std::string key_prefix = "model-load-gpu-limit-device-";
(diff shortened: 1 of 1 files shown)


kthui (approved):

jbkyang-nvi (approved):
do we need to switch other switch cases?

nv-kmcgill53:
I don't believe so as this error only pertains to variables declared inside a case block. The Windows compiler would have failed those other compilations had there been a similar issue. With the CI passing, both the windows builds and tests. I think we are safe to not touch the other switch statements.

PR #114, opened 2022-08-09
Add equal input check while merging payloads

GuanLuo:
Before this change, the payload merging may be performed without checking the shape consistency. This change extend the same check in batcher to the merging function for completeness.

@tanmayv25 I think this is needed to fix the root cause. That being said, I will still revisit the GetDynamicBatch() logic for better batching: we should prefer growing existing payload over creating new payload because we have limit on payloads enqueued (2 payload / instance). If 4 requests arrive, forming 2 2-request payloads has different result from 4 1-request payloads even if payload merging is performed.

### DIFF
--- a/src/payload.cc
+++ b/src/payload.cc
@@ -37,34 +37,38 @@ Payload::Payload()
   exec_mu_.reset(new std::mutex());
 }
 
-Status
+const Status&
 Payload::MergePayload(std::shared_ptr<Payload>& payload)
 {
   if ((payload->GetOpType() != Operation::INFER_RUN) ||
       (op_type_ != Operation::INFER_RUN)) {
-    return Status(
+    static Status op_type_error(
         Status::Code::INTERNAL,
         "Attempted to merge payloads of type that are not INFER_RUN");
+    return op_type_error;
   }
   if (payload->GetInstance() != instance_) {
-    return Status(
+    static Status instance_error(
         Status::Code::INTERNAL,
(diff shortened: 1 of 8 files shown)


src/payload.cc
```diff
@@ -25,6 +25,7 @@
 // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include "payload.h"
+#include "triton/common/logging.h"
```
dyastremsky, line 28: Is this import necessary? It doesn't look you added any logging calls.

GuanLuo: This is from debugging, will remove

rmccorm4 (commented):
Logic and encapsulation make sense to me.

To summarize, is this basically moving the shape check/tracking into Payload class (since it all goes through Payload anyways), and also performing this check in PayloadMerge in addition to existing checks in scheduler? 

I'll leave approval to Tanmay

tanmayv25 (approved):
The performance impact will be only for the cases where the timeout was encountered. It looks okay for resolving the issue.

rmccorm4 (approved):

Tabrizian (commented):
Do we need a test for this to make sure we have coverage in L0_batcher? Just to make sure I'm understanding the issue correctly, before this change we could be creating batches with inconsistent shapes?

tanmayv25:
Yes. Our current tests don't capture the timeout cases where this merge can happen.

PR #115, opened 2022-08-12
Update prometheus-cpp and Metric API unit tests

rmccorm4:
- [x] Update Prometheus API usage for new prometheus-cpp version for metter support in managing duplicate metrics/metric families. Before this bump:
    - Duplicately named metric families of different types (counter, gauge, etc.) could be added to the registry which is against the prometheus spec
    - Duplicately named metric families (even of same type) would keep being added as separate objects to registry. Now duplicates will return a reference to the same family.
- [x] Update unit tests to more thoroughly test duplicates and edge cases
- [x] Think about out-of-order deletion: i.e. calling `MetricFamilyDelete()` before `MetricDelete()` where `Metric->family_->Remove()` reference is called in destructor even though `family_` should be invalid memory now.
- [x] Check if existing metric families from core (nv_request..., nv_gpu_util..., etc.) can be inadvertently removed by creating a duplicate MF with Metrics API and then calling registry->Remove() on it in destructor
    - it can, delaying registry removal until explicitly asked for, see DLIS-4072

### DIFF
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -2269,8 +2269,7 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_MetricFamilyNew(
 /// Delete a metric family object.
 /// A TRITONSERVER_MetricFamily* object should be deleted AFTER its
 /// corresponding TRITONSERVER_Metric* objects have been deleted.
-/// If a family is deleted before its metrics, an error will be
-/// returned on TRITONSERVER_MetricDelete.
+/// Attempting to delete a family before its metrics will return an error.
 ///
 /// \param family The metric family object to delete.
 /// \return a TRITONSERVER_Error indicating success or failure.
@@ -2296,8 +2295,7 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_MetricNew(
 /// Delete a metric object.
 /// All TRITONSERVER_Metric* objects should be deleted BEFORE their
 /// corresponding TRITONSERVER_MetricFamily* objects have been deleted.
-/// If a family is deleted before its metrics, an error will be
-/// returned on TRITONSERVER_MetricDelete.
+/// If a family is deleted before its metrics, an error will be returned.
 ///
 /// \param metric The metric object to delete.
 /// \return a TRITONSERVER_Error indicating success or failure.
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -2267,6 +2267,10 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_MetricFamilyNew(
     const char* name, const char* description);
 
 /// Delete a metric family object.
+/// A TRITONSERVER_MetricFamily* object should be deleted AFTER its
+/// corresponding TRITONSERVER_Metric* objects have been deleted.
+/// If a family is deleted before its metrics, an error will be
+/// returned on TRITONSERVER_MetricDelete.
 ///
 /// \param family The metric family object to delete.
 /// \return a TRITONSERVER_Error indicating success or failure.
@@ -2275,7 +2279,10 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_MetricFamilyDelete(
 
 /// Create a new metric object. The caller takes ownership of the
 /// TRITONSERVER_Metric object and must call
-/// TRITONSERVER_MetricDelete to release the object.
+/// TRITONSERVER_MetricDelete to release the object. The caller is also
+/// responsible for ownership of the labels passed in. Each label can be deleted
+/// immediately after creating the metric with TRITONSERVER_ParameterDelete
+/// if not re-using the labels.
 ///
 /// \param metric Returns the new metric object.
 /// \param family The metric family to add this new metric to.
@@ -2287,6 +2294,10 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_MetricNew(
     const TRITONSERVER_Parameter** labels, const uint64_t label_count);
 
 /// Delete a metric object.
+/// All TRITONSERVER_Metric* objects should be deleted BEFORE their
+/// corresponding TRITONSERVER_MetricFamily* objects have been deleted.
+/// If a family is deleted before its metrics, an error will be
+/// returned on TRITONSERVER_MetricDelete.
 ///
 /// \param metric The metric object to delete.
 /// \return a TRITONSERVER_Error indicating success or failure.
--- a/src/metric_family.cc
+++ b/src/metric_family.cc
@@ -35,6 +35,12 @@ namespace triton { namespace core {
 //
 // Implementation for TRITONSERVER_MetricFamily.
 //
+
+// Prometheus returns existing family pointer if a family with the same name
+// already exists. So we must track the reference count of the prometheus
+// family to know when we can remove it from the Prometheus Registry.
+static std::unordered_map<void*, size_t> family_ref_cnt_;
+
 MetricFamily::MetricFamily(
     TRITONSERVER_MetricKind kind, const char* name, const char* description)
 {
@@ -59,6 +65,10 @@ MetricFamily::MetricFamily(
   }
 
   kind_ = kind;
+
+  // Maintain reference count to manage duplicate names for registry removal
+  std::lock_guard<std::mutex> lk(mtx_);
+  ++family_ref_cnt_[family_];
 }
 
 void*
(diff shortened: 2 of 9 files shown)


rmccorm4:
New unit tests:
```
[==========] Running 9 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 9 tests from MetricsApiTest
[ RUN      ] MetricsApiTest.TestCounterEndToEnd
[       OK ] MetricsApiTest.TestCounterEndToEnd (241 ms)
[ RUN      ] MetricsApiTest.TestGaugeEndToEnd
[       OK ] MetricsApiTest.TestGaugeEndToEnd (1 ms)
[ RUN      ] MetricsApiTest.TestDupeMetricFamilyDiffKind
[       OK ] MetricsApiTest.TestDupeMetricFamilyDiffKind (0 ms)
[ RUN      ] MetricsApiTest.TestDupeMetricFamilyDiffDescription
[       OK ] MetricsApiTest.TestDupeMetricFamilyDiffDescription (1 ms)
[ RUN      ] MetricsApiTest.TestDupeMetricFamily
[       OK ] MetricsApiTest.TestDupeMetricFamily (1 ms)
[ RUN      ] MetricsApiTest.TestDupeMetricLabels
metric1 value: 7.5 == metric2 value: 7.5
[       OK ] MetricsApiTest.TestDupeMetricLabels (0 ms)
[ RUN      ] MetricsApiTest.TestDupeMetricEmptyLabels
metric1 value: 7.5 == metric2 value: 7.5
[       OK ] MetricsApiTest.TestDupeMetricEmptyLabels (1 ms)
[ RUN      ] MetricsApiTest.TestOutOfOrderDelete
[       OK ] MetricsApiTest.TestOutOfOrderDelete (0 ms)
[ RUN      ] MetricsApiTest.TestCoreMetricAccess
[       OK ] MetricsApiTest.TestCoreMetricAccess (0 ms)
[----------] 9 tests from MetricsApiTest (250 ms total)

[----------] Global test environment tear-down
[==========] 9 tests from 1 test suite ran. (250 ms total)
[  PASSED  ] 9 tests.
```

src/test/metrics_api_test.cc
```diff
@@ -236,30 +344,239 @@ TEST_F(MetricsApiTest, TestGaugeEndToEnd)
...
+  const char* description = "diff kind description";
+  FAIL_TEST_IF_ERR(
+      TRITONSERVER_MetricFamilyNew(&family1, kind1, name, description),
+      "Creating new metric family1");
+
+  // Create duplicate metric family with different kind
+  TRITONSERVER_MetricFamily* family2 = nullptr;
+  TRITONSERVER_MetricKind kind2 = TRITONSERVER_METRIC_KIND_GAUGE;
+  // Expect this to fail, can't have duplicate name of different kind
+  auto err = TRITONSERVER_MetricFamilyNew(&family2, kind2, name, description);
```
Tabrizian, line 376: If the user's Python implementation have two instances of the model and they are calling `TRITONSERVER_MetricFamilyNew` in their `initialize` function, the initialize call for the second model instance will fail since the metric family already exists. Do you think it makes sense to return the existing metric family instead of returning an error?

src/test/metrics_api_test.cc
```diff
@@ -236,30 +344,239 @@ TEST_F(MetricsApiTest, TestGaugeEndToEnd)
...
+
+  std::vector<const TRITONSERVER_Parameter*> labels2;
+  labels2.emplace_back(TRITONSERVER_ParameterNew(
+      metric_key, TRITONSERVER_PARAMETER_STRING, "label2"));
+  TRITONSERVER_Metric* metric2 = nullptr;
+  FAIL_TEST_IF_ERR(
+      TRITONSERVER_MetricNew(&metric2, family2, labels2.data(), labels2.size()),
+      "Creating new metric2");
+  for (const auto label : labels2) {
+    TRITONSERVER_ParameterDelete(const_cast<TRITONSERVER_Parameter*>(label));
```
Tabrizian, line 480: I think it would be good to update the backend API documentation to mention that `TRITONSERVER_MetricNew` does not take the ownership of labels and the caller must delete them.

src/tritonserver.cc
```diff
@@ -2962,8 +2960,16 @@ TRITONSERVER_Error*
 TRITONSERVER_MetricDelete(TRITONSERVER_Metric* metric)
 {
 #ifdef TRITON_ENABLE_METRICS
-  delete reinterpret_cast<tc::Metric*>(metric);
-  return nullptr;  // Success
+  auto lmetric = reinterpret_cast<tc::Metric*>(metric);
+  if (lmetric->Family() == nullptr) {
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_INTERNAL,
+        "MetricFamily reference was invalidated before Metric was deleted. "
```
Tabrizian, line 2967: I think we should update the documentation for the metrics API to mention that it is best to delete all the metrics corresponding to a metric family before deleting the metric family. Otherwise, an error object will be received.

src/test/metrics_api_test.cc
rmccorm4: If two pb instances call `MetricFamilyNew(same_name, same_kind, ...)`, it will succeed and they will get the exiting metric family back.

This test is checking if they try to do `family1 = MetricFamilyNew(same_name, kind=COUNTER)` and `family2 = MetricFamilyName(same_name, kind=GAUGE)`

It is not allowed to have duplicate names of different `kind`

src/tritonserver.cc
rmccorm4: Great suggestion, done

include/triton/core/tritonserver.h
```diff
@@ -2267,6 +2267,10 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_MetricFamilyNew(
     const char* name, const char* description);
 
 /// Delete a metric family object.
+/// A TRITONSERVER_MetricFamily* object should be deleted AFTER its
+/// corresponding TRITONSERVER_Metric* objects have been deleted.
+/// If a family is deleted before its metrics, an error will be
```
GuanLuo, line 2272: If the family has been deleted, what happened if user calls other `TRITONSERVER_MetricXxx` functions? Should also return error as the metric will no longer be valid?

src/metric_family.cc
```diff
@@ -62,62 +62,71 @@ MetricFamily::MetricFamily(
...
     }
     default:
       throw std::invalid_argument(
           "Unsupported family kind passed to Metric constructor.");
   }
 
-  std::lock_guard<std::mutex> lk(mtx_);
-  ++metric_ref_cnt_[metric];
-  return metric;
+  std::lock_guard<std::mutex> lk(metric_mtx_);
```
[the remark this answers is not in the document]
rmccorm4: Just to check if I understand roughly what you're going for, you're suggesting something along these lines?
```cpp
MetricFamily::MetricFamily() {
  ...
  std::lock_guard<std::mutex> lk(metric_mtx);
  auto prom_metric_sp = std::shared_ptr<void>(prom_metric, MyCustomDeleter);
  return prom_metric_sp;
}
```

```cpp
MetricFamily::Remove(std::shared_ptr<void> prom_metric_sp) {
  ...
  delete prom_metric_sp;
}
```

```cpp
void MetricFamily::MyCustomDeleter() {
  switch (kind_) {
    case TRITONSERVER_METRIC_KIND_COUNTER: {
      auto counter_family_ptr =
          reinterpret_cast<prometheus::Family<prometheus::Counter>*>(family_);
      auto counter_ptr = reinterpret_cast<prometheus::Counter*>(prom_metric);
      counter_family_ptr->Remove(counter_ptr);
      break;
    }
    case TRITONSERVER_METRIC_KIND_GAUGE: {
      auto gauge_family_ptr =
          reinterpret_cast<prometheus::Family<prometheus::Gauge>*>(family_);
      auto gauge_ptr = reinterpret_cast<prometheus::Gauge*>(prom_metric);
      gauge_family_ptr->Remove(gauge_ptr);
      break;
    }
    default:
      // Invalid kind should be caught in constructor
      LOG_ERROR << "Unsupported kind in Metric destructor.";
      break;
  }
}
```

CC @GuanLuo who added the ref counting

src/test/metrics_api_test.cc
Tabrizian: Got it. Thanks for the clarification. Resolving.

src/metric_family.cc
```diff
@@ -62,62 +62,71 @@ MetricFamily::MetricFamily(
...
+  child_metrics_.insert(metric);
+  return prom_metric;
 }
 
 void
-MetricFamily::Remove(void* metric)
+MetricFamily::Remove(void* prom_metric, Metric* metric)
 {
+  // Remove reference to dependent Metric object
+  child_metrics_.erase(metric);
```
GuanLuo, line 98: lock?

include/triton/core/tritonserver.h
rmccorm4: Adding more unit tests around this

rmccorm4: Technically MetricSet/Increment/Value etc. will work as they operate directly on the metric and don't need to communicate with tc::MetricFamily object. 

Since we don't remove a MetricFamily's prometheus::Family from the prometheus::Registry on ~MetricFamily destruction, the tc::Metric will have a valid reference to the prometheus::Metric still in the prometheus::registry and can update it accordingly.

If we want to disallow this, we can also invalidate the reference by setting `tc::Metric::metric_ = nullptr`? Let me know what you think.

rmccorm4: I also added a similar check to MetricFamilyDelete to instead enforce that FamilyDelete is called only once its dependent metrics are deleted.

src/metric_family.cc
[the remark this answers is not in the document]
Tabrizian: It looks like it might not be easy to implement in this scenario. For me the original `ResponseFactory` was a shared pointer.

Tabrizian (approved):

GuanLuo (approved):

PR #116, opened 2022-08-22
Improve PriorityQueue performance with large priority_levels

QMassoz:
Modify PriorityQueue to dynamically create/remove PolicyQueue items instead of instantiating them in the constructor. User-given PolicyQueue (defined in priority_queue_policy) are instantiated in constructor and are kept instantiated. Thus only default PolicyQueue are affected by this PR.

This PR results in significant performance boost when priority_levels takes a large value (above ~100k). Next step would be to have the priority_ variable of InferenceRequest to be expressed in uint64_t instead of uint32_t.

### DIFF
--- a/src/scheduler_utils.cc
+++ b/src/scheduler_utils.cc
@@ -247,11 +247,18 @@ PriorityQueue::PolicyQueue::TimeoutAt(size_t idx)
   }
 }
 
-PriorityQueue::PriorityQueue()
-    : size_(0), front_priority_level_(0), last_priority_level_(0)
+bool
+PriorityQueue::PolicyQueue::ReadyForErasure()
 {
-  inference::ModelQueuePolicy default_policy;
-  queues_.emplace(0, PolicyQueue(default_policy));
+  size_t total_size = Size() + rejected_queue_.size();
+  return !keep_instantiated_ && total_size == 0;
+}
+
+PriorityQueue::PriorityQueue()
+    : size_(0), front_priority_level_(0), last_priority_level_(0),
+      default_policy_()
+{
+  queues_.emplace(0, PolicyQueue(default_policy_, true));
   front_priority_level_ = queues_.begin()->first;
   ResetCursor();
 }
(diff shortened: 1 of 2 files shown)


QMassoz:
Hello, how do we request reviews?

src/scheduler_utils.cc
```diff
@@ -247,41 +247,54 @@ PriorityQueue::PolicyQueue::TimeoutAt(size_t idx)
...
-      auto it = queue_policy_map.find(level);
-      if (it == queue_policy_map.end()) {
-        queues_.emplace(level, PolicyQueue(default_queue_policy));
-      } else {
-        queues_.emplace(level, PolicyQueue(it->second));
-      }
+    // All priorities with user-given policy are instantiated. We do not
+    // permanently add default PolicyQueue because those will be dynamically
+    // created and erased to keep memory footprint low
+    for (auto it = queue_policy_map.begin(); it != queue_policy_map.end();
```
dyastremsky, line 281: For consistency with the other code in the core repo, can we use the shorter form a for loop here? E.g. for (const auto& queue_policy : queue_policy_map)

src/scheduler_utils.cc
```diff
@@ -303,38 +316,43 @@ Status
 PriorityQueue::Dequeue(std::unique_ptr<InferenceRequest>* request)
 {
   pending_cursor_.valid_ = false;
-  while (true) {
-    if (!queues_[front_priority_level_].Empty()) {
-      RETURN_IF_ERROR(queues_[front_priority_level_].Dequeue(request));
+  auto it_start = queues_.lower_bound(front_priority_level_);
+  for (auto it = it_start; it != queues_.end(); ++it) {
```
dyastremsky, line 320: Shorter for loop for consistency.

src/scheduler_utils.cc
QMassoz: Done, replaced it with a range-based for loop

src/scheduler_utils.cc
QMassoz: A range-based for loop (e.g. for (auto& queue : queues_)) will range from begin to end and not benefit from the (small) optimization using it_start (found in logarithmic time using front_priority_level_).

A suggestion for a shorter for loop:
auto it = queues_.lower_bound(front_priority_level_);
for (; it != queues_.end(); ++it) {
  ...
}

QMassoz: Or we can drop the use of "front_priority_level_", as the request to be dequeued is likely to be found in the first key values of "queues_" (excepted when there is a large amount of user-specified non-default PolicyQueue)

dyastremsky: Thanks for explaining and addressing the other loop. This looks good to me.

I'm less familiar with these files, so I'll wait for @GuanLuo to give this PR a once-over as well,

QMassoz:
Any update?

QMassoz:
I pushed commit [003a2c1 ](https://github.com/triton-inference-server/core/pull/116/commits/003a2c16235daaf43edc1edbd38ebadceb9f2e2a) because of a segfault that occurred very rarely in very particular situations since updating to newer version of triton server/client.

dyastremsky:
Thanks! Guan, are you able to review this PR?

src/scheduler_utils.cc
GuanLuo: Let's keep `front_priority_level_` for now to minimize change until necessary, but I think `last_priority_level_` can be safely removed as it is no longer used anywhere.

GuanLuo (approved):
Thanks for the PR, it looks good to me. @dyastremsky do you want to trigger a pipeline for the change?

dyastremsky:
Fantastic, sure. Running tests. Will approve and merge if it passes.

I can open a PR to remove last_priority_level_ once merged.

dyastremsky:
Thank you for your contribution! Great work, merged.

PR #117, opened 2022-08-22
Add CPU metrics to core

rmccorm4:
- [x] CMake option (default true)
- [x] Bump C API version
- [x] Use `Status` instead of `bool` internally more?
- [x] CpuInfo struct
    - [x] Add link / document `/proc/stat` values info
    - [x] Add more values to struct? iowait, interrupts, etc. or keep it simple?


src/metrics.cc
```diff
@@ -378,6 +412,134 @@ Metrics::PollCacheMetrics(std::shared_ptr<RequestResponseCache> response_cache)
...
+  std::string line;
+  constexpr uint64_t KB = 1024;
+  while (std::getline(ifs, line)) {
+    std::istringstream iss(line);
+    std::string name;
+    uint64_t value = 0;
+    if (iss >> name >> value) {
+      // Remove trailing ":" after field names, convert KB to bytes
+      name.pop_back();
+      (*info)[name] = value * KB;
```
rmccorm4, line 466: I compared this code with only  parsing for the "MemTotal" and "MemAvailable" fields and exiting early once they're found, and found minimal difference in performance. 

For 1000 iterations of each:
```
exiting early after finding both fields:   avg=3.454ms   // + extra code to track when we're done
parsing all fields:                        avg=3.524ms
```

rmccorm4:
Added info line to server startup similar to GPU metrics to raise awareness
```
I0824 21:20:05.783147 167727 metrics.cc:873] Collecting metrics for GPU 0: NVIDIA GeForce RTX 3090
I0824 21:20:05.783423 167727 metrics.cc:766] Collecting CPU metrics
```

rmccorm4:
CC @Tabrizian I ran a few iterations of `perf_analyzer -m simple` with the server running cpu metrics and NOT running cpu metrics, and the performance stayed relatively stable for both tests, whether metrics were enabled or not. 

So I think it is OK to have ON by default.

```
# CPU Metrics enabled
Concurrency: 1, throughput: 1098.03 infer/sec, latency 909 usec
Concurrency: 1, throughput: 1116.06 infer/sec, latency 894 usec
Concurrency: 4, throughput: 2745.2 infer/sec, latency 1455 usec
Concurrency: 4, throughput: 2635.23 infer/sec, latency 1516 usec

# CPU Metrics disabled
Concurrency: 1, throughput: 1105.53 infer/sec, latency 903 usec
Concurrency: 1, throughput: 1055.1 infer/sec, latency 946 usec
Concurrency: 4, throughput: 2706.1 infer/sec, latency 1476 usec
Concurrency: 4, throughput: 2663.41 infer/sec, latency 1500 usec
```

src/metrics.cc
```diff
@@ -378,6 +412,156 @@ Metrics::PollCacheMetrics(std::shared_ptr<RequestResponseCache> response_cache)
...
+    return Status(
+        Status::Code::INTERNAL,
+        "Failed to find aggregate CPU info in /proc/stat.");
+  }
+
+  std::string _;
+  std::istringstream iss(line);
+  // Use _ to skip "cpu" at start of line
+  if (!(iss >> _ >> info->user >> info->nice >> info->system >> info->idle >>
+        info->iowait >> info->irq >> info->softirq >> info->steal)) {
```
GuanLuo, line 441: You can override `operator>>` for your CpuInfo class https://docs.microsoft.com/en-us/cpp/standard-library/overloading-the-input-operator-for-your-own-classes?view=msvc-170

src/metrics.cc
```diff
@@ -378,6 +412,156 @@ Metrics::PollCacheMetrics(std::shared_ptr<RequestResponseCache> response_cache)
...
+  // Account for overflow
+  const auto wrap_sub = [](uint64_t a, uint64_t b) {
+    return (a > b) ? (a - b) : 0;
+  };
+  uint64_t util_diff = wrap_sub(info_new->user, info_old->user) +
+                       wrap_sub(info_new->nice, info_old->nice) +
+                       wrap_sub(info_new->system, info_old->system) +
+                       wrap_sub(info_new->irq, info_old->irq) +
+                       wrap_sub(info_new->softirq, info_old->softirq) +
+                       wrap_sub(info_new->steal, info_old->steal);
```
GuanLuo, line 522: How to interpret "util" in this context? Can we sum "user" and "nice" with other fields?

src/metrics.cc
```diff
@@ -378,6 +412,156 @@ Metrics::PollCacheMetrics(std::shared_ptr<RequestResponseCache> response_cache)
...
+  if (status.IsOk()) {
+    cpu_util = CpuUtilization(cpu_info, last_cpu_info_);
+    *last_cpu_info_ = *cpu_info;
+  }
+  cpu_utilization_->Set(cpu_util);  // [0.0, 1.0]
+
+  // RAM / Memory
+  double mem_total_bytes = 0.0;
+  double mem_used_bytes = 0.0;
+  auto mem_info = std::make_shared<MemInfo>();
```
GuanLuo, line 549: Why use smart pointer here

src/metrics.cc
```diff
@@ -378,6 +412,156 @@ Metrics::PollCacheMetrics(std::shared_ptr<RequestResponseCache> response_cache)
...
+    return Status(
+        Status::Code::INTERNAL,
+        "Failed to parse aggregate CPU info in /proc/stat.");
+  }
+  return Status::Success;
+#endif  // OS
+}
+
+Status
+Metrics::ParseMemInfo(std::shared_ptr<MemInfo> info)
```
GuanLuo, line 451: Pass by reference

Tabrizian:
@rmccorm4 Thanks for verifying this!

src/metrics.cc
rmccorm4: All fields other than `idle` and `iowait` are considered "active/utilized" time spent on CPU. Would you prefer a breakdown like [this](https://github.com/htop-dev/htop/blob/15652e7b8102e86b3405254405d8ee5d2a239004/linux/LinuxProcessList.c#L1286-L1294)?

GuanLuo (approved):

src/metrics.cc
GuanLuo: I just don’t have the knowledge of what the fields mean, I don’t have preference as long as the fields are in the same “unit”

rmccorm4: Ah yes they're all in the same [units](https://man7.org/linux/man-pages/man5/procfs.5.html):
> The amount of time, measured in units of USER_HZ
>                     (1/100ths of a second on most architectures ...

PR #120, opened 2022-09-20
Adding the fixes for arm32

thmca:
I intend on using the Triton Inference server on 3 devices (amd64, arm64 and arm/v7). After trying to build the server for my arm/v7 device, I noticed that there where casting errors. I realized that I needed to make similar fixes in the backend and server repositories as well. Once all the casting issues have been solved in the 3 repos, I was able to correctly run inferences on all three of my devices.


nv-kmcgill53 (approved):
CI passed

PR #127, opened 2022-10-10
Fix Payload Queues Race Condition

dyastremsky:
Locking the payload queues when deallocating to prevent concurrent deletion of models by multiple threads. The prevents a potential segfault during model unregistering.

### DIFF
--- a/src/rate_limiter.cc
+++ b/src/rate_limiter.cc
@@ -108,7 +108,7 @@ RateLimiter::UnregisterModel(const TritonModel* model)
   }
 
   {
-    std::lock_guard<std::mutex> lk1(payload_queues_mu_);
+    std::lock_guard<std::mutex> lk(payload_queues_mu_);
     if (payload_queues_.find(model) != payload_queues_.end()) {
       payload_queues_.erase(model);
     }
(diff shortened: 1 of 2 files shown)


src/rate_limiter.cc
```diff
@@ -107,8 +107,11 @@ RateLimiter::UnregisterModel(const TritonModel* model)
     RETURN_IF_ERROR(resource_manager_->UpdateResourceLimits());
   }
 
-  if (payload_queues_.find(model) != payload_queues_.end()) {
-    payload_queues_.erase(model);
+  {
+    std::lock_guard<std::mutex> lk1(payload_queues_mu_);
```
GuanLuo, line 111: just `lk`

GuanLuo: Should also add lock on creating the payload_queue

rmccorm4 (commented):
Unrelated to the PR, but is there any reason why we don't return an ERROR status instead of logging in these two spots @tanmayv25 ?
- https://github.com/triton-inference-server/core/blob/ad4ad54bc420c2ed7b44241aa6995174c8613b95/src/rate_limiter.cc#L139
- https://github.com/triton-inference-server/core/blob/ad4ad54bc420c2ed7b44241aa6995174c8613b95/src/rate_limiter.cc#L182

src/rate_limiter.cc
dyastremsky: Thanks, updated and added.

GuanLuo (approved):

rmccorm4 (approved):

PR #129, opened 2022-10-10
Check if model config override is provided

krishung5:


src/model_repository_manager.h
```diff
@@ -81,7 +81,6 @@ class ModelRepositoryManager {
     std::string model_name_;
     Status status_;
     bool checked_;
-    // FIXME
```
rmccorm4, line 84: was this not needed? or is there some corresponding ticket?

rmccorm4: Looks like this was added by @GuanLuo a while back here: https://github.com/triton-inference-server/server/commit/09271f9c4d4d935bd9667dd2be2208d2dd5effd0

@GuanLuo any context?

src/model.cc
```diff
@@ -96,7 +96,7 @@ Model::Init()
     const auto config_path = JoinPath({model_dir_, kModelConfigPbTxt});
     bool exists = false;
     RETURN_IF_ERROR(FileExists(config_path, &exists));
-    if (!exists) {
+    if (!exists && !is_config_override_) {
```
GuanLuo, line 99: Instead of passing `is_config_override_`, you can pass a generic flag `is_config_provided_` which is `config_override | config_file_exisits`. In fact, you can pass `parsed_config` in `ModelRepositoryManager::InitializeModelInfo()` which should have the same evaluation as what is mentioned above.

krishung5: Updated, please review.

tanmayv25 (approved):

src/model.cc
```diff
@@ -88,15 +88,15 @@ Model::SetScheduler(std::unique_ptr<Scheduler> scheduler)
...
 Status
-Model::Init()
+Model::Init(const bool is_config_provided)
 {
   // If the model configuration has not been set, then look
   // whether the config file was explicitly provided.
   if (!set_model_config_) {
     const auto config_path = JoinPath({model_dir_, kModelConfigPbTxt});
     bool exists = false;
     RETURN_IF_ERROR(FileExists(config_path, &exists));
```
GuanLuo, line 98: Don't need this

GuanLuo: parsed_config already check the existence https://github.com/triton-inference-server/core/blob/7aa014ca08ade31f338158ddbd3aa83fb6a54726/src/model_repository_manager.cc#L1126-L1137

krishung5: Updated, thanks for the catch.

krishung5:
[CI](https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/jobs/46251103) passed.

GuanLuo (approved):

PR #131, opened 2022-10-11
Make the finalize_mtx_ scheduler specific

tanmayv25:
If trying to send the response within the response callback of another model that holds the finalize_mtx_ lock, the thread will run into deadlock as it tries to re-acquire the lock.

The change move the ordering mutexes specific to the scheduler object.

### DIFF
--- a/src/dynamic_batch_scheduler.cc
+++ b/src/dynamic_batch_scheduler.cc
@@ -667,8 +667,7 @@ DynamicBatchScheduler::FinalizeResponses()
 {
   // Need exclusive access of the function to ensure responses are sent
   // in order
-  static std::mutex finalize_mtx;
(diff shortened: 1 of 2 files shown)


GuanLuo (approved):

rmccorm4 (approved):

Tabrizian (approved):
Does this summary explain why this fixes the deadlock?

The oldest sequence batcher was being used which means that `preserver_ordering` will be true. The BLS backend will call `InferenceResponse::Send` in the response callback. Because the `mutex` was static it was shared among all models. In the BLS backend, because InferenceResponse::Send was triggering the response send in another model it lead to a deadlock because of shared mutex?

tanmayv25:
@Tabrizian  Exactly!

PR #142, opened 2022-11-23
Add tritoncache APIs

rmccorm4:
Main Changes:
- General structure of new TRITONCACHE APIs and scaffolding from CLI to Core has added
    - `tritoncache.cc` was added with core-side TRITONCACHE API implementations
    - `cache_manager.*` and `cache_entry.*` have been added to encapsulate shared library communication
    - Some optimizations can definitely be made in the number of copies being done and lifetimes/ownership, open to a lot of feedback here
- `--response-cache-byte-size` and corresponding C API has been deprecated in favor of `--cache-config` and `--cache-dir`. SetResponseCacheByteSize API will now call SetCacheConfig with specified size for now.
- Some "band-aid" fixes have been made on the metrics/stats related to caching, but will be fixed more thoroughly in a follow-up PR
- Response Cache unit test has been updated to use new Cache API via TritonCacheManager 

Some (potential) follow-up PRs:
- [ ] docs updates (response_cache.md, metrics.md, etc.)
- [ ] (?) per-model caches, cache config specified in model `config.pbtxt` as suggested by Iman
- [ ] remove copy from cache impl side by instead adding some callback/signal to hold onto a buffer until Triton has copied it to response output
    - may profile with nsys to see the specific impact here
 
<details>
<summary>TODO Items</summary>

- [x] serialize/deserialize responses
- [x] simple test that responses are cached correctly with example client
- [x] Remove or update `--response-cache-byte-size > 0` logic for enabling cache
    - this can be controlled by cache/cache_config
- [x] pass cache_config to TritonCache
- [x] pass cache_dir to TritonCache
- [x] re-enable minimal unit test
- [x] minimally fix cache stats for perf analyzer (follow up PR for thorough fixes)
- [x] Check memory leaks
- [x] check copyrights
- [x] Update cache unit test (follow up test PR: #144)
- [ ] Add API descriptions/docs (TBD on some copy/ownership changes)
---
- [ ] Performance benchmarking
    - [ ] Optimize copies / remove unnecessary copies
    - [ ] Consider Exists API, may be tricky concurrency wise (exists when checking but not when looking up), or hurt performance (add concurrency around Exists + Insert/Lookup)
    - [ ] Nsight compare
    - [ ] Verify no C++17 related bottlenecks
- [ ] Check stats/metrics (follow up metric PR)
    - [ ] Check L0_perf_analyzer_report
    - [x] Update cache miss stats through model rather than request due to request lifetime issues (released by backend before callback)
 - [ ] Update response_cache.md doc (follow up doc PR)

</details>

Corresponding server changes: https://github.com/triton-inference-server/server/pull/5119

### DIFF
--- a/include/triton/core/tritoncache.h
+++ b/include/triton/core/tritoncache.h
@@ -1,4 +1,4 @@
-// Copyright 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
--- a/include/triton/core/tritoncache.h
+++ b/include/triton/core/tritoncache.h
@@ -129,9 +129,6 @@ TRITONCACHE_ISPEC TRITONSERVER_Error* TRITONCACHE_CacheInsert(
 TRITONCACHE_ISPEC TRITONSERVER_Error* TRITONCACHE_CacheLookup(
     TRITONCACHE_Cache* cache, const char* key, TRITONCACHE_CacheEntry* entry);
 
-TRITONCACHE_ISPEC TRITONSERVER_Error* TRITONCACHE_CacheEvict(
-    TRITONCACHE_Cache* cache);
-
 /* CacheEntry Lifetime Management */
 
 // TODO: Add API descriptions
--- a/include/triton/core/tritoncache.h
+++ b/include/triton/core/tritoncache.h
@@ -157,15 +157,22 @@ TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryDelete(
 
 // TODO: Add API descriptions
 
-// Sets items in entry
-TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntrySetItems(
-    TRITONCACHE_CacheEntry* entry, void* items, size_t* byte_sizes,
-    size_t num_items);
+// Get number of items in entry
+TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItemCount(
+    TRITONCACHE_CacheEntry* entry, size_t* count);
 
-// Gets items from entry
-TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItems(
-    TRITONCACHE_CacheEntry* entry, void** items, size_t** byte_sizes,
-    size_t* num_items);
+// Gets the index'th item from entry where 0 <= index < count where
+// 'count' is the value returned by TRITONCACHE_CacheEntryItemCount
+TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItem(
+    TRITONCACHE_CacheEntry* entry, size_t index, void** item, size_t* byte_size);
+
+// Adds item to entry
+TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryAddItem(
+    TRITONCACHE_CacheEntry* entry, void* item, size_t byte_size);
+
+/* Not currently used, may be added in the future for grouping
+   cache entries for actions like evicting all entries associated
+   with certain model, tags, etc.
 
 // Sets tags in entry
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntrySetTags(
@@ -174,6 +181,8 @@ TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntrySetTags(
 // Gets tags from entry
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryTags(
     TRITONCACHE_CacheEntry* entry, char*** tags, size_t* num_tags);
+
+*/
 
 #ifdef __cplusplus
 }
--- a/include/triton/core/tritoncache.h
+++ b/include/triton/core/tritoncache.h
@@ -55,6 +55,7 @@ extern "C" {
 #endif
 
 struct TRITONCACHE_Cache;
+struct TRITONCACHE_CacheEntry;
 
 ///
 /// TRITONCACHE API Version
@@ -98,6 +99,8 @@ struct TRITONCACHE_Cache;
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_ApiVersion(
     uint32_t* major, uint32_t* minor);
 
+/* Cache Lifetime Management */
+
 /// Create a new cache object. The caller takes ownership of the
 /// TRITONCACHE_Cache object and must call
 /// TRITONCACHE_CacheDelete to release the object.
@@ -115,18 +118,62 @@ TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheNew(
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheDelete(
     TRITONCACHE_Cache* cache);
 
-// TODO
+/* Cache Usage */
+
+// TODO: Add API descriptions
+
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheInsert(
-    TRITONCACHE_Cache* cache);
+    TRITONCACHE_Cache* cache, const char* key, TRITONCACHE_CacheEntry* entry);
 
-// TODO
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheLookup(
-    TRITONCACHE_Cache* cache, const char* key, void** entries, size_t** sizes,
-    size_t* num_entries);
+    TRITONCACHE_Cache* cache, const char* key, TRITONCACHE_CacheEntry** entry);
 
-// TODO
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEvict(
     TRITONCACHE_Cache* cache);
+
+/* CacheEntry Lifetime Management */
+
+/* Cache Entry implementation example
+**
+** TODO: Move to tritoncache.cc or similar
+**
+** struct CacheEntry {
+**   void*  items;         // blobs of data to insert or retrieve with cache
+**   size_t* byte_sizes;   // size of each item in items
+**   size_t  num_items;    // number of items and byte_sizes
+**   char**  tags;         // (optional) tags to associate entry, groups, etc.
+**   size_t  num_tags;     // (optional) number of tags provided
+** }
+*/
+
+// TODO: Add API descriptions
+TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryNew(
+    TRITONCACHE_CacheEntry** entry);
+
+TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryDelete(
+    TRITONCACHE_CacheEntry* entry);
+
+/* CacheEntry Field Management */
+
+// TODO: Add API descriptions
+
+// Sets items in entry
+TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntrySetItems(
+    TRITONCACHE_CacheEntry* entry, void* items, size_t* byte_sizes,
+    size_t num_items);
+
+// Gets items from entry
+TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItems(
+    TRITONCACHE_CacheEntry* entry, void** items, size_t** byte_sizes,
+    size_t* num_items);
+
+// Sets tags in entry
+TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntrySetTags(
+    TRITONCACHE_CacheEntry* entry, char** tags, size_t num_tags);
+
+// Gets tags from entry
+TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryTags(
+    TRITONCACHE_CacheEntry* entry, char*** tags, size_t* num_tags);
 
 #ifdef __cplusplus
 }
--- a/include/triton/core/tritoncache.h
+++ b/include/triton/core/tritoncache.h
@@ -100,7 +100,9 @@ struct TRITONCACHE_CacheEntryItem;
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_ApiVersion(
     uint32_t* major, uint32_t* minor);
 
-/* Cache Lifetime Management */
+///
+/// Cache Lifetime Management
+///
 
 /// Create a new cache object. The caller takes ownership of the
 /// TRITONCACHE_Cache object and must call
@@ -119,7 +121,9 @@ TRITONCACHE_ISPEC TRITONSERVER_Error* TRITONCACHE_CacheNew(
 TRITONCACHE_ISPEC TRITONSERVER_Error* TRITONCACHE_CacheDelete(
     TRITONCACHE_Cache* cache);
 
-/* Cache Usage */
+///
+/// Cache Usage
+///
 
 // TODO: Add API descriptions
 
@@ -129,18 +133,19 @@ TRITONCACHE_ISPEC TRITONSERVER_Error* TRITONCACHE_CacheInsert(
 TRITONCACHE_ISPEC TRITONSERVER_Error* TRITONCACHE_CacheLookup(
     TRITONCACHE_Cache* cache, const char* key, TRITONCACHE_CacheEntry* entry);
 
-/* CacheEntry Lifetime Management */
+///
+/// CacheEntry Lifetime Management
+///
 
-// TODO: Add API descriptions
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryNew(
     TRITONCACHE_CacheEntry** entry);
 
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryDelete(
     TRITONCACHE_CacheEntry* entry);
 
-/* CacheEntry Field Management */
-
-// TODO: Add API descriptions
+///
+/// CacheEntry Field Management
+///
 
 // Get number of items in entry
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItemCount(
@@ -156,16 +161,19 @@ TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryGetItem(
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryAddItem(
     TRITONCACHE_CacheEntry* entry, TRITONCACHE_CacheEntryItem* item);
 
-/* CacheEntryItem Lifetime Management */
+///
+/// CacheEntryItem Lifetime Management
+///
 
-// TODO: flesh out
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItemNew(
     TRITONCACHE_CacheEntryItem** item);
 
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItemDelete(
     TRITONCACHE_CacheEntryItem* item);
 
-/* CacheEntryItem Field Management */
+///
+/// CacheEntryItem Field Management
+///
 
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItemBufferCount(
     TRITONCACHE_CacheEntryItem* item, size_t* count);
@@ -179,7 +187,8 @@ TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItemGetBuffer(
     TRITONCACHE_CacheEntryItem* item, size_t index, void** base,
     size_t* byte_size);
 
-/* Not currently used, may be added in the future for grouping
+/*
+   Not currently used, may be added in the future for grouping
    cache entries for actions like evicting all entries associated
    with certain model, tags, etc.
 
@@ -190,7 +199,6 @@ TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntrySetTags(
 // Gets tags from entry
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryTags(
     TRITONCACHE_CacheEntry* entry, char*** tags, size_t* num_tags);
-
 */
 
 #ifdef __cplusplus
--- a/include/triton/core/tritoncache.h
+++ b/include/triton/core/tritoncache.h
@@ -164,11 +164,11 @@ TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItemCount(
 // Gets the index'th item from entry where 0 <= index < count where
 // 'count' is the value returned by TRITONCACHE_CacheEntryItemCount
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItem(
-    TRITONCACHE_CacheEntry* entry, size_t index, void** item, size_t* byte_size);
+    TRITONCACHE_CacheEntry* entry, size_t index, void** base, size_t* byte_size);
 
 // Adds item to entry
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryAddItem(
-    TRITONCACHE_CacheEntry* entry, void* item, size_t byte_size);
+    TRITONCACHE_CacheEntry* entry, const void* base, size_t byte_size);
 
 /* Not currently used, may be added in the future for grouping
    cache entries for actions like evicting all entries associated
--- a/include/triton/core/tritoncache.h
+++ b/include/triton/core/tritoncache.h
@@ -158,6 +158,28 @@ TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItem(
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryAddItem(
     TRITONCACHE_CacheEntry* entry, const void* base, size_t byte_size);
 
+/*
+
+// TODO : May benefit from another level of indirection, one less copy
+          by passing list of pointers rather than forming contiguous buffer
+
+// Get number of buffers in item
+TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItemBufferCount(
+    TRITONCACHE_CacheEntryItem* item, size_t* count);
+
+// Gets the index'th buffer from item where 0 <= index < count where
+// 'count' is the value returned by TRITONCACHE_CacheItemBufferCount
+TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItemBuffer(
+    TRITONCACHE_CacheEntryItem* item, size_t index, void** base,
+    size_t* byte_size);
+
+// Adds item to entry
+TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItemAddBuffer(
+    TRITONCACHE_CacheEntryItem* item, const void* base, size_t byte_size);
+
+*/
+
+
 /* Not currently used, may be added in the future for grouping
    cache entries for actions like evicting all entries associated
    with certain model, tags, etc.
--- a/include/triton/core/tritoncache.h
+++ b/include/triton/core/tritoncache.h
@@ -121,7 +121,7 @@ TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheInsert(
 
 // TODO
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheLookup(
-    TRITONCACHE_Cache* cache, char* key, void** entries, size_t** sizes,
+    TRITONCACHE_Cache* cache, const char* key, void** entries, size_t** sizes,
     size_t* num_entries);
 
 // TODO
--- a/include/triton/core/tritoncache.h
+++ b/include/triton/core/tritoncache.h
@@ -164,7 +164,8 @@ TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItemCount(
 // Gets the index'th item from entry where 0 <= index < count where
 // 'count' is the value returned by TRITONCACHE_CacheEntryItemCount
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItem(
-    TRITONCACHE_CacheEntry* entry, size_t index, void** base, size_t* byte_size);
+    TRITONCACHE_CacheEntry* entry, size_t index, void** base,
+    size_t* byte_size);
 
 // Adds item to entry
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryAddItem(
@@ -185,5 +186,5 @@ TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryTags(
 */
 
 #ifdef __cplusplus
-}
+}  // extern C
 #endif
--- a/include/triton/core/tritoncache.h
+++ b/include/triton/core/tritoncache.h
@@ -101,57 +101,40 @@ TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_ApiVersion(
     uint32_t* major, uint32_t* minor);
 
 ///
-/// CacheEntry Lifetime Management
-///
-
-// TODO: Currently not using EntryNew/EntryDelete APIs at all. Instead
-//       we are creating an Entry in TritonCacheManager and passing an opaque
-//       pointer to it for cache to add Items/Buffers to over C APIs. See
-//       TritonCache::Lookup as an example. This ownership flow should be
-//       clarified/cleaned up.
-
-/// Create a new cache entry object. The caller takes ownership of the
-/// TRITONCACHE_CacheEntry object and must call
-/// TRITONCACHE_CacheEntryDelete to release the object.
-///
-/// \param entry Returns the new cache entry object.
-/// \return a TRITONSERVER_Error indicating success or failure.
-
-// TODO: Not currently using this API. See TritonCache Lookup/Insert
-// for explanation. Should clean up / clarify the expected ownerships here.
-TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryNew(
-    TRITONCACHE_CacheEntry** entry);
-
-/// Delete a cache entry object.
-///
-/// \param entry The cache entry object to delete.
-/// \return a TRITONSERVER_Error indicating success or failure.
-
-// TODO: Not currently using this API. See TritonCache Lookup/Insert
-// for explanation. Should clean up / clarify the expected ownerships here.
-TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryDelete(
-    TRITONCACHE_CacheEntry* entry);
-
-///
 /// CacheEntry Field Management
 ///
 
-// TODO: Proper API descriptions, lifetimes, copies, etc.
-
-// Get number of items in entry
+/// Get the number of items available in the entry.
+///
+/// \param entry The CacheEntry object to query.
+/// \param count Returns the number of items in entry.
+/// \return a TRITONSERVER_Error indicating success or failure.
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItemCount(
     TRITONCACHE_CacheEntry* entry, size_t* count);
 
-// Adds item to entry
+/// Adds item to the entry.
+///
+/// TODO: ownership/lifetime
+///
+/// \param entry The CacheEntry object to add item to.
+/// \param item The CacheEntryItem being added to entry.
+/// \return a TRITONSERVER_Error indicating success or failure.
 TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryAddItem(
     TRITONCACHE_CacheEntry* entry, TRITONCACHE_CacheEntryItem* item);
 
-// Gets item at index from entry where 0 <= index < count and
-// 'count' is the value returned by TRITONCACHE_CacheEntryItemCount
+/// Gets the item at index from entry.
+///
+/// The caller does not own the returned item and must not modify or delete it.
+/// The lifetime of the item extends until 'entry' or 'item' is deleted.
+///
+/// \param entry The CacheEntry object.
+/// \param index The index of the item, must be 0 <= index < count, where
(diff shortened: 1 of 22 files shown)


src/cache_manager.cc
```diff
@@ -0,0 +1,414 @@
...
+Status
+TritonCache::InitializeCacheImpl()
+{
+  if (init_fn_ == nullptr) {
+    return Status(Status::Code::NOT_FOUND, "cache init function is nullptr");
+  }
+
+  // TODO: TRITONSERVER_Message doesn't seem to do any validation on the
+  //       string to assert valid JSON. Not sure why passing
+  //       'TritonServerMessage' would be better than just passing 'const char*'
```
rmccorm4, line 148: @GuanLuo curious of your opinion on passing `const char* config_json` vs. `TRITONSERVER_Message* config_json` to shared library. 

I thought `TritonServerMessage` would do some JSON validation, but it doesn't, so I wasn't sure what value it added in this use case.

include/triton/core/tritoncache.h
```diff
@@ -0,0 +1,261 @@
...
+///         - TRITONSERVER_ERROR_INVALID_ARG
+///           - bad argument passed, nullptr, etc.
+///         - TRITONSERVER_ERROR_ALREADY_EXISTS
+///           - key already exists and will not be inserted again
+///         - TRITONSERVER_ERROR_INTERNAL
+///           - internal errors
+///         - nullptr
+///           - success
+TRITONCACHE_ISPEC TRITONSERVER_Error* TRITONCACHE_CacheInsert(
+    TRITONCACHE_Cache* cache, const char* key, TRITONCACHE_CacheEntry* entry);
```
GuanLuo, line 154: I don't see why you need `CacheEntry`, you can have insert and lookup directly with respect to a list of `CacheItem`. Is this just to avoid `***` for lookup?

include/triton/core/tritoncache.h
```diff
@@ -0,0 +1,261 @@
...
+
+// TODO: Proper API descriptions, lifetimes, copies, etc.
+
+// Returns number of buffers held by item
+TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItemBufferCount(
+    TRITONCACHE_CacheEntryItem* item, size_t* count);
+
+// Adds buffer to item
+TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItemAddBuffer(
+    TRITONCACHE_CacheEntryItem* item, const void* base, size_t byte_size);
```
GuanLuo, line 251: Doesn't look like we are making any copy, so the buffer owner must not release the buffer until some point. Should have a signal mechanism that the buffer owner can know the buffer can be safely release. Perhaps a callback to be provided in `TRITONCACHE_CacheEntryItemNew` and invoke in `TRITONCACHE_CacheEntryItemDelete`

include/triton/core/tritoncache.h
```diff
@@ -0,0 +1,261 @@
...
+/// TRITONCACHE_CacheDelete to release the object.
+///
+/// This API is implemented by the user-provided cache implementation,
+/// so specific details will be found within the cache implementation.
+///
+/// \param cache Returns the new cache object.
+/// \param config The config options to initialize the cache with.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONCACHE_ISPEC TRITONSERVER_Error* TRITONCACHE_CacheNew(
+    TRITONCACHE_Cache** cache, TRITONSERVER_Message* config);
```
GuanLuo, line 118: I don't think the type of `config` matters, Triton itself doesn't need to know anything about the config, so you can even pass a buffer (void* buf, uint32_t byte_size) that is read directly from the config file. The cache implementation will have complete control over how the config is formatted (can be JSON, protobuf, or custom key-value list)

include/triton/core/tritonserver.h
```diff
@@ -1692,16 +1692,39 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error*
...
+/// FIXME: Included for future extensibility, but not currently used.
+/// \param cache_name The name of the cache.
+/// \param config_json The string representation of config JSON.
 /// \return a TRITONSERVER_Error indicating success or failure.
 TRITONSERVER_DECLSPEC TRITONSERVER_Error*
-TRITONSERVER_ServerOptionsSetResponseCacheByteSize(
-    TRITONSERVER_ServerOptions* options, uint64_t size);
+TRITONSERVER_ServerOptionsSetCacheConfig(
+    TRITONSERVER_ServerOptions* options, const char* cache_name,
+    const char* config_json);
```
Tabrizian, line 1715: I am a bit hesitant in adding a new configuration file. I would prefer to have all the control options from either command line or model configuration. This would introduce a third place that the user would need to modify in order to change some of the options.

include/triton/core/tritoncache.h
```diff
@@ -0,0 +1,275 @@
...
+
+// TODO: Proper API descriptions, lifetimes, copies, etc.
+
+// Returns number of buffers held by item
+TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItemBufferCount(
+    TRITONCACHE_CacheEntryItem* item, size_t* count);
+
+// Adds buffer to item
+TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItemAddBuffer(
+    TRITONCACHE_CacheEntryItem* item, const void* base, size_t byte_size);
```
Tabrizian, line 176: Does the buffer API need to be memory type aware? For example, in the future we may decide to add support for GPU buffers. It seems like there is no way right now to specify what memory that the `base` pointer is in.

include/triton/core/tritoncache.h
```diff
@@ -0,0 +1,275 @@
...
+TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItemCount(
+    TRITONCACHE_CacheEntry* entry, size_t* count);
+
+// Adds item to entry
+TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryAddItem(
+    TRITONCACHE_CacheEntry* entry, TRITONCACHE_CacheEntryItem* item);
+
+// Gets item at index from entry where 0 <= index < count and
+// 'count' is the value returned by TRITONCACHE_CacheEntryItemCount
+TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryGetItem(
```
Tabrizian, line 139: To summarize, a CacheEntry has many CacheEntryItems which can have many CacheEntryBuffers? Why do we need two level hierarchy? Can we flatten the `CacheEntryBuffers` into `CacheEntryItems`?

Tabrizian (changes requested):

include/triton/core/tritoncache.h
rmccorm4: Yes `CacheEntry` was to avoid `***` for lookup.

src/CMakeLists.txt
```diff
@@ -62,7 +64,10 @@ endif() # TRITON_ENABLE_GPU
 #
 # Boost
 #
-find_package(Boost REQUIRED)
+# Minimum of 1.78 required for use of boost::span. This can eventually be 
+# relaxed and replaced with std::span in C++20.
+#
+find_package(Boost 1.78 REQUIRED)
```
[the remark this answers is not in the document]
rmccorm4: It is installed in `build.py` from `tar.gz` instead of `apt install`: https://github.com/triton-inference-server/server/pull/5119

src/tritoncache.cc
```diff
@@ -0,0 +1,239 @@
...
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+// OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "cache_entry.h"
+
+// For unknown reason, windows will not export the TRITONCACHE_*
```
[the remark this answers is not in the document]
rmccorm4: No the cache is not currently tested on windows before nor after these C API changes. I believe this is the same with the repo_agent: https://github.com/triton-inference-server/core/blob/main/src/repo_agent.cc#L35-L45

include/triton/core/tritoncache.h
```diff
@@ -0,0 +1,275 @@
...
+///         - TRITONSERVER_ERROR_INVALID_ARG
+///           - bad argument passed, nullptr, etc.
+///         - TRITONSERVER_ERROR_ALREADY_EXISTS
+///           - key already exists and will not be inserted again
+///         - TRITONSERVER_ERROR_INTERNAL
+///           - internal errors
+///         - nullptr
+///           - success
+TRITONCACHE_ISPEC TRITONSERVER_Error* TRITONCACHE_CacheInsert(
+    TRITONCACHE_Cache* cache, const char* key, TRITONCACHE_CacheEntry* entry);
```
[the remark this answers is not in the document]
rmccorm4: I discussed this with the Redis team and they did not see value in abstracting the keys. The argument was that strings are flexible and ubiquitously used by the majority share of caches. When they're not used, they're still castable/convertible to other types fairly easily if needed. So I decided to go with the "make the common case fast" approach for the workflow.

```
# some cache test or application
cache->Insert("mykey1", obj)
cache->Lookup("mykey1")
```
vs. having to do this to setup a key for each access
```
# some cache test or application
TRITONCACHE_CacheKey* key = nullptr;
TRITONCACHE_CacheKeyNew(&key);
cache->Insert(key, obj)
TRITONCACHE_CacheKeyDelete(key);
```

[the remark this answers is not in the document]
rmccorm4: >  I think const char* key would prevent us from having key names that require `\0` in the middle of the key.
This seems like a reasonable thing for us to not support in exchange for the ease-of-use it offers. 

If we really wanted to support that, I believe including a `size_t byte_size` along with it could support that use case if copied correctly.

src/cache_manager.h
```diff
@@ -0,0 +1,134 @@
...
+#include "status.h"
+#include "triton/common/model_config.h"
+#include "tritonserver_apis.h"
+
+namespace triton { namespace core {
+
+//
+// Proxy to a cache shared library.
+//
+class TritonCache {
```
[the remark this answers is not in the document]
rmccorm4: I was just following the convention of `TritonBackend/TritonBackendManager` and `TritonRepoAgent/TritonRepoAgentManager` as this is a class for how Triton will interact with any cache shared library implementing TRITONCACHE APIs. 

I'm not partial one way or the other.

src/cache_manager.h
```diff
@@ -0,0 +1,134 @@
...
+  TritonCacheLookupFn_t lookup_fn_;
+  typedef TRITONSERVER_Error* (*TritonCacheInsertFn_t)(
+      TRITONCACHE_Cache* cache, const char* key, TRITONCACHE_CacheEntry* entry);
+  TritonCacheInsertFn_t insert_fn_;
+};
+
+//
+// Manage communication with Triton caches and their lifecycle.
+//
+class TritonCacheManager {
```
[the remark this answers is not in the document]
rmccorm4: Answered above.

include/triton/core/tritonserver.h
rmccorm4: This is a bit tricky, as the cache config is designed to be arbitrary from Triton's POV, it is just passed to the cache implementation to be used however it wants. 

1. For CLI, we would have to either add new fields for every cache and every cache config option we want to officially support, which we don't want to do. Or we can support a generic string CLI argument that supports the same idea of reading arbitrary config from file: `--cache-config "field1=123; field2=456; field3=789"`, but this may be large for some caches or as the scope of config grows. 

I expected to ship the container with default config that should not require too much tweaking for users that want to get going immediately, or is easy to copy-in and replace.

I'm not totally opposed to this option if you think having another file for users to manage is cumbersome.

2. For model config, this is also tricky as we don't currently support unique caches per-model. We would need to map each model to it's own cache with it's own unique config. Maybe this is a route we could go, but I didn't think it was the route we were taking at this time.

include/triton/core/tritoncache.h
rmccorm4: This is not currently memory type aware. Currently everything would have to go through CPU memory. I agree it would be better if we included the fields to support other memory types for the API, even if they weren't completely supported yet.

To avoid too many single large PRs, I want to gather the changes in a side `rmccormick-cache-main` branch, When we sort out the final desired APIs in terms of entries/items/buffers, I think I can make this a follow-up PR so there's less to review.

include/triton/core/tritoncache.h
rmccorm4: Similar discussion here: https://github.com/triton-inference-server/core/pull/142#discussion_r1043883249

include/triton/core/tritoncache.h
Tabrizian: Thanks. Resolving.

include/triton/core/tritoncache.h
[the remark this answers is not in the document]
Tabrizian: I think having `byte_size` is a good enhancement. If I'm understanding correctly, there could be scenarios where the cache key is a result of a hash that contains `\0` in it. The cache implementer would need to have other workarounds to avoid this issue.

src/tritoncache.cc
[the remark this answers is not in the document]
Tabrizian: Can you file a ticket for the same?

include/triton/core/tritonserver.h
Tabrizian: Thanks for the detailed response! So this is just a byte blob passed from Triton to the cache library? Can you provide some examples for the fields that are currently used for TritonCache and RedisCache? I think one of them is `TRITONSERVER_ServerOptionsSetResponseCacheByteSize`.

rmccorm4: `TRITONSERVER_ServerOptionsSetResponseCacheByteSize` is going to be deprecated in favor of passing a config that specifies a `size` (or similar) field. ex:
```
{ "response_cache_byte_size": 1048576 }
```
Currently [LocalCache](https://github.com/triton-inference-server/local_cache/pull/1) (the default implementation/shared library) will only expect this size field in a json config (open to change, but it's a convenient format to enforce). New fields may be added in the future.

`RedisCache` will be expecting Redis related configurations such as an IP:Port of a Redis server, some kind of token/cert/authentication, etc. - but those choices can be up to them. ex:
```
{ "redis_server_addr": "localhost:6379", ... }
```
They were also happy with a JSON config file, but functionally it can be any bytes read from a file (when using CLI) or any string (when using the API directly) passed to the `cache_implementation->CacheNew(&cache, cache_config)`. Currently, Triton won't validate or enforce the contents to look any type of way.

src/tritoncache.cc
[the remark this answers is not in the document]
rmccorm4: Filed https://jirasw.nvidia.com/browse/DLIS-4406

src/CMakeLists.txt
```diff
@@ -225,7 +233,7 @@ add_library(
   TritonCore::triton-core ALIAS triton-core
 )
 
-target_compile_features(triton-core PRIVATE cxx_std_11)
+target_compile_features(triton-core PRIVATE cxx_std_17)
```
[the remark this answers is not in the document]
rmccorm4: The only issue I can think of would be the OS not shipping gcc/g++ with -std=c++17 by default, but we are currently on Ubuntu 20.04, moving to 22.04 soon, and Ubuntu 18.04 had c++17 support by default as well for customers staying on an older version.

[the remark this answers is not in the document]
Tabrizian: Are you using any of the C++17 features in this PR?

include/triton/core/tritonserver.h
Tabrizian: Unfortunately, we can't deprecate an TRITONSERVER/BACKEND API. The APIs should stay backwards compatible. Would it be possible to define an override mechanism? For example, if the cache byte size is specified in `TRITONSERVER_ServerOptionsSetResponseCacheByteSize` it takes priority over `cache_config`.

>For model config, this is also tricky as we don't currently support unique caches per-model. 

Also, I was wondering who provides the `config_json` file? It seems like currently the cache library would call `TRITONSERVER_ServerOptionsSetCacheConfig`?

GuanLuo: I can see why you want cache to behave differently for different models, i.e. you want to perform group operation on certain "cache entries" (release the cache if the model is unloaded). Then you can think about adding "tags" to cache entry, which will be a placeholder (no tags defined) for now and extensible in the future.

rmccorm4: > Then you can think about adding "tags" to cache entry, which will be a placeholder (no tags defined) for now and extensible in the future.

My plan was to eventually add some `TRITONCACHE_CacheEntryAddTags(...)` / `TRITONCACHE_CacheEntryGetTags(...)` for things like flushing specific entries from cache, but didn't include as it wasn't being used currently.

src/CMakeLists.txt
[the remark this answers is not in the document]
rmccorm4: There are several currently:
- std::byte
- std::optional
- std::shared_mutex - for better read-only Lookup concurrency
    - After some reading, I plan to re-evaluate the benefit  when benchmarking
- structure bindings (`auto& [k, v] = pair/tuple/map`) just syntactic sugar for `x = iter.first, y = iter.second everywhere`

The `shared_mutex` was the main motivation as I suspected a noticeable improvement, and the rest were used while upgrading. But if there are any issues, I don't mind swapping them back to C++11 equivalents.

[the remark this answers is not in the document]
Tabrizian: Hopefully, it should be fine. I'll be resolving this thread.

include/triton/core/tritoncache.h
rmccorm4: I added some comments in-line about this, please see here: https://github.com/triton-inference-server/core/pull/142/commits/4c7a7c0850df23a30c31a0939b51f16f2db67e7d

Similarly, it turned out I didn't even end up using the TRITONCACHE_CacheEntryNew/Delete APIs, and instead created the `CacheEntry`s in cache_manager and passed their opaque pointers for cache to temporarily use (get or add buffers).

I could use an extra pair of eyes/brain to help define the expected ownership, and if we even need these APIs. 

I think `TRITONCACHE_CacheEntryItemNew` makes sense to use/create an item on cache side, since we don't know the expected number of items on Triton side before calling CacheLookup. 

For `TRITONCACHE_CacheEntryItemDelete`, I don't know if it makes more sense for the cache to maintain or transfer ownership of a `CacheItem` after creating it - this would help clarify the need for signal to delete. Transferring ownership seems to work because the cache simply copies it's respective buffers (copy since cache may evict the relevant data before it's used) into the item for Triton to use.

include/triton/core/tritoncache.h
```diff
@@ -0,0 +1,294 @@
...
+
+// Returns number of buffers held by item
+TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItemBufferCount(
+    TRITONCACHE_CacheEntryItem* item, size_t* count);
+
+// Adds buffer of specified memory_type/id to item. Only CPU memory supported
+// currently.
+TRITONCACHE_DECLSPEC TRITONSERVER_Error* TRITONCACHE_CacheEntryItemAddBuffer(
+    TRITONCACHE_CacheEntryItem* item, const void* base, size_t byte_size,
+    TRITONSERVER_MemoryType memory_type, int64_t memory_type_id);
```
Tabrizian, line 193: Use buffer attributes API so that we can stay backwards compatible, in case we decide to add new fields to the buffers: https://github.com/triton-inference-server/core/blob/e0c85317a08565bfdccc73eed490734c49887098/include/triton/core/tritonserver.h#L1428-L1441

include/triton/core/tritoncache.h
```diff
@@ -0,0 +1,294 @@
...
+///
+/// This API is implemented by the user-provided cache implementation,
+/// so specific details will be found within the cache implementation.
+///
+/// \param cache Returns the new cache object.
+/// \param config The config options to initialize the cache with.
+///               This will be passed as-is to the cache implementation, so
+///               the expected format and parsing is up to the cache as well.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONCACHE_ISPEC TRITONSERVER_Error* TRITONCACHE_CacheNew(
```
Tabrizian, line 225: How about using `CacheInitialize` and `CacheFinalize` so that it is more similar to the backend API?

include/triton/core/tritonserver.h
```diff
@@ -1692,16 +1692,39 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error*
 TRITONSERVER_ServerOptionsSetCudaMemoryPoolByteSize(
     TRITONSERVER_ServerOptions* options, int gpu_device, uint64_t size);
 
-/// Set the total response cache byte size that the server can allocate in CPU
-/// memory. The response cache will be shared across all inference requests and
-/// across all models.
+// Deprecated. See TRITONSERVER_ServerOptionsSetCacheConfig instead.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error*
```
Tabrizian, line 1696: Can you add the old comment based on the discussion here:

https://github.com/triton-inference-server/core/pull/142/files#r1046033041

src/cache_entry.cc
```diff
@@ -0,0 +1,349 @@
...
+  std::unique_lock lk(buffer_mu_);
+  return buffers_;
+}
+
+void
+CacheEntryItem::AddBuffer(const void* base, size_t byte_size)
+{
+  std::unique_lock lk(buffer_mu_);
+  // COPY: Make a copy of buffer for Triton to own
+  void* new_base = malloc(byte_size);
```
Tabrizian, line 78: If we make a copy here, it means that we can transfer the ownership of the buffer to the cache library? I was wondering when implementing the redis cache would it also make a copy of this buffer or it is able to use the raw pointer?

src/cache_entry.cc
```diff
@@ -0,0 +1,349 @@
...
+  int64_t memory_type_id = 0;
+  void* userp = nullptr;
+  RETURN_NULLOPT_IF_STATUS_ERROR(output.DataBuffer(
+      &output_base, &output_byte_size, &memory_type, &memory_type_id, &userp));
+
+  // DLIS-2673: Add better memory_type support
+  if (memory_type != TRITONSERVER_MEMORY_CPU &&
+      memory_type != TRITONSERVER_MEMORY_CPU_PINNED) {
+    LOG_ERROR
+        << "Only input buffers in CPU memory are allowed in cache currently";
```
Tabrizian, line 223: Does this message gets printed on every inference run? Instead of logging the error, do you think it would be possible to return the error and respond with error for request execution?

include/triton/core/tritoncache.h
Tabrizian: How about having the first CacheItem represent the total number of responses (n) and the number of outputs in each response (m_1..m_n), then Triton will consider the first `m_1` buffers as outputs of response 1 and so on?

include/triton/core/tritonserver.h
Tabrizian: > I can see why you want cache to behave differently for different models, i.e. you want to perform group operation on certain "cache entries" (release the cache if the model is unloaded). Then you can think about adding "tags" to cache entry, which will be a placeholder (no tags defined) for now and extensible in the future.

Actually, what I had in mind was a bit different from what you described here. I was thinking about a case where the user wants to use one cache implementation for some models and another implementation for other models (e.g., for large inputs use local cache and for small inputs use redis cache). Right now the user needs to swap the `libtritoncache` in and out of `/opt/tritonserver/caches` to use a different cache library. It means that we cannot ship both of these cache library implementations in the container. I think having the cache config as a part of model configuration would allow us to:

1. Easily change the cache configuration
2. Have a unified place for all the model configuration parameters.

Looks like, in model repository agents we also allow arbitrary key/value paths similar to what Ryan was describing.
https://github.com/triton-inference-server/common/blob/5313d0be99204c89ab696023534043f2ba535868/protobuf/model_config.proto#L1727-L1758

include/triton/core/tritoncache.h
GuanLuo: I think the `EntryNew/Delete` can be removed since it is not going to interfere with the cache implementation.

I would keep `TRITONCACHE_CacheEntryItemDelete` if `New` is exposed to the cache implementation, If the cache calls `New` without appending to the entry, there can be memory leak. Exposing `Delete` provides a way to clean it up in this case.

> cache simply copies it's respective buffers (copy since cache may evict the relevant data before it's used) into the item for Triton to use.

Then there are two copies? One from cache internal to cache item, another from cache item to response?

include/triton/core/tritoncache.h
GuanLuo: > How about having the first CacheItem represent the total number of responses (n) and the number of outputs in each response (m_1..m_n), then Triton will consider the first m_1 buffers as outputs of response 1 and so on?

This may not work unless the cache insertion is done differently. If the insertion is triggered whenever a response is returned, in the case of decoupled model, the first insertion doesn't know how many responses will there be. Unless it holds and perform insertion at once when knowing the last response has been received (`RESPONSE_COMPLETE` flag), the "holding" part may not as simple as insertion on every response

include/triton/core/tritonserver.h
rmccorm4: @Tabrizian I could see per-model caches as a follow-up refactor (within this release). I agree it's more in-line with the way Backend and RepoAgent APIs are setup and used.

Per-model caches should inherently be less prone to collisions and hopefully involve less evictions per-model being caused by other models (or more if evenly dividing memory but there are unbalanced number of requests coming in per-model, this would be a configuration knob to optimize per use case). This would induce extra thought in how to configure the cache for each model.

Some issues I would want to sort out:

1. If a model is loaded/unloaded, what should we do with the cache? Should this be configurable? Do we sometimes want to keep the cache while a model is unloaded because we plan to load the model again later? Do we sometimes want to delete the cache on model unload because we know we're done with the model? (Possible config options: `keep_loaded: true/false`, `spill_to_disk: true/false`, etc.

2. Do we scrap the concept of server-wide cache metrics in this case and only have per-model cache metrics? 

3. If models each request a certain amount of cache memory, but their total exceeds available memory - is it first come first serve? Should a model fail to load if the cache can not be initialized? (Possible config options: `load_on_cache_fail: true/false`, etc.)

Trying to think of any other potential issues, let me know if you can think of any upfront.

CC @GuanLuo for thoughts on moving to per-model caches + cache_config in model_config

GuanLuo: Interesting idea, I think it can be a future step if multiple kinds of cache is necessary. I don't think the API needs to be changed to make room for this, everything needed seems to be additions: the API doesn't need to change, just need internal manager to identify and dispatch to the entry points of correct cache library; Add additional way to accept model-specific config (really just pointer to which kind of cache to be used).

rmccorm4: C API would stay same, but `tritonserver` CLI interaction would have to change. i.e. whether or not to have  `--cache-config`/`--cache-dir` flags, or move to new `config.pbtxt` fields

include/triton/core/tritoncache.h
Tabrizian: I see. For decoupled models, are we going to insert into the cache whenever a response is sent? In that case what happens if another request arrives with the same inputs before the insertion of previous request is finished? I think we probably need to the "holding" part regardless of the API we have for the cache.

include/triton/core/tritonserver.h
Tabrizian: Re Ryan>

1. I think for now we can delete the cache associated with the model on model unloading. For example, the user may upload a new model with different outputs and we may return the old cached results which is not the expected behavior. As you pointed out, we can in the future add new options to the model configuration to further customize cache behavior. 

2. I think we kind of had per-model cache previously too. By including the model name and version in the hash we were expecting to have different caches for each model. For now, I think we can stick to a per-model cache. We can add a "shared cache" concept later by adding more options in the model config if needed.

3. I would say if they have enabled caching but it exceeds the available memory, we should fail the model loading and ask the user to adjust the cache size.

>C API would stay same, but tritonserver CLI interaction would have to change. i.e. whether or not to have --cache-config/--cache-dir flags

I think only the `--cache-config` will be removed but `--cache-dir` would still exist. You can have `/opt/tritonserver/caches/redis`, `/opt/tritonserver/caches/local` and `--cache-dir` would point out to where to find the cache implementations.

include/triton/core/tritoncache.h
rmccorm4: > Then there are two copies? One from cache internal to cache item, another from cache item to response?

`TRITONCACHE_CacheLookup`:
Currently yes, one copy in `TRITONCACHE_CacheEntryItemAddBuffer()` from cache_impl -> triton_item_buffer, and then a second copy from triton_item_buffer -> response_output_buffer.

We might be able to skip the 2nd copy if there's a method to directly set an response output buffer without `AllocateDataBuffer`. I don't see one in `infer_response.cc`, but I imagine there must be some mechanism we use elsewhere.

Or we might be able to skip the first copy through some signaling mechanism where cache doesn't deallocate the buffer until cache manager copies and signals, but that might be a bit tricky.

include/triton/core/tritoncache.h
rmccorm4: Along these lines of future decoupled support, do you think we should include some sort of insertion behavior to support this?
```
typedef enum TRITONCACHE_InsertBehavior {
  // default: fail on inserting a key that already exists
  TRITONCACHE_UNIQUE,
  // support decoupled: append items to existing key (response1 added previously, response2,3,... added later)
  TRITONCACHE_APPEND, 
  // overwrite existing key if it exists: maybe useful for some cases, not sure yet
  TRITONCACHE_REPLACE
};

TRITONCACHE_CacheInsert(..., TRITONCACHE_InsertBehavior behavior)
```

Tabrizian: I think it would be easier on the Cache API part if Triton can handle the "holding" part and insert all the responses at once. IMO, making cache responsible for the holding part is a bit unclean. @GuanLuo/Ryan do you think we can store all the responses belonging to a request in the Triton core, so that we don't have to add a new API to handle decoupled models in cache?

GuanLuo: I agree that Triton should handle the holding because of the question that @Tabrizian raised,
> In that case what happens if another request arrives with the same inputs before the insertion of previous request is finished?

In the decoupled case, the "cache entry" shouldn't be considered ready until the last response is processed. From the perspective of cache implementation, it makes sense that "insertion" is a sign that the cache entry is ready. Otherwise, it is extra requirement for cache to implement "holding" state, and the insert behavior above doesn't really address that, it is instead indicating that the content of cache entry could be mutable after insertion.

I am okay with insert behavior though, it makes sense for cache in general, but I can't think of when Triton may use other behaviors on top of my head. And if you do want to add it, should note that all behaviors are optional, the cache implementation should return error for behaviors that are not supported.

include/triton/core/tritonserver.h
rmccorm4: As discussed offline, we'll keep the concept of a global cache and corresponding cache config, but leave the C APIs designed in a way that allows per-model caches in the future. 

Similarly, will move from JSON config to Protobuf/pbtxt config that can be used standalone for global cache config for now, and then simply be plugged into model `config.pbtxt` section for `response_cache` following the same structure as global cache config to be more intuitive.

Copying response here:

> The CLI will point to the global cache config file, perhaps with a default of `/opt/tritonserver/caches/global_config.pbtxt` or similar. The global cache will not be removed when adding support for per-model caches. The global cache will be used instead if a model enables caching but doesn't specify any per-model cache config. The `--cache-config` CLI flag won't be removed.

rmccorm4:
@GuanLuo @Tabrizian let me know if you have any additional high-level feedback here. I will be merging into an `rmccormick-cache-main` side branch before final PR into `main`.

To try to keep the logical changes grouped and smaller, I will have follow-up PRs for:
- protobuf config file (DLIS-4469)
- copy optimization (DLIS-4471)
- metrics cleanup (DLIS-4372)

src/cache_entry.cc
```diff
@@ -308,37 +325,30 @@ CacheEntryItem::FromBytes(boost::span<const std::byte> packed_bytes)
...
-  LOG_VERBOSE(2) << "[DECODE] dtype=" << dtype;
-  LOG_VERBOSE(2) << "[DECODE] shape_byte_size=" << shape_byte_size;
-  LOG_VERBOSE(2) << "[DECODE] shape=";
-  for (const auto& dim : shape) {
-    LOG_VERBOSE(2) << dim << " ";
+    auto fail = Status(
+        Status::Code::INTERNAL, "Unexpected number of bytes received: " +
+                                    std::to_string(packed_bytes.size()) +
+                                    ", expected: " + std::to_string(position));
+    return {fail, output};
```
Tabrizian, line 343: nit: I think it would be better to just return the error instead of pair and get the output as one of the arguments. 1. the memory copies are more clear. 2. We can use the `RETURN_IF_ERROR` macros. It can be fixed in a future PR.

include/triton/core/tritoncache.h
Tabrizian: What is the conclusion on this thread? Are we going to flatten the cache items into cache buffers?

src/cache_entry.cc
rmccorm4: Sure, will do

include/triton/core/tritoncache.h
rmccorm4: I believe the choices mentioned to remove a level of indirection are:
- pass list of CacheItems instead of one CacheEntry (I believe this will require a triple pointer in API, not necessarily a bad thing, just avoiding if possible)
- flatten list of buffers into one buffer inside CacheItem (I believe this will have a few downsides below): 
    - extra allocation and extra copies to form contiguous buffer of concatenated responses on triton side before sending to cache
    - local cache will still request allocated buffer from fixed-size boost managed memory, so the pre-allocated contiguous buffer from triton will be wasted
    - I believe storing a larger contiguous buffer (of concatenated buffers) will be likely be less efficient and have more fragmentation issues in boost managed buffer

Leaving the way it is now was to avoid the issues listed above. Let me know your thoughts

include/triton/core/tritonserver.h
```diff
@@ -1692,6 +1692,8 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error*
 TRITONSERVER_ServerOptionsSetCudaMemoryPoolByteSize(
     TRITONSERVER_ServerOptions* options, int gpu_device, uint64_t size);
 
+/// Deprecated. See TRITONSERVER_ServerOptionsSetCacheConfig instead.
```
Tabrizian, line 1695: Remove "Deprecated".

src/cache_entry.cc
```diff
@@ -0,0 +1,354 @@
...
+  std::unique_lock lk(buffer_mu_);
+  return buffers_;
+}
+
+void
+CacheEntryItem::AddBuffer(const void* base, size_t byte_size)
+{
+  std::unique_lock lk(buffer_mu_);
+  // COPY: Make a copy of buffer for Triton to own
+  void* new_base = malloc(byte_size);
```
Tabrizian, line 78: Do we need to user Triton's internal memory pool for this? CC @GuanLuo

include/triton/core/tritoncache.h
```diff
@@ -0,0 +1,278 @@
+// Copyright 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
Tabrizian, line 1: Copyrights need to be updated.

Tabrizian (approved):

include/triton/core/tritonserver.h
rmccorm4: I don't mind removing it, but I think it's more clear to the user to have "deprecated" included to stand out more. We do so in some other places as well, ex: https://github.com/triton-inference-server/core/blob/46eb9847a096c1d1a8cfa93c528991de926246f5/include/triton/core/tritonserver.h#L680-L683

include/triton/core/tritoncache.h
rmccorm4: done

src/cache_entry.cc
GuanLuo: We can but I think the allocation is temporary as it looks like there is plan to avoid the copy
```
/// NOTE: (DLIS-4471) Currently this will make a copy of the buffer at 'base'
/// to avoid lifetime issues.
```

src/cache_entry.h
```diff
@@ -0,0 +1,129 @@
...
+  ~CacheEntryItem();
+  Status FromResponse(const InferenceResponse* response);
+  Status ToResponse(InferenceResponse* response);
+  std::vector<Buffer> Buffers();
+  size_t BufferCount();
+  void AddBuffer(boost::span<const std::byte> buffer);
+  void AddBuffer(const void* base, size_t byte_size);
+  void AddBuffer(std::pair<void*, size_t> buffer_pair);
+  void AddBuffer(std::pair<void*, size_t> buffer_pair, bool copy);
+  void AddBuffer(void* base, size_t byte_size, bool copy);
```
GuanLuo, line 90: Are we going to use all overloads?

src/cache_entry.cc
```diff
@@ -0,0 +1,354 @@
...
+CacheEntryItem::AddBuffer(void* base, size_t byte_size, bool copy)
+{
+  std::unique_lock lk(buffer_mu_);
+  if (copy) {
+    void* new_base = malloc(byte_size);
+    memcpy(new_base, base, byte_size);
+    buffers_.emplace_back(std::make_pair(new_base, byte_size));
+  } else {
+    buffers_.emplace_back(std::make_pair(base, byte_size));
+  }
```
GuanLuo, line 93: Call above function if copy?
```suggestion
  if (copy) {
    AddBuffer(base, byte_size);
  } else {
    std::unique_lock lk(buffer_mu_);
    buffers_.emplace_back(std::make_pair(base, byte_size));
  }
```

src/cache_manager.cc
```diff
@@ -0,0 +1,422 @@
...
+    LOG_ERROR << "cache finalize function is nullptr";
+  }
+
+  ClearHandles();
+}
+
+void
+TritonCache::ClearHandles()
+{
+  dlhandle_ = nullptr;
```
GuanLuo, line 97: Need to close the library handle via SharedLibrary API

src/dynamic_batch_scheduler.cc
```diff
@@ -601,31 +611,34 @@ DynamicBatchScheduler::DelegateResponse(
...
           std::unique_ptr<InferenceResponse>&& response, const uint32_t flags) {
-        if (response_cache_enabled_ && raw_request_ptr->CacheKeyIsSet()) {
+        if (response_cache_enabled_) {
           // Cache insertion happens here because we need the backend to have
           // computed the inference response first in the case of cache miss
-          auto cache = model_->Server()->GetResponseCache();
-          auto status = cache->Insert(*response, raw_request_ptr);
+          auto cache = model_->Server()->CacheManager()->Cache();
+          auto insert_start_ns = CaptureTimeNs();
+          auto status = cache->Insert(response.get(), key);
```
GuanLuo, line 624: I don't see handling for "holding" responses for decoupled model, is it going to be in a separate PR?

GuanLuo (commented):
Minor comments

src/dynamic_batch_scheduler.cc
rmccorm4: No decoupled cache support before or after this PR, this is parity with existing cache, just moving to use of C API instead.

src/cache_entry.h
rmccorm4: removed one unused overload

src/cache_entry.cc
rmccorm4: Re-used above function as suggested, renamed the `const void*` overloads to `AddBufferCopy` to be more clear.

src/cache_manager.cc
rmccorm4: done

GuanLuo (approved):

PR #145, opened 2022-12-14
Add load serialization per backend to needed backends

kthui:
There are backends which does not fully support parallel loading yet. This PR is to serialize their loading before they fully support parallel loading.

### DIFF
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -55,6 +55,10 @@
 
 namespace triton { namespace core {
 
+// FIXME: remove when no longer needed DLIS-4415
+std::map<std::string, std::mutex> TritonModel::non_concurrent_backends_;
+std::mutex TritonModel::non_concurrent_backends_mu_;
+
 Status
 TritonModel::Create(
     InferenceServer* server, const std::string& model_path,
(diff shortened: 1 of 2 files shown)


kthui:
Before the change, hang on model loading and segmentation fault on model unloading was observed by running the L0_infer cpu-only repeatedly with this script:
```
#!/bin/bash

export TRITON_SERVER_CPU_ONLY=1
export BACKENDS="onnx python graphdef savedmodel libtorch"

for (( i=1; i<=1024; i++ )); do
        set +e
        rm -rf *log models *txt
        bash -x test.sh
        if [ $? -ne 0 ]; then
                echo -e "\n***\n*** Failed\n***"
                exit 1
        fi
        set -e
done
```
After this change, the issues are no longer observable.

GuanLuo (commented):
Why this is done within core? Core shouldn't care if individual backend work well with parallel model loading. The backend developer must be aware of the parallel nature and serialize it properly if the backend must do so in sequence.

kthui:
> Why this is done within core? Core shouldn't care if individual backend work well with parallel model loading. The backend developer must be aware of the parallel nature and serialize it properly if the backend must do so in sequence.

This is a patch to get everything going first. I agree we should find the root cause of the issue within each of the affected backend for a permanent fix. This is a ticket to remove this patch after the backends are fixed: DLIS-4415.

kthui:
Closed due to https://github.com/triton-inference-server/core/pull/146

PR #149, opened 2022-12-28
Added cache to custom batching functions

dyastremsky:
This PR builds on #143. For custom batching where the user defines additional constraints for the dynamic batcher, this PR allows the user to call two new optional functions that initialize and finalize once. This allows them to create a cached object that can be read for every batch without needing to be re-initialized.

An example use case is where the user needs to read a value from the model config for batching, so they would only want to read it from the config once and save it to use with their custom batch-handling logic. See volume_batching.cc in this PR as an example.

The core PR adds the loading of shared library functions, saving of the cached user pointer, and initialization/finalization calls as needed.

Backend: https://github.com/triton-inference-server/backend/pull/76
Server: https://github.com/triton-inference-server/server/pull/5204

### DIFF
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -1441,10 +1441,10 @@ TRITONSERVER_Error* TRITONBACKEND_ModelBatchFinalize(void* userp);
 /// model loading. The batcher will point to a user-defined data structure that
 /// holds read-only data used for custom batching.
 ///
-/// \param model The backend model for which Triton is forming a batch.
 /// \param batcher User-defined placeholder for backend to store and
 /// retrieve information about the batching strategy for this model.
 /// return a TRITONSERVER_Error indicating success or failure.
+/// \param model The backend model for which Triton is forming a batch.
 TRITONSERVER_Error* TRITONBACKEND_BatcherInitialize(
     TRITONBACKEND_Batcher** batcher, TRITONBACKEND_Model* model);
 
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -1416,12 +1416,14 @@ TRITONBACKEND_BackendAttributeAddPreferredInstanceGroup(
 /// \param userp The placeholder for backend to store and retrieve information
 /// about this pending batch. When the callback returns, this should reflect
 /// the latest batch information.
+/// \param cache_userp The read-only placeholder for backend to retrieve
+// information about the batching strategy for this model.
 /// \param should_include The pointer to be updated on whether the request
 /// should be included in the batch.
 /// \return a TRITONSERVER_Error indicating success or failure.
 TRITONSERVER_Error* TRITONBACKEND_ModelBatchIncludeRequest(
     TRITONBACKEND_Model* model, TRITONBACKEND_Request* request, void* userp,
-    bool* should_include);
+    const void* cache_userp, bool* should_include);
 
 /// Callback to be invoked when Triton has begun forming a batch.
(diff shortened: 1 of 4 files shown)


include/triton/core/tritonbackend.h
```diff
@@ -1427,16 +1427,34 @@ TRITONSERVER_Error* TRITONBACKEND_ModelBatchIncludeRequest(
 /// \param model The backend model for which Triton is forming a batch.
 /// \param userp The placeholder for backend to store and retrieve information
 /// about this pending batch.
+/// \param cache_userp The read-only placeholder for backend to retrieve
+// information about the batching strategy for this model.
 /// \return a TRITONSERVER_Error indicating success or failure.
 TRITONSERVER_Error* TRITONBACKEND_ModelBatchInitialize(
-    TRITONBACKEND_Model* model, void** userp);
+    TRITONBACKEND_Model* model, void** userp, const void* cache_userp);
```
GuanLuo, line 1434: Since you are adding the new APIs to store persistent batching info, you can wrap the "cache" in the way similar to [cache API](https://github.com/triton-inference-server/core/pull/142/files#diff-83cc1e14e9b115d1acaed238f93f014fc99fee507d0f467b0afd59669cb5bf4aR209-R238), I would call the implementation defined type to `TRITONBACKEND_Batcher` by the way. And because in the "new" function, you already pass `model` so that the implementation can retrieve information needed upfront. the signature for `TRITONBACKEND_ModelBatchInitialize` can be simplified to
```suggestion
    TRITONBACKEND_Batcher* batcher, void** userp);
```

include/triton/core/tritonbackend.h
```diff
@@ -1427,16 +1427,34 @@ TRITONSERVER_Error* TRITONBACKEND_ModelBatchIncludeRequest(
...
+/// information about the batching strategy for this model.
+/// return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_Error* TRITONBACKEND_ModelBatchCacheInitialize(
+    TRITONBACKEND_Model* model, void** cache_userp);
+
+/// Callback to be invoked when Triton unloads model.
+/// \param cache_userp The placeholder for backend to store and retrieve
+/// information about the batching strategy for this model.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_Error* TRITONBACKEND_ModelBatchCacheFinalize(void* cache_userp);
```
GuanLuo, line 1456: You should talk about the lifecycle of this object, "init" called after model (even instance?) init and "final" before model (instance?) final

nv-kmcgill53 (approved):

include/triton/core/tritonbackend.h
dyastremsky: Thanks for this suggestion, Guan.

Are you suggesting replacing the (optional) InitializeCache functions with (mandatory) `TRITONBACKEND_BatcherNew`? Then, creating a separate `TRITONBACKEND_Batcher` struct that the user defines (but needs to store the model)?

dyastremsky: Thanks for chatting about this. Updated to use `TRITONBACKEND_Batcher` and make its initialization/finalization implementation required.

include/triton/core/tritonbackend.h
dyastremsky: Clarified the comments to make this clearer. Init/finalize is called during model init and finialization.

dyastremsky:
To try to get this in quickly, I'm going to merge this into the main PR branch. All feedback should have been addressed. :) If anyone has additional feedback, please comment it in the main dyas-custom-batching PR.

PR #151, opened 2022-12-31
 Return the composing model error failure on ensemble load

Tabrizian:
- [X] Return error on cudaStream create failure.
- [X] Return the composing model error failure on ensemble load.

### DIFF
--- a/src/ensemble_scheduler.cc
+++ b/src/ensemble_scheduler.cc
@@ -1,4 +1,4 @@
-// Copyright 2019-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2019-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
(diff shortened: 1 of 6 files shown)


src/ensemble_scheduler.cc
```diff
@@ -1269,7 +1269,13 @@ EnsembleScheduler::Create(
     InferenceServer* const server, const inference::ModelConfig& config,
     std::unique_ptr<Scheduler>* scheduler)
 {
-  scheduler->reset(new EnsembleScheduler(stats_aggregator, server, config));
+  cudaStream_t stream = nullptr;
+#ifdef TRITON_ENABLE_GPU
+  RETURN_IF_ERROR(
+      CreateCudaStream(&stream, server->MinSupportedComputeCapability()));
```
GuanLuo, line 1275: Do we need to fail the ensemble creation if it fails to create its own CUDA stream? I think it should still function without dedicated CUDA stream? Unless the error is caused by sticky CUDA error which will fail the inference anyway.

Tabrizian: I agree. I have removed that from the PR.

tanmayv25 (approved):

GuanLuo (approved):

PR #148, opened 2022-12-21
fallback to default GCS client if json credentials file were not found

amircodota:
We store our model repository in GCS.

We upgraded from triton 22.07 to 22.11.

In 22.07 it worked correctly, but in 22.11 it stopped working.

We run our triton server on GKE, and use [workload identity](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity) to connect to GCS. Unfortunately, there's no json file credentials created in this auth method.

PR #152, opened 2023-01-04
Fix GCS public bucket access

kthui:
This is a placeholder PR for reviewing the change that enables accessing GCS public buckets anonymously (without credentials). After approval, #148 should be merged instead.


kthui:
The first commit [b1b4b64](https://github.com/triton-inference-server/core/pull/152/commits/d1b4b6414ce38c5b6c0187a264ab21ac973b13c8) is authored by #148

GuanLuo (approved):

PR #153, opened 2023-01-05
Instance Group Storage Refactor

nv-kmcgill53:
This refactor is the first step for being able to manipulate instance groups without the need to completely unload and reload the model.

### DIFF
--- a/src/backend_model.h
+++ b/src/backend_model.h
@@ -125,7 +125,7 @@ class TritonModel : public Model {
 
   // The model instances for this model stored by the
   // instance groups defined in the model configuration.
-  // Passive instance groups are those instances which are 
+  // Passive instance groups are those instances which are
   // loaded but not added to the scheduler.
   std::map<std::string, TritonInstanceGroup> instance_group_map_;
   std::map<std::string, TritonInstanceGroup> passive_instance_group_map_;
--- a/src/backend_model.h
+++ b/src/backend_model.h
@@ -66,7 +66,7 @@ class TritonModel : public Model {
       const uint32_t config_version,
       TRITONSERVER_Message* updated_config_message);
   const std::shared_ptr<TritonBackend>& Backend() const { return backend_; }
-  const std::map<std::string, TritonInstanceGroup>& InstanceGroups() const
+  const std::unordered_map<std::string, TritonInstanceGroup>& InstanceGroups() const
   {
     return instance_group_map_;
   }
@@ -127,8 +127,8 @@ class TritonModel : public Model {
   // instance groups defined in the model configuration.
   // Passive instance groups are those instances which are
   // loaded but not added to the scheduler.
-  std::map<std::string, TritonInstanceGroup> instance_group_map_;
-  std::map<std::string, TritonInstanceGroup> passive_instance_group_map_;
+  std::unordered_map<std::string, TritonInstanceGroup> instance_group_map_;
+  std::unordered_map<std::string, TritonInstanceGroup> passive_instance_group_map_;
 
   // Opaque state associated with this model.
   void* state_;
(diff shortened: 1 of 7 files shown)


src/backend_model.h
```diff
@@ -123,8 +124,8 @@ class TritonModel : public Model {
   std::shared_ptr<TritonBackend> backend_;
 
   // The model instances for this model.
```
dyastremsky, line 126: Unrelated to this ticket, but general quality-of-life: it'd be great to clarify in this comment (or in AddInstance) what passive means  here. No need to update it here, but if you're working on these files, would be awesome to clarify it in this or a future PR.

src/sequence_batch_scheduler.cc
```diff
@@ -57,7 +57,11 @@ SequenceBatchScheduler::Create(
              << " backlog queued requests...";
   }
 
-  auto instance_count = model->Instances().size();
+  uint64_t instance_count = 0;
+  for (const auto& pair : model->InstanceGroups()) {
+    instance_count += pair.second.size();
+  }
+  // auto instance_count = model->Instances().size();
```
dyastremsky, line 64: Did you mean to remove this?

dyastremsky (changes requested):

src/sequence_batch_scheduler.cc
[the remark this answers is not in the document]
nv-kmcgill53: I did, removed.

src/backend_model.h
nv-kmcgill53: Updated the comment (hopefully) clarifying the storage container and the definition of the passive instance groups.

dyastremsky (dismissed):

nv-kmcgill53:
Note: `filesystem.h` and `model_config_utils.h` only have clang formatting changes.

src/backend_model.h
```diff
@@ -122,9 +123,12 @@ class TritonModel : public Model {
...
 
-  // The model instances for this model.
-  std::vector<std::unique_ptr<TritonModelInstance>> instances_;
-  std::vector<std::unique_ptr<TritonModelInstance>> passive_instances_;
+  // The model instances for this model stored by the
+  // instance groups defined in the model configuration.
+  // Passive instance groups are those instances which are
+  // loaded but not added to the scheduler.
+  std::map<std::string, TritonInstanceGroup> instance_group_map_;
+  std::map<std::string, TritonInstanceGroup> passive_instance_group_map_;
```
rmccorm4, line 131: Do the `map`s need to be ordered?

src/backend_model.h
```diff
@@ -65,9 +66,9 @@ class TritonModel : public Model {
       const uint32_t config_version,
       TRITONSERVER_Message* updated_config_message);
   const std::shared_ptr<TritonBackend>& Backend() const { return backend_; }
-  const std::vector<std::unique_ptr<TritonModelInstance>>& Instances() const
+  const std::map<std::string, TritonInstanceGroup>& InstanceGroups() const
```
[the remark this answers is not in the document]
nv-kmcgill53: This map doesn't need to be ordered. It ends up being mostly constant once the model is loaded so we mostly read from it. In the context of hot instance loading/unloading, for now the group names will not change. We will assert the instance groups are in the same order and identical as in the previous model config (except for the `count`). Therefore, we will not have any insertions/deletions into the map. All that said, you are right in that we should not incur debt where not needed and I will fix to `unordered_map`.

src/backend_model.h
nv-kmcgill53: See [this comment](https://github.com/triton-inference-server/core/pull/153#discussion_r1063758281)

rmccorm4 (approved):

dyastremsky (approved):

PR #156, opened 2023-01-15
Fix model loading hang

Tabrizian:


rmccorm4 (dismissed):
What recently introduced this deadlock?

Tabrizian:
@rmccorm4 Thanks for the quick review! Theoretically, this could have happened since we added support for thread pool in model loading. It could be that some changes in CUDA 12 have affected the timing and it has exposed this issue. There is a thread in the teams channel about why this could happen.

rmccorm4:
> Theoretically, this could have happened since we added support for thread pool in model loading
Interesting. I believe even before the model-load thread pool change, new threads were spawned for each model load, so I think more or less the same level of concurrency existed before as well.

Tabrizian:
I see, as long as we were using multiple threads for loading the models and removed persistent backends (i.e., backends that were always loaded if existed) this issue existed.

rmccorm4 (dismissed):

rmccorm4:
- Do we have a ticket for a more permanent fix? This seems more like a band-aid fix for now.

src/backend_manager.cc
```diff
@@ -70,6 +74,16 @@ TritonBackend::Create(
...
   TritonServerMessage backend_config(backend_config_json);
 
+#ifdef TRITON_ENABLE_GPU
+  // This call is to prevent a deadlock issue with dlopening backend shared
+  // libraries and calling CUDA APIs in other threads. Since CUDA API also
+  // dlopens some libraries and dlopen has an internal lock, it can create a
+  // deadlock. Intentionally ignore the CUDA_ERROR (if any) for containers
+  // running on a CPU.
+  int device_count;
+  cudaGetDeviceCount(&device_count);
```
GuanLuo, line 84: Should this be moved to shared library utils to make sure all shared libraries ops have CUDA init before accessing any libraries?

Tabrizian:
> Do we have a ticket for a more permanent fix? This seems more like a band-aid fix for now.

There is not much we can do about this bug. It seems like a problem with dlopen/CUDA.

GuanLuo (approved):

rmccorm4 (approved):

PR #157, opened 2023-01-17
Map cache name to library and config

rmccorm4:
This enables user to specify which cache implementation they want to use, for example `local` or `redis`, and specify them on the CLI in the related server PR: https://github.com/triton-inference-server/server/pull/5119.

The `cache_name` must match a folder in the CACHE_DIR (ex: `/opt/tritonserver/caches/`) similar to how backends and repoagents are loaded and configured today.

> **NOTE**:
> There may only be one cache configured (and therefore instantiated) at this time. In the future, there may be support for multiple global caches such as a `local` and a `redis` cache co-existing with no API changes, only minor changes to `TritonCacheManager` and how it's interacted with.

### DIFF
--- a/src/tritonserver.cc
+++ b/src/tritonserver.cc
@@ -308,9 +308,9 @@ class TritonServerOptions {
     return host_policy_map_;
   }
 
-  const std::string& CacheConfig() const { return cache_config_; }
-  void SetCacheConfig(const std::string& cfg) { cache_config_ = cfg; }
-
+  tc::CacheConfigMap CacheConfig() { return cache_config_map_; }
+  TRITONSERVER_Error* AddCacheConfig(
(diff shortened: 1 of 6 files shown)


src/tritonserver.cc
```diff
@@ -308,9 +308,9 @@ class TritonServerOptions {
     return host_policy_map_;
   }
 
-  const std::string& CacheConfig() const { return cache_config_; }
-  void SetCacheConfig(const std::string& cfg) { cache_config_ = cfg; }
-
+  tc::CacheConfigMap CacheConfig() { return cache_config_map_; }
```
Tabrizian, line 311: Use const&

Tabrizian (approved):

PR #161, opened 2023-01-18
Remove global cache metric polling from core

rmccorm4:
- Removes the polling cache metrics from Triton core (total hits/misses, utillization, num entries, etc.), these will be moved into the `local` cache implementation via the Metrics C API
- The per-model cache hits/misses used in infer_stats and perf analyzer will remain.

Corresponding docs PR: https://github.com/triton-inference-server/server/pull/5261

### DIFF
--- a/src/dynamic_batch_scheduler.cc
+++ b/src/dynamic_batch_scheduler.cc
@@ -634,23 +634,25 @@ DynamicBatchScheduler::DelegateResponse(
   completion_queue_.emplace_back();
   auto queue_slot = &completion_queue_.back();
   // Cache plumbing
-  auto key = request->CacheKey();
-  uint64_t lookup_ns =
-      request->CacheLookupEndNs() - request->CacheLookupStartNs();
-  if (!request->CacheKeyIsSet()) {
-    key = "";
-    LOG_ERROR << "Request cache key was not set correctly.";
-  }
-  if (request->CacheLookupStartNs() > request->CacheLookupEndNs()) {
-    lookup_ns = 0;
-    LOG_ERROR << "Request lookup duration was not set correctly.";
-  }
+  const std::string& key = request->CacheKey();
+  const bool is_key_set = request->CacheKeyIsSet();
+  const uint64_t lookup_end_ns = request->CacheLookupEndNs();
+  const uint64_t lookup_start_ns = request->CacheLookupStartNs();
 
   request->SetResponseDelegator(
-      [this, queue_slot, key, lookup_ns](
-
+      [this, queue_slot, key, is_key_set, lookup_end_ns, lookup_start_ns](
           std::unique_ptr<InferenceResponse>&& response, const uint32_t flags) {
         if (response_cache_enabled_) {
+          // Logical error checks, these shouldn't happen
+          if (!is_key_set) {
+            LOG_ERROR << "Request cache key was not set correctly.";
+          }
+          uint64_t lookup_ns = lookup_end_ns - lookup_start_ns;
+          if (lookup_start_ns > lookup_end_ns) {
+            lookup_ns = 0;
+            LOG_ERROR << "Request lookup duration was not set correctly.";
+          }
+
           // Cache insertion happens here because we need the backend to have
           // computed the inference response first in the case of cache miss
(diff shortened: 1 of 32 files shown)


src/dynamic_batch_scheduler.cc
```diff
@@ -633,31 +633,41 @@ DynamicBatchScheduler::DelegateResponse(
...
   completion_queue_.emplace_back();
   auto queue_slot = &completion_queue_.back();
-  const auto key = request->CacheKey();
+  // Cache plumbing
+  auto key = request->CacheKey();
+  uint64_t lookup_ns =
+      request->CacheLookupEndNs() - request->CacheLookupStartNs();
+  if (!request->CacheKeyIsSet()) {
+    key = "";
+    LOG_ERROR << "Request cache key was not set correctly.";
```
Tabrizian, line 642: Does this error gets printed when the model is not using any caching too?

src/shared_library.cc
```diff
@@ -107,6 +111,16 @@ SharedLibrary::OpenLibraryHandle(const std::string& path, void** handle)
...
 
+#ifdef TRITON_ENABLE_GPU
+  // This call is to prevent a deadlock issue with dlopening backend shared
+  // libraries and calling CUDA APIs in other threads. Since CUDA API also
+  // dlopens some libraries and dlopen has an internal lock, it can create a
+  // deadlock. Intentionally ignore the CUDA_ERROR (if any) for containers
+  // running on a CPU.
+  int device_count;
+  cudaGetDeviceCount(&device_count);
+#endif
```
Tabrizian, line 122: I think you need to rebase with `main`.

src/metric_model_reporter.h
```diff
@@ -128,10 +125,9 @@ class MetricModelReporter {
   prometheus::Counter* metric_inf_compute_infer_duration_us_;
   prometheus::Counter* metric_inf_compute_output_duration_us_;
   prometheus::Counter* metric_cache_hit_count_;
-  prometheus::Counter* metric_cache_hit_lookup_duration_us_;
+  prometheus::Counter* metric_cache_hit_duration_us_;
   prometheus::Counter* metric_cache_miss_count_;
-  prometheus::Counter* metric_cache_miss_lookup_duration_us_;
-  prometheus::Counter* metric_cache_miss_insertion_duration_us_;
+  prometheus::Counter* metric_cache_miss_duration_us_;
```
Tabrizian, line 130: Are we removing one of the metrics from the cache API? Looks like we no longer have `metric_cache_miss_insertion_duration_us_`.

rmccorm4: Under the hood, these metrics were combined in the `infer_stats` endpoint (what's reported to perf_analyzer)

cache_miss insert+lookup are being combined into one for consistency `cache_hit` metric, consistency with what infer_stats reports, and less clutter in metrics output.

src/shared_library.cc
rmccorm4: This should be resolved when `rmccormick-cache-api` -> `rmccormick-cache-main`

src/dynamic_batch_scheduler.cc
rmccorm4: Good catch. I think you're right, but it's actually surprisingly hard to hit this code path with caching disabled, so I didn't see it. 

Just as fun fact, it looks like you would need to have:
- dynamic batching disabled (which means you need `--disable-auto-complete-config` because you can't specify default scheduler in config file and autocomplete chooses dynamic_batching) 
- `preserve_ordering` to be true, but it is a [property of dynamic_batching](https://github.com/triton-inference-server/common/blob/5313d0be99204c89ab696023534043f2ba535868/protobuf/model_config.proto#L1104-L1135) ? So I don't understand how this path would get hit. This is true even in 21.09 before cache was added: https://github.com/triton-inference-server/server/blob/39b93374b88ecba2c4912323ed6406c13d0bc6b7/src/core/dynamic_batch_scheduler.cc#L153-L156

rmccorm4: Moved error checking logic into cache_enabled section. These checks indicate a bug if they are hit, so it shouldn't be spamming users unless something is wrong.

Tabrizian (approved):

PR #162, opened 2023-01-26
Add ifdefs around stats, and no-op for cache size of zero with old API

rmccorm4:


src/tritonserver.cc
```diff
@@ -1182,7 +1182,14 @@ TRITONAPI_DECLSPEC TRITONSERVER_Error*
 TRITONSERVER_ServerOptionsSetResponseCacheByteSize(
     TRITONSERVER_ServerOptions* options, uint64_t size)
 {
-  // For backwards compatibility, forward this API call to new CacheConfig API.
+  // Setting a cache config is equivalent to enabling the cache, so triton will
+  // attempt to create it. To maintain old behavior, simply return and do
+  // nothing for a cache size of zero.
+  if (size == 0) {
+    return nullptr;  // success
+  }
```
rmccorm4, line 1190: Regarding this part, downstream tests for the Triton Developer Tools / C++ call the `TRITONSERVER_SetResponseCacheByteSize(0)` with a sizer of zero by default, which internally was enabling the cache. To match old behavior where a non-zero size was required to enable cache, I made this change.

Tabrizian (approved):

PR #164, opened 2023-02-10
Enable GCS compute engine credentials

kthui:
GCS supports a range of different authentication method for connecting to the service, specifically we want to support [JSON file](https://googleapis.dev/cpp/google-cloud-storage/1.42.0/namespacegoogle_1_1cloud_1_1storage_1_1oauth2.html#a8e14b8e892425fcc5c9cdad80120eead), [Compute Engine](https://googleapis.dev/cpp/google-cloud-storage/1.42.0/namespacegoogle_1_1cloud_1_1storage_1_1oauth2.html#ad47cf9331fa6c0999407b1da378368d0) and [Anonymous](https://googleapis.dev/cpp/google-cloud-storage/1.42.0/namespacegoogle_1_1cloud_1_1storage_1_1oauth2.html#a086aa0ad3dad05d6a500252ba144dff7) credentials. This PR adds the Compute Engine credential which automatically connects to the metadata server and fetch an access token for accessing GCS (a.k.a. identity mechanism), when inside Kubernetes or Compute Engine if properly setup.

If the JSON credential is not supplied (either via [env var](https://github.com/triton-inference-server/server/blob/main/docs/user_guide/model_repository.md#google-cloud-storage) or [credential file](https://github.com/triton-inference-server/server/blob/main/docs/user_guide/model_repository.md#cloud-storage-with-credential-file-beta)), then the Compute Engine credential is tried next. If the Compute Engine credential is unable to fetch [AuthorizationHeader](https://googleapis.dev/cpp/google-cloud-storage/1.42.0/classgoogle_1_1cloud_1_1storage_1_1oauth2_1_1ComputeEngineCredentials.html#a8c3a5d405366523e2f4df06554f0a676), then Anonymous credential is used.

Related PR: https://github.com/triton-inference-server/server/pull/5355
Closes https://github.com/triton-inference-server/server/issues/5039

### DIFF
--- a/src/filesystem.cc
+++ b/src/filesystem.cc
@@ -1,4 +1,4 @@
-// Copyright 2019-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2019-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
@@ -540,14 +540,18 @@ class GCSFileSystem : public FileSystem {
 
 GCSFileSystem::GCSFileSystem(const GCSCredential& gs_cred)
 {
+  google::cloud::Options options;
   auto creds = gcs::oauth2::CreateServiceAccountCredentialsFromJsonFilePath(
       gs_cred.path_);
   if (creds) {
-    client_ = gcs::Client(gcs::ClientOptions(*creds));
+    options.set<gcs::Oauth2CredentialsOption>(*creds);  // json credential
   } else {
-    client_ = gcs::Client(
-        gcs::ClientOptions(gcs::oauth2::CreateAnonymousCredentials()));
+    options.set<google::cloud::UnifiedCredentialsOption>(
+        google::cloud::MakeGoogleDefaultCredentials());  // metadata service
+    options.set<google::cloud::UnifiedCredentialsOption>(
+        google::cloud::MakeInsecureCredentials());  // public bucket
   }
+  client_ = gcs::Client(options);
 }
 
 Status
--- a/src/filesystem.cc
+++ b/src/filesystem.cc
@@ -535,7 +535,7 @@ class GCSFileSystem : public FileSystem {
       const std::string path, bool* exists,
       google::cloud::StatusOr<gcs::ObjectMetadata>* metadata);
 
-  google::cloud::StatusOr<gcs::Client> client_;
+  std::unique_ptr<gcs::Client> client_;
 };
 
 GCSFileSystem::GCSFileSystem(const GCSCredential& gs_cred)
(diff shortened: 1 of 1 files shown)


kthui:
Before patch on GCP compute engine, 23.01 container:
```
/opt/tritonserver# tritonserver --model-repository=${BUCKET_URL}/model_repo                 
...
error: creating server: Internal - Could not get MetaData for bucket with name triton-bucket-41498862 : Permanent error in GetBucketMetadata: {
  "error": {
    "code": 401,
    "message": "Anonymous caller does not have storage.buckets.get access to the Google Cloud Storage bucket. Permission 'storage.buckets.get' denied on resource (or it may not exist).",
    "errors": [
      {
        "message": "Anonymous caller does not have storage.buckets.get access to the Google Cloud Storage bucket. Permission 'storage.buckets.get' denied on resource (or it may not exist).",
        "domain": "global",
        "reason": "required",
        "locationType": "header",
        "location": "Authorization"
      }
    ]
  }
}
```
After patch on 23.01 container:
```
/opt/tritonserver# tritonserver --model-repository=${BUCKET_URL}/model_repo
...
I0213 22:02:49.502805 1951 server.cc:633] 
+-------------+---------+--------+
| Model       | Version | Status |
+-------------+---------+--------+
| simple_int8 | 1       | READY  |
+-------------+---------+--------+
...
I0213 22:02:49.561809 1951 grpc_server.cc:4868] Started GRPCInferenceService at 0.0.0.0:8001
I0213 22:02:49.562049 1951 http_server.cc:3477] Started HTTPService at 0.0.0.0:8000
I0213 22:02:49.603095 1951 http_server.cc:184] Started Metrics Service at 0.0.0.0:8002
```
Inference also works after patch:
```
/opt/tritonserver# python3 client.py
b'{"id":"1","model_name":"simple_int8","model_version":"1","outputs":[{"name":"OUTPUT0","datatype":"INT8","shape":[1,16],"data":[1,3,5,7,9,11,13,15,17,19,21,23,25,27,29,31]},{"name":"OUTPUT1","datatype":"INT8","shape":[1,16],"data":[-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1]}]}'
```

rmccorm4:
Changes look good to me. Just to clarify, if a user uses either credentials file or compute engine metadata service, they can access both private (their own) and public buckets?

kthui:
If the credentials, from file or metadata service, is allowed by GCS to access the buckets (allowed on private means: the credentials have all permissions; and on public means: the credentials have read and list permission from the bucket up to all objects), then yes, I do not see any reason why it would not work.

kthui:
The only part I think that might be problematic is the `creds->AuthorizationHeader() == true` but it should be `false`, maybe due to the user have some service at "metadata.google.internal" while not on Google Cloud, which is very unusual. Then, Triton will try to use compute engine metadata service, but there is no metadata service, which is why I state very clearly on the doc that "Triton will use credentials from the attached service account providing a value for the Authorization HTTP header can be obtained"

Tabrizian (approved):

rmccorm4 (approved):

PR #165, opened 2023-02-14
Add parallel model loading

kthui:
Related PRs: https://github.com/triton-inference-server/server/pull/5360, https://github.com/triton-inference-server/onnxruntime_backend/pull/174
Closes: https://github.com/triton-inference-server/server/issues/5342

### DIFF
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -174,6 +174,9 @@ TritonModel::Create(
     // case the backend library attempts to load additional shared libaries.
     // Currently, the set and reset function is effective only on Windows, so
     // there is no need to set path on non-Windows.
+    // However, parallel model loading will not see any speedup on Windows and
+    // the global lock inside the SharedLibrary is a WAR.
+    // [FIXME] Reduce lock WAR on SharedLibrary (DLIS-4300)
 #ifdef _WIN32
     std::unique_ptr<SharedLibrary> slib;
     RETURN_IF_ERROR(SharedLibrary::Acquire(&slib));
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -167,19 +167,25 @@ TritonModel::Create(
 
   TritonModel* raw_local_model = local_model.get();
 
-  // Model initialization is optional... The TRITONBACKEND_Model
-  // object is this TritonModel object. We must set set shared library
-  // path to point to the backend directory in case the backend
-  // library attempts to load additional shared libaries.
+  // Model initialization is optional... The TRITONBACKEND_Model object is this
+  // TritonModel object.
   if (backend->ModelInitFn() != nullptr) {
+    // We must set set shared library path to point to the backend directory in
+    // case the backend library attempts to load additional shared libaries.
+    // Currently, the set and reset function is effective only on Windows, so
+    // there is no need to set path on non-Windows.
+#ifdef _WIN32
     std::unique_ptr<SharedLibrary> slib;
     RETURN_IF_ERROR(SharedLibrary::Acquire(&slib));
     RETURN_IF_ERROR(slib->SetLibraryDirectory(backend->Directory()));
+#endif
 
     TRITONSERVER_Error* err = backend->ModelInitFn()(
         reinterpret_cast<TRITONBACKEND_Model*>(raw_local_model));
 
+#ifdef _WIN32
     RETURN_IF_ERROR(slib->ResetLibraryDirectory());
+#endif
     RETURN_IF_TRITONSERVER_ERROR(err);
   }
 
--- a/src/model_repository_manager.cc
+++ b/src/model_repository_manager.cc
@@ -310,7 +310,7 @@ ModelRepositoryManager::ModelRepositoryManager(
     : autofill_(autofill), polling_enabled_(polling_enabled),
       model_control_enabled_(model_control_enabled),
       min_compute_capability_(min_compute_capability),
-      dependency_graph_(global_map_),
+      dependency_graph_(&global_map_),
       enable_model_namespacing_(enable_model_namespacing),
       repository_paths_(repository_paths),
       model_life_cycle_(std::move(life_cycle))
@@ -396,6 +396,7 @@ ModelRepositoryManager::Create(
     for (const auto& model_name : startup_models) {
       models[model_name];
     }
+    std::lock_guard<std::mutex> lock((*model_repository_manager)->poll_mu_);
     RETURN_IF_ERROR(
         (*model_repository_manager)
             ->LoadUnloadModels(
@@ -476,20 +477,20 @@ ModelRepositoryManager::PollAndUpdateInternal(bool* all_models_polled)
 
   infos_.swap(new_infos);
 
-  UpdateDependencyGraph(added, deleted, modified);
+  UpdateDependencyGraph(&dependency_graph_, infos_, added, deleted, modified);
 
   for (const auto& name : deleted) {
     model_life_cycle_->AsyncUnload(name);
   }
 
   // model loading / unloading error will be printed but ignored
-  LoadModelByDependency();
+  LoadModelByDependency(false /* release_mu */);
 
   return Status::Success;
 }
 
 std::map<ModelIdentifier, Status>
-ModelRepositoryManager::LoadModelByDependency()
+ModelRepositoryManager::LoadModelByDependency(bool release_mu)
 {
   // [FIXME] This function involves iterative interaction between
   // ModelLifeCycle object and DependencyGraph object, should better
@@ -507,63 +508,82 @@ ModelRepositoryManager::LoadModelByDependency()
   //  - Reflect the lifecycle change result to the dependency graph so that the
   //    downstream nodes may be ready for lifecycle changes.
   // Repeat until no more changes can be made.
+  // Dependency node pointer can be swapped when the mutex is released, so any
+  // pointers must be re-acquired from its model identifier afterward.
   std::map<ModelIdentifier, Status> res;
   struct ModelState {
-    ModelState(DependencyNode* node) : node_(node), status_(Status::Success) {}
-    DependencyNode* node_;
+    ModelState(const ModelIdentifier& model_id)
+        : model_id_(model_id), status_(Status::Success)
+    {
+    }
+    ModelIdentifier model_id_;
     Status status_;
     std::promise<void> ready_;
   };
-  NodeSet loaded_models;
+  std::set<ModelIdentifier> loaded_models;
   auto set_pair = ModelsToLoadUnload(loaded_models, res);
   // Loop until all model are loaded / unloaded
   while ((!set_pair.first.empty()) || (!set_pair.second.empty())) {
     loaded_models.clear();
     // Unload invalid models first
-    for (auto& invalid_model : set_pair.second) {
-      model_life_cycle_->AsyncUnload(invalid_model->model_id_);
-      res[invalid_model->model_id_] = invalid_model->status_;
-      LOG_ERROR << invalid_model->status_.AsString();
-      invalid_model->loaded_versions_ = std::set<int64_t>();
-      loaded_models.emplace(invalid_model);
+    for (auto& invalid_id : set_pair.second) {
+      if (release_mu)
+        poll_mu_.unlock();
+      model_life_cycle_->AsyncUnload(invalid_id);
+      if (release_mu)
+        poll_mu_.lock();
+      auto invalid_node = dependency_graph_.FindNode(
+          invalid_id, false /* allow_fuzzy_matching */);
+      res[invalid_id] = invalid_node->status_;
+      LOG_ERROR << invalid_node->status_.AsString();
+      invalid_node->loaded_versions_ = std::set<int64_t>();
+      loaded_models.emplace(invalid_id);
     }
     // load valid models and wait for load results
     std::vector<std::unique_ptr<ModelState>> model_states;
-    for (auto& valid_model : set_pair.first) {
-      model_states.emplace_back(new ModelState(valid_model));
+    for (auto& valid_id : set_pair.first) {
+      model_states.emplace_back(new ModelState(valid_id));
       auto model_state = model_states.back().get();
-      const auto itr = infos_.find(valid_model->model_id_);
+      const auto itr = infos_.find(valid_id);
+      auto valid_node = dependency_graph_.FindNode(
+          valid_id, false /* allow_fuzzy_matching */);
       auto status = model_life_cycle_->AsyncLoad(
-          valid_model->model_id_, itr->second->model_path_,
-          valid_model->model_config_, itr->second->is_config_provided_,
-          itr->second->agent_model_list_, [model_state](Status load_status) {
+          valid_id, itr->second->model_path_, valid_node->model_config_,
+          itr->second->is_config_provided_, itr->second->agent_model_list_,
+          [model_state](Status load_status) {
             model_state->status_ = load_status;
             model_state->ready_.set_value();
           });
       if (!status.IsOk()) {
         model_state->status_ = status;
         model_state->ready_.set_value();
-        LOG_ERROR << "failed to load model '" << valid_model->model_id_.str()
+        LOG_ERROR << "failed to load model '" << valid_id.str()
                   << "': " << status.Message();
       }
-      loaded_models.emplace(valid_model);
+      loaded_models.emplace(valid_id);
     }
     for (auto& model_state : model_states) {
+      if (release_mu)
+        poll_mu_.unlock();
       model_state->ready_.get_future().wait();
-      res[model_state->node_->model_id_] = model_state->status_;
+      if (release_mu)
+        poll_mu_.lock();
+      res[model_state->model_id_] = model_state->status_;
       const auto version_state =
-          model_life_cycle_->VersionStates(model_state->node_->model_id_);
-      model_state->node_->loaded_versions_.clear();
+          model_life_cycle_->VersionStates(model_state->model_id_);
+      auto node = dependency_graph_.FindNode(
+          model_state->model_id_, false /* allow_fuzzy_matching */);
+      node->loaded_versions_.clear();
       for (const auto& vs : version_state) {
         if (vs.second.first == ModelReadyState::READY) {
-          model_state->node_->loaded_versions_.emplace(vs.first);
+          node->loaded_versions_.emplace(vs.first);
         }
       }
       // If the model failed to load, should revert the timestamp to
       // ensure the next load request will attempt to load the model again
       // for operation idempotence. See comment on 'infos_'
       if (!model_state->status_.IsOk()) {
-        auto& model_info = infos_.find(model_state->node_->model_id_)->second;
+        auto& model_info = infos_.find(model_state->model_id_)->second;
         model_info->mtime_nsec_ = model_info->prev_mtime_ns_;
       }
     }
@@ -675,7 +695,10 @@ ModelRepositoryManager::LoadUnloadModels(
 {
   auto status = Status::Success;
   *all_models_polled = true;
-  // Update ModelInfo related to file system accordingly
+  // Prepare ModelInfo related to file system accordingly, do not write yet
+  // until committed to load/unload the models.
+  ModelInfoMap infos;
+  CopyModelInfos(&infos);
   std::set<ModelIdentifier> added, deleted, modified, unmodified;
   {
     if (type == ActionType::UNLOAD) {
@@ -752,11 +775,11 @@ ModelRepositoryManager::LoadUnloadModels(
       // Only update the infos when all validation is completed
       for (const auto& model_name : added) {
         auto nitr = new_infos.find(model_name);
-        infos_.emplace(model_name, std::move(nitr->second));
+        infos.emplace(model_name, std::move(nitr->second));
       }
       for (const auto& model_name : modified) {
         auto nitr = new_infos.find(model_name);
-        auto itr = infos_.find(model_name);
+        auto itr = infos.find(model_name);
         // Also check the 'explicitly_load_' in previous snapshot,
         // if the model has been explicitly loaded, we shouldn't change
         // its state as it may be polled as composing model and flag is set
@@ -768,24 +791,64 @@ ModelRepositoryManager::LoadUnloadModels(
   }
   std::set<ModelIdentifier> deleted_dependents;
 
-  // Update dependency graph and load
-  UpdateDependencyGraph(
-      added, deleted, modified,
+  // Make a copy of the dependency graph for rollback if there is a collision
+  // between the models to be loaded/unloaded by this thread with another
+  // thread actively loading/unloading models. The global map referenced by the
+  // dependency graph is managed by this object, which is the responsibility of
+  // this object to make a copy.
+  std::unordered_map<std::string, std::set<ModelIdentifier>> global_map(
+      global_map_);
+  DependencyGraph dependency_graph(dependency_graph_, &global_map);
+
+  // Check if the prepared ModelInfo does not collide with any currently
+  // loading/unloading models, by trying to update dependency graph and mark
+  // loading/unloading and dependent models as is_locked. At this point, the
+  // prepared ModelInfo is not written to this, so it is safe to simply return
+  // and "rollback" if there is a collision.
+
+  // Update the dependency graph independently for the possibility of rollback.
+  auto affected_models = UpdateDependencyGraph(
+      &dependency_graph, infos, added, deleted, modified,
       unload_dependents ? &deleted_dependents : nullptr);
+
+  // Check for collision in affected models and lock those models
+  auto conflict_model = dependency_graph.LockUnlockNodes(
+      affected_models, true /* to_lock */,
+      &dependency_graph_ /* prev_dependency_graph */);
+  if (conflict_model) {
+    // A collision is found. Since info is only written to local copy, so it is
+    // safe to rollback by simply returning an error.
+    return Status(
+        Status::Code::INTERNAL,
+        "a related model '" + conflict_model->str() +
+            "' to a load/unload request is currently loading or unloading");
+  }
+
+  // At this point, it is committed to load/unload the models, so write the
+  // local copy into this object. The state of the models cannot be modified by
+  // another thread while is_locked = true. Thus, it is safe to release the
+  // lock while actively loading/unloading models, but the lock must be re-held
+  // when reading/updating any model info to avoid any corruption.
+  infos_.swap(infos);
+  global_map_.swap(global_map);
+  dependency_graph_.swap(dependency_graph);
 
   // The models are in 'deleted' either when they are asked to be unloaded or
   // they are not found / are duplicated across all model repositories.
   // In all cases, should unload them and remove from 'infos_' explicitly.
   for (const auto& name : (unload_dependents ? deleted_dependents : deleted)) {
     infos_.erase(name);
+    poll_mu_.unlock();
     model_life_cycle_->AsyncUnload(name);
+    poll_mu_.lock();
   }
 
   // Check global map for a set of same named models
   //
   // load / unload the models affected, and check the load status of
   // the requested models
-  const auto& load_status = LoadModelByDependency();
+  const auto& load_status = LoadModelByDependency(true /* release_mu */);
+  dependency_graph_.LockUnlockNodes(affected_models, false /* to_lock */);
   if (status.IsOk() && (type == ActionType::LOAD)) {
     std::string load_error_message = "";
 
@@ -1375,12 +1438,13 @@ ModelRepositoryManager::InitializeModelInfo(
   return Status::Success;
 }
 
-Status
+std::set<ModelIdentifier>
 ModelRepositoryManager::UpdateDependencyGraph(
+    DependencyGraph* dependency_graph, const ModelInfoMap& model_infos,
     const std::set<ModelIdentifier>& added,
     const std::set<ModelIdentifier>& deleted,
     const std::set<ModelIdentifier>& modified,
-    std::set<ModelIdentifier>* deleted_dependents)
+    std::set<ModelIdentifier>* deleted_dependents) const
 {
   // Update dependency graph in following steps:
   //   1. Apply all node changes.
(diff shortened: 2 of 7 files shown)


src/backend_model.cc
```diff
@@ -167,19 +167,25 @@ TritonModel::Create(
...
-  // path to point to the backend directory in case the backend
-  // library attempts to load additional shared libaries.
+  // Model initialization is optional... The TRITONBACKEND_Model object is this
+  // TritonModel object.
   if (backend->ModelInitFn() != nullptr) {
+    // We must set set shared library path to point to the backend directory in
+    // case the backend library attempts to load additional shared libaries.
+    // Currently, the set and reset function is effective only on Windows, so
+    // there is no need to set path on non-Windows.
+#ifdef _WIN32
```
dyastremsky, line 177: Why does only the Windows build need to lock and se the library directory now? Also, it might be good to figure out a way to log if ModelInit fails even if resetting the library fails. Not reporting the model initialization failure if the library resetting fails for some reasons seems like undesirable behavior.

kthui: >Why does only the Windows build need to lock and se the library directory now?

There is no effect for calling `slib->SetLibraryDirectory(backend->Directory())` and `slib->ResetLibraryDirectory()` on non-Windows platform, they are basically no-ops on non-Windows. Notice the `SharedLibrary::Acquire(&slib)` is required before calling "set" and "reset" library, so another thread cannot mess with the library setting while the model initialization is ongoing (on Windows). The mechanism is a global mutex that will remove all concurrency if it is held, so, it is important to not hold this global mutex if we are looking into any speedup with parallel model loading (at least on non-Windows).

> Also, it might be good to figure out a way to log if ModelInit fails even if resetting the library fails. Not reporting the model initialization failure if the library resetting fails for some reasons seems like undesirable behavior.

It is a good point, I think a block scope can be used to set/reset the shared library (on Windows), so the error could be printed immediate as it is returned.

src/model_repository_manager.cc
```diff
@@ -768,24 +822,63 @@ ModelRepositoryManager::LoadUnloadModels(
...
+  }
+
+  // At this point, it is committed to load/unload the models, so write the
+  // local copy into this object. The state of the models cannot be modified by
+  // another thread while they are locked. Thus, it is safe to release the
+  // mutex while actively loading/unloading models, but the mutex must be
+  // re-held when reading/updating any model info to avoid any corruption.
+  infos_.swap(infos);
+  global_map_.swap(global_map);
+  dependency_graph_.swap(dependency_graph);
```
GuanLuo, line 863: So from line 731 to 863, it depicts that "model info, global map, depedency graph" can be grouped as a structure, and there are three operations on the structure: acquire (make a copy), update, commit (set the copy as ground truth). Can you reorganize the code to make this clear?

src/model_repository_manager.cc
```diff
@@ -768,24 +822,63 @@ ModelRepositoryManager::LoadUnloadModels(
...
+
```
GuanLuo, line 864: You don't really need the lock afterward, which should save you from all the reverse locking. Idea is that by this point, the structure has been committed, which implies the "node" under change doesn't interfere with other (/ previous) ongoing model operation and there shouldn't be concurrent access on any data structure.

Then you only need to acquire the lock again once the whole operation is completed and unmarked the flag in dependency graph.

src/model_repository_manager.cc
```diff
@@ -768,24 +822,63 @@ ModelRepositoryManager::LoadUnloadModels(
...
+  // between the models to be loaded/unloaded by this thread with another
+  // thread actively loading/unloading models. The global map referenced by the
+  // dependency graph is managed by this object, which is the responsibility of
+  // this object to make a copy.
+  std::unordered_map<std::string, std::set<ModelIdentifier>> global_map(
+      global_map_);
+  DependencyGraph dependency_graph(dependency_graph_, &global_map);
+
+  // Update the dependency graph independently for the possibility of rollback.
+  auto affected_models = UpdateDependencyGraph(
```
GuanLuo, line 835: I think this function and maybe other doesn't really access any member variables after this change?

src/rate_limiter.cc
```diff
@@ -121,7 +121,11 @@ bool
 RateLimiter::PayloadSlotAvailable(const TritonModel* model)
```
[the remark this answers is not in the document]
kthui: Segfault was observed with parallel model loading when the mutex is not held when reading.

src/model_repository_manager.h
```diff
@@ -422,14 +455,24 @@ class ModelRepositoryManager {
   Status GetModelInfo(
       const ModelIdentifier& model_id, ModelInfo** model_info) const;
 
+  /// Get the model info for a named model, from the provided model infos.
+  /// \param model_infos The model infos where the model is to be lookup.
+  /// \param name The model name.
+  /// \param model_info Returns the model information.
+  /// \return OK if found, NOT_FOUND otherwise.
+  static Status GetModelInfo(
```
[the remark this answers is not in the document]
kthui: It does not, it belongs to the ModelInfo map held by the manager class. I think it would be better to refactor the map to its own class rather than depending on `std::unordered_map`.

src/model_repository_manager.h
```diff
@@ -117,29 +122,58 @@ class ModelRepositoryManager {
...
 
+    // Set the nodes to either lock or unlock state. If a node is at the same
+    // state as the new state to be set (i.e. to lock a node while it is
+    // already locked), then the identifier of the node is returned. Otherwise,
+    // nullptr is returned.
+    // The previous dependency graph is optional for checking if the lock state
+    // is already set when the node is not found on this dependency graph, and
+    // return the correct identifier, as the node could be deleted on this
+    // graph from the previous.
+    std::unique_ptr<ModelIdentifier> LockUnlockNodes(
```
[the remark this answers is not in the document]
kthui: Yes.

src/model_repository_manager.cc
```diff
@@ -476,20 +498,20 @@ ModelRepositoryManager::PollAndUpdateInternal(bool* all_models_polled)
...
-  UpdateDependencyGraph(added, deleted, modified);
+  UpdateDependencyGraph(&dependency_graph_, infos_, added, deleted, modified);
 
   for (const auto& name : deleted) {
     model_life_cycle_->AsyncUnload(name);
   }
 
   // model loading / unloading error will be printed but ignored
-  LoadModelByDependency();
+  LoadModelByDependency(false /* release_mu */);
```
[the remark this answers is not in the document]
kthui: In case the poll frequency is too fast (poll every 1 second and it takes 2 seconds to poll), there will be parallel poll if the mutex is released. This is to prevent this scenario.

src/model_repository_manager.cc
```diff
@@ -396,6 +417,7 @@ ModelRepositoryManager::Create(
     for (const auto& model_name : startup_models) {
       models[model_name];
     }
+    std::lock_guard<std::mutex> lock((*model_repository_manager)->poll_mu_);
```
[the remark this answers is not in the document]
kthui: It is because `LoadUnloadModels()` expects this mutex to be held before calling, this will unify with the rest of the callers.

src/model_repository_manager.cc
```diff
@@ -58,6 +58,27 @@ AddToSet(const std::set<T>& src, std::set<T>* dest)
...
+    if (op_)
+      mu_.lock();
+  }
+  RevLockGuard(const RevLockGuard&) = delete;
+  RevLockGuard& operator=(const RevLockGuard&) = delete;
+
+ private:
+  Mutex& mu_;
+  bool op_;
+};
```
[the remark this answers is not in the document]
kthui: It is because the lock scope span across different nested function calls. It needs to be held at the outer function and then release and re-held at an inner function.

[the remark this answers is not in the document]
GuanLuo: See another comment below, I think this nested locking can be avoided if you clean up the logic. And you should consider declare additional mutexes for finer control, the use of `poll_mtx` is everywhere and the chance of over-use is high.

src/backend_model.cc
kthui: Yes.

src/model_repository_manager.cc
kthui: Then I will bring back the writeback step, because the loaded info needs to be written back to the master records held by the manager after loading completes.

The steps will be: acquire (make a copy), update, commit (set the copy as ground truth), and writeback.

src/backend_model.cc
kthui: Updated.

src/model_repository_manager.cc
kthui: Yes.

src/model_repository_manager.h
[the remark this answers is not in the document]
kthui: Update commit: Separate LockUnlockNodes to LockNodes and UnlockNodes

src/model_repository_manager.h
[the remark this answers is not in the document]
kthui: Update commit: Group ModelInfoMap functions

src/model_repository_manager.cc
```diff
@@ -611,7 +650,17 @@ ModelRepositoryManager::LoadUnloadModel(
   }
 
   bool polled = true;
-  RETURN_IF_ERROR(LoadUnloadModels(models, type, unload_dependents, &polled));
+  while (true) {
```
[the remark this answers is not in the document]
kthui: To be updated.

src/model_repository_manager.cc
```diff
@@ -611,7 +650,17 @@ ModelRepositoryManager::LoadUnloadModel(
...
+    bool no_parallel_conflict = true;
+    RETURN_IF_ERROR(LoadUnloadModels(
+        models, type, unload_dependents, &polled, &no_parallel_conflict));
+    // Check for parallel load/unload conflict
+    if (no_parallel_conflict) {
+      break;
+    }
+    RevLockGuard<std::mutex> unlock(poll_mu_);
+    std::this_thread::yield();
```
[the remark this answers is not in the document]
kthui: To be updated.
It is possible to return a model specific condition variable instead of the model id when a conflict is detected, so the while loop can wait for signal on the condition variable.

src/model_repository_manager.cc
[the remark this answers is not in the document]
kthui: Yes.
Update commit: [Refactor parallel load logic](https://github.com/triton-inference-server/core/pull/165/commits/0fa4e296b613806d0bd7de77775f09e51d866d21)](https://github.com/triton-inference-server/core/pull/165/commits/0fa4e296b613806d0bd7de77775f09e51d866d21)

src/model_repository_manager.cc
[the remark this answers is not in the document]
kthui: The lock is no longer needed, after some logic refactoring.
Update commit: [Refactor parallel load logic](https://github.com/triton-inference-server/core/pull/165/commits/0fa4e296b613806d0bd7de77775f09e51d866d21)

src/model_repository_manager.cc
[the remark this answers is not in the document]
kthui: `LoadModelByDependency()` no longer require the mutex to be held, after some logic refactoring.
Update commit: [Refactor parallel load logic](https://github.com/triton-inference-server/core/pull/165/commits/0fa4e296b613806d0bd7de77775f09e51d866d21)

src/model_repository_manager.cc
kthui: They are somewhat refactored. Now, the grouping is only depicts within `LoadUnloadModels()`, and more separated in other functions.
Update commit: [Refactor parallel load logic](https://github.com/triton-inference-server/core/pull/165/commits/0fa4e296b613806d0bd7de77775f09e51d866d21)

src/model_repository_manager.cc
kthui: Update commit: [Refactor parallel load logic](https://github.com/triton-inference-server/core/pull/165/commits/0fa4e296b613806d0bd7de77775f09e51d866d21)

src/model_repository_manager.cc
kthui: Update commit: [Refactor parallel load logic](https://github.com/triton-inference-server/core/pull/165/commits/0fa4e296b613806d0bd7de77775f09e51d866d21)

kthui:
Rebased.

src/model_repository_manager.cc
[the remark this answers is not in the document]
kthui: Update commit: [Refactor parallel load logic](https://github.com/triton-inference-server/core/pull/165/commits/56ce83eb3d33e2253af0084d8ad4efa9bb800ab2)

src/model_repository_manager.cc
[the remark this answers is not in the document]
kthui: [Refactor parallel load logic](https://github.com/triton-inference-server/core/pull/165/commits/56ce83eb3d33e2253af0084d8ad4efa9bb800ab2)

src/model_repository_manager.h
```diff
@@ -72,12 +73,73 @@ class ModelRepositoryManager {
...
+    // Write updated model info back to this object after model load/unload.
+    void Writeback(
+        const ModelInfoMap& updated_model_info,
+        const std::set<ModelIdentifier>& affected_models);
+
+    std::unordered_map<ModelIdentifier, std::unique_ptr<ModelInfo>> map_;
+  };
+
+  // std::condition_variable wrapper with mutex included
+  class ConditionVariable {
```
GuanLuo, line 122: Give it a different name, it is easy to be confused with `std::condition_variable`, and the use of this class doesn't seem to be reflected by the name, i.e. it is not how I would imagine to use when I see a "condition variable".

src/model_repository_manager.h
```diff
@@ -72,12 +73,73 @@ class ModelRepositoryManager {
...
+  };
+
   /// A basic unit in dependency graph that records the models seen by the model
   /// repository manager.
   struct DependencyNode {
     DependencyNode(const ModelIdentifier& model_id)
         : status_(Status::Success), model_id_(model_id), checked_(false),
-          connected_(false)
+          connected_(false), is_locked_(false),
+          conflict_retry_cv_(new ConditionVariable())
```
GuanLuo, line 142: Each node holds one of this class at initialization?

src/model_repository_manager.cc
```diff
@@ -594,24 +570,29 @@ ModelRepositoryManager::LoadUnloadModel(
...
+    std::lock_guard<std::mutex> lock(mu_);  // guard global_map_
     const auto& git = global_map_.find(model_name);
     if (git != global_map_.end()) {
       to_be_unloaded_ids = git->second;
     }
   }
 
   bool polled = true;
-  RETURN_IF_ERROR(LoadUnloadModels(models, type, unload_dependents, &polled));
+  std::shared_ptr<ConditionVariable> conflict_retry_cv;
```
GuanLuo, line 588: I think you should call this class event listener? If I understand it correctly, `LoadUnloadModels()` set and return this object when the operation can not proceed (not due to error), then the caller will wait for change in state (event trigger, in this case node unlocked).

src/model_repository_manager.h
```diff
@@ -117,29 +179,108 @@ class ModelRepositoryManager {
...
+        std::shared_ptr<ConditionVariable>* conflict_retry_cv = nullptr);
+
+    // Set the nodes to unlock state. If a node is already unlocked, then the
+    // identifier of the node is returned. Otherwise, nullptr is returned.
+    std::unique_ptr<ModelIdentifier> UnlockNodes(
+        const std::set<ModelIdentifier>& nodes);
+
+    // Write updated graph back to this object after model load/unload.
+    // This will also notify load/unload conflict to retry.
+    void Writeback(
```
GuanLuo, line 264: Provide more document on what is `Writeback()` and make distinction between this function and `Assign()`

src/model_repository_manager.cc
```diff
@@ -671,155 +653,237 @@ ModelRepositoryManager::LoadUnloadModels(
...
+
+  // Unload 'deleted' models.
+  for (const auto& name : (unload_dependents ? deleted_dependents : deleted)) {
+    model_life_cycle_->AsyncUnload(name);
+  }
+  // Load/Unload the models affected
+  const auto& load_status = LoadModelByDependency(&dependency_graph, &infos);
+
+  {
+    std::lock_guard<std::mutex> lock(mu_);
```
GuanLuo, line 747: why not use `unique_lock` instead if you need to lock / unlock mutex multiple time in the same scope.

src/model_repository_manager.cc
```diff
@@ -671,155 +653,237 @@ ModelRepositoryManager::LoadUnloadModels(
...
-          }
-        }
+    std::lock_guard<std::mutex> lock(mu_);
+
+    // Prepare model infos related to file system accordingly, do not write yet
+    // until committed to load/unload the models.
+    // Also poll the added, deleted, modified, unmodified models.
+    infos = infos_;
+    std::set<ModelIdentifier> added, modified, unmodified;
+    RETURN_IF_ERROR(LoadUnloadModelsPoll(
```
GuanLuo, line 678: Give it different name, i.e. `PollModel()`, and should avoid keep passing the `type` because `UNLOAD` really doesn't need to perform the poll at all, just resovle the different type here:
```
switch (type) {
    case LOAD:
         PollModels(...);
    case UNLOAD:
         ListDeleted(...); // can be different name or just process the result in place.
}
```

src/model_repository_manager.h
kthui: Yes, because the load/unload conflict could be caused by any node (model), but at any time, there will only be one of this class per model, which is why it is held by a shared pointer to be shared between copies of the node.

src/model_repository_manager.cc
kthui: Update commit: Rename LoadUnloadModelsPoll to PollModels

src/model_repository_manager.h
kthui: Update comment: More comment on Writeback vs Assign

src/model_repository_manager.cc
kthui: I will call it event notifier, because event listener usually refer to an object that a callback function is provided, which the function will be called when an event happens.

kthui: Update commit: Rename ConditionVariable to EventNotifier

src/model_repository_manager.h
kthui: Update commit: Rename ConditionVariable to EventNotifier

src/model_repository_manager.cc
```diff
@@ -671,155 +653,237 @@ ModelRepositoryManager::LoadUnloadModels(
...
-  auto status = Status::Success;
-  *all_models_polled = true;
-  // Update ModelInfo related to file system accordingly
-  std::set<ModelIdentifier> added, deleted, modified, unmodified;
+  // They are here to hold information for loading/unloading models later, and
+  // also hold after loading/unloading information to be written back to this
+  // object later.
+  ModelInfoMap infos;
+  std::unordered_map<std::string, std::set<ModelIdentifier>> global_map;
+  DependencyGraph dependency_graph;
```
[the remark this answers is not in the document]
kthui: Update commit: Refactor LoadUnloadModels

src/model_repository_manager.cc
kthui: The ConditionVariable / EventNotifier / event listener intermediate class is no longer necessary with some logic simplification.
Update commit: Refactor conflict notify mechanism

src/model_repository_manager.h
kthui: The ConditionVariable / EventNotifier / event listener intermediate class is no longer necessary with some logic simplification.
Update commit: Refactor conflict notify mechanism

tanmayv25 (dismissed):

src/model_repository_manager.h
```diff
@@ -72,12 +73,73 @@ class ModelRepositoryManager {
...
+    std::string model_path_;
+    // Temporary location to hold agent model list before creating the model
+    // the ownership must transfer to ModelLifeCycle to ensure
+    // the agent model life cycle is handled properly.
+    std::shared_ptr<TritonRepoAgentModelList> agent_model_list_;
+    bool is_config_provided_;
+  };
+
+  // Map from model name to information about the model.
+  struct ModelInfoMap {
```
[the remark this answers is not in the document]
kthui: Will make it a proper class.
Update commit: Hide the map var inside model info map

tanmayv25 (approved):

kthui:
Rebased.

src/model_repository_manager.cc
```diff
@@ -1938,4 +1990,157 @@ ModelRepositoryManager::DependencyGraph::CircularDependencyCheck(
...
+    std::unordered_map<std::string, std::set<ModelIdentifier>>* global_map)
+{
+  DependencyGraph tmp(rhs, global_map);
+  Swap(tmp);
+}
+
+void
+ModelRepositoryManager::DependencyGraph::Swap(DependencyGraph& rhs)
+{
+  std::swap(global_map_ptr_, rhs.global_map_ptr_);
```
GuanLuo, line 2031: call swap on pointer?

src/model_repository_manager.h
```diff
@@ -350,10 +499,27 @@ class ModelRepositoryManager {
...
+  /// If 'no_parallel_conflict' is provided and there is a conflict, then the
+  /// function will block until the conflict is resolved and set the correct
+  /// value into the variable, and return with success status for retrying.
+  /// If 'no_parallel_conflict' is not provided and there is a conflict, then
+  /// the function will return immediately with an error status.
   Status LoadUnloadModels(
       const std::unordered_map<
           std::string, std::vector<const InferenceParameter*>>& models,
       const ActionType type, const bool unload_dependents,
+      bool* all_models_polled, bool* no_parallel_conflict = nullptr);
```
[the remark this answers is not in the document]
kthui: For two reasons:
1. When the `LoadUnloadModels()` is called by `ModelRepositoryManager::Create()`, parallel conflict cannot exist since only the create function is manipulating the models, so `Create()` does not have to provide the parameter and can safely leave it at `nullptr`.
2. If we later decide to add the functionality of returning an error immediately back to the user when there is a load/unload conflict, then the caller can set the `no_parallel_conflict` to `nullptr`. The function will return an error `Status` immediately when a conflict is seen.

src/model_repository_manager.h
```diff
@@ -72,12 +73,91 @@ class ModelRepositoryManager {
...
+    }
+    const std::unique_ptr<ModelInfo>& At(const ModelIdentifier& key) const
+    {
+      return map_.at(key);
+    }
+    Iterator Find(const ModelIdentifier& key) { return map_.find(key); }
+    ConstIterator Find(const ModelIdentifier& key) const
+    {
+      return map_.find(key);
+    }
```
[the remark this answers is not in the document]
kthui: I agree this seems unnecessary work at only manipulating the underlying map object, but the helper functions will also only be manipulating the map object, which the functions belong to nowhere but the map object itself. This is why I think grouping the helper functions and the map object together makes them more portable, but I think it is up to you. If you think the helper functions are clearer, I can switch over to the helper functions. 🙂

There will be 3 helper functions:
- `CopyInfoMap()`: copies the map
- `AssignInfoMap()`: destroys the LHS map and copy from RHS to LHS
- `WritebackInfoMap()`: writes post load/unload info back to the master map held by the `ModelRepositoryManager`

src/model_repository_manager.cc
```diff
@@ -1938,4 +1990,157 @@ ModelRepositoryManager::DependencyGraph::CircularDependencyCheck(
...
+  nodes_.swap(rhs.nodes_);
+  missing_nodes_.swap(rhs.missing_nodes_);
+}
+
+std::unique_ptr<ModelIdentifier>
```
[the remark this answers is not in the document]
kthui: To return a `nullptr` if there is no conflict. If there is a conflict, the `unique_ptr` contains the `ModelIdentifier` of the conflict model.

src/model_repository_manager.cc
kthui: This is correct, because the `global_map_` object is owned and managed by the `ModelRepositoryManager` and not by the `DependencyGraph`, so the `DependencyGraph` should only (and always) swap the pointer owned by the `DependencyGraph` (and not the content owned by the `ModelRepositoryManager`). This also avoid a surprising behavior from the `ModelRepositoryManager` standpoint that swapping the `DependencyGraph` causes the content of the `global_map_` object it manages to be updated.

src/model_repository_manager.h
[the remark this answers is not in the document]
kthui: @GuanLuo what do you think?

src/model_repository_manager.h
```diff
@@ -117,29 +197,111 @@ class ModelRepositoryManager {
...
+    // locked when it is not found on this dependency graph, as the node could
+    // be deleted on this graph from the previous, and correctly return the
+    // identifier.
+    // The conflict notifier can be provided optionally. If a node is already
+    // locked, it will be set to a notifier that the calling thread can wait on
+    // until notified to retry, to reduce the number of retries.
+    std::unique_ptr<ModelIdentifier> LockNodes(
+        const std::set<ModelIdentifier>& nodes,
+        const DependencyGraph& prev_dependency_graph,
+        std::shared_ptr<std::condition_variable>* retry_notify_cv = nullptr);
```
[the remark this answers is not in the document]
kthui: @GuanLuo what do you think?

src/rate_limiter.cc
```diff
@@ -302,6 +314,7 @@ RateLimiter::InitializePayloadQueues(const TritonModelInstance* instance)
   } else {
     max_queue_delay_microseconds = 0;
   }
+  PayloadQueue* payload_queue;
```
dyastremsky, line 317: You may want to initialize this immediately to a nullptr to avoid unexpected behavior (e.g. segfault) if the code changes later.

src/rate_limiter.cc
```diff
@@ -178,10 +186,14 @@ RateLimiter::DequeuePayload(
     std::shared_ptr<Payload>* payload)
 {
   payload->reset();
-  if (payload_queues_.find(instances[0]->Model()) == payload_queues_.end()) {
-    LOG_INFO << "Should not print this ";
+  PayloadQueue* payload_queue;
```
dyastremsky, line 189: You may want to initialize this immediately to a nullptr to avoid unexpected behavior (e.g. segfault) if the code changes later.

src/rate_limiter.cc
```diff
@@ -135,10 +139,14 @@ RateLimiter::EnqueuePayload(
     const TritonModel* model, std::shared_ptr<Payload> payload)
 {
   auto pinstance = payload->GetInstance();
-  if (payload_queues_.find(model) == payload_queues_.end()) {
-    LOG_INFO << "Should not print this ";
+  PayloadQueue* payload_queue;
```
dyastremsky, line 142: You may want to initialize this immediately to a nullptr to avoid unexpected behavior (e.g. segfault) if the code changes later.

src/rate_limiter.cc
```diff
@@ -121,7 +121,11 @@ bool
 RateLimiter::PayloadSlotAvailable(const TritonModel* model)
 {
   bool result;
-  PayloadQueue* payload_queue = payload_queues_[model].get();
+  PayloadQueue* payload_queue;
```
dyastremsky, line 124: You may want to initialize this immediately to a nullptr to avoid unexpected behavior (e.g. segfault) if the code changes later.

src/model_repository_manager.h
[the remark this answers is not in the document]
GuanLuo: 1. But there is also no harm for `Create()` to provide `no_parallel_conflict` variable, right? It can do nothing with it, or actually sanity check on indeed nothing needs to be waited on. In the current implementation, your `LoadUnloadModels()` is slightly more complicated because the implementation needs to be aware that `no_parallel_conflict` may not be provided and need to add conditionals for the two scenarios.
2. Then isn't it the same for the caller to check if `*no_parallel_conflict == false` and decide the follow-up action (retry / abort)?

src/model_repository_manager.h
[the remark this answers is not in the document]
GuanLuo: You can keep it this way.

src/rate_limiter.cc
kthui: Update commit: [Init rate limiter ptr to null once declared](https://github.com/triton-inference-server/core/pull/165/commits/15781d6d2a08c7053bb9fa0b0ec304bb3de7fb6e)

src/rate_limiter.cc
kthui: Update commit: [Init rate limiter ptr to null once declared](https://github.com/triton-inference-server/core/pull/165/commits/15781d6d2a08c7053bb9fa0b0ec304bb3de7fb6e)

src/rate_limiter.cc
kthui: Update commit: [Init rate limiter ptr to null once declared](https://github.com/triton-inference-server/core/pull/165/commits/15781d6d2a08c7053bb9fa0b0ec304bb3de7fb6e)

src/rate_limiter.cc
kthui: Update commit: [Init rate limiter ptr to null once declared](https://github.com/triton-inference-server/core/pull/165/commits/15781d6d2a08c7053bb9fa0b0ec304bb3de7fb6e)

src/model_repository_manager.cc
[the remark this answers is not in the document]
GuanLuo: You can just return empty ModelIdentifier for nothing if you really need it. Looking at how the caller use the returned value, it is only for logging purposes, i.e. the `retry_notify_cv` is already sufficient indicator of whether there is a conflict and caller can use it to wait on to something.

src/model_repository_manager.h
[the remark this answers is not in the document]
GuanLuo: Similar argument on the `nullptr` in `LoadUnloadModels`, and if not returning model identifier, the signature can even be
`std::shared_ptr<std::condition_variable>LockNodes(...)`, return value == nullptr if nothing to be waited on and contain an cv if there is conflict, and caller can decide if to wait for it or abort.

GuanLuo (approved):
so can be merged.

src/model_repository_manager.h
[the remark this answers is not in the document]
kthui: 👌

src/model_repository_manager.h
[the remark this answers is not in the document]
kthui: 1. Yes.
2. They are the same, except if the caller provides `no_parallel_conflict`, the `LoadUnloadModels()` will block until it is notified to retry (from the CV returned from `LockNodes()`) before returning. If `no_parallel_conflict == nullptr`, then the `LoadUnloadModels()` returns immediately (with an error `Status`) without waiting.

src/model_repository_manager.h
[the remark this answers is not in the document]
kthui: Yes, the only thing is the caller might want to know the `ModelIdentifier` for log printing purpose. Maybe it would be easier to simply return the `DependencyNode*` which contains everything and can be set to `nullptr` if no conflict.

tanmayv25 (approved):

GuanLuo (approved):

PR #166, opened 2023-02-14
Add namespace as model identifier to relax ensemble specification

GuanLuo:
This PR is to address https://github.com/triton-inference-server/server/issues/5150 where there are multiple ensemble repositories that happens to have composing models with the same name (see diagram below). Before the change, it will fail the model loading as the duplicated naming can't be resolved. So name-spacing is introduced to better differentiate the composing model: models with the same name will have different model identifier to be referred internal to the ensemble.

There are a few rules and assumptions when model name-spacing is enabled:
1. Different model repository will be assigned different namespaces.
2. Ensemble will use composing models that are in the same namespace, with 1. it implies locality, as we assume the models in the same repository should be coupled.
3. If composing model is not found in namespace, the ensemble will only use composing model in other namespace if it is not ambiguous by model name (i.e. there is only one such model across all registered model repositories).

Screenshot from 2023-01-04 14-28-57

### DIFF
--- a/src/ensemble_scheduler.cc
+++ b/src/ensemble_scheduler.cc
@@ -327,7 +327,7 @@ class EnsembleContext {
   std::unordered_map<std::string, TensorData> tensor_data_;
 
   // Handle to all models that may be used in the ensemble
-  std::unordered_map<std::string, VersionMap> handles_;
+  std::unordered_map<ModelIdentifier, VersionMap> handles_;
 
   // Request specific information that obtained from ensemble request and
   // should be applied to all internal requests
@@ -368,16 +368,16 @@ EnsembleContext::EnsembleContext(
   // they have the same lifetime as the ensemble request to avoid unloading
   // while the ensemble is executing.
   for (const auto& step_info : info_->steps_) {
-    auto it = handles_.find(step_info.model_name_);
+    auto it = handles_.find(step_info.model_id_);
     if (it == handles_.end()) {
-      it = handles_.emplace(std::make_pair(step_info.model_name_, VersionMap()))
+      it = handles_.emplace(std::make_pair(step_info.model_id_, VersionMap()))
                .first;
     }
     auto ver_it = it->second.find(step_info.model_version_);
     if (ver_it == it->second.end()) {
       std::shared_ptr<Model> model = nullptr;
       ensemble_status_ = is_->GetModel(
-          step_info.model_name_, step_info.model_version_, &model);
+          step_info.model_id_, step_info.model_version_, &model);
       if (!ensemble_status_.IsOk()) {
         break;
       }
@@ -860,7 +860,7 @@ EnsembleContext::InitStep(
     std::unique_ptr<Step>* step)
 {
   const auto& istep = info_->steps_[step_idx];
-  auto& version_map = handles_[istep.model_name_];
+  auto& version_map = handles_[istep.model_id_];
   auto& model = version_map[istep.model_version_];
 
   const bool allow_batching = (model->Config().max_batch_size() > 0);
@@ -912,7 +912,7 @@ EnsembleContext::InitStep(
                             (flags != tensor.flags_))) {
         LOG_ERROR << irequest->LogRequest()
                   << "Different set of response parameters are set for '"
-                  << istep.model_name_ << "'. Parameter correlation ID "
+                  << istep.model_id_ << "'. Parameter correlation ID "
                   << correlation_id << ", flags " << flags << " is used.";
         continue;
       }
@@ -1347,7 +1347,7 @@ EnsembleScheduler::EnsembleScheduler(
 
   for (const auto& element : config.ensemble_scheduling().step()) {
     size_t step_idx = info_->steps_.size();
-    info_->steps_.emplace_back(element.model_name(), element.model_version());
+    info_->steps_.emplace_back(ModelIdentifier(element.model_namespace(), element.model_name()), element.model_version());
     for (const auto& pair : element.input_map()) {
       auto it = info_->tensor_to_step_.find(pair.second);
       if (it == info_->tensor_to_step_.end()) {
--- a/src/ensemble_scheduler.cc
+++ b/src/ensemble_scheduler.cc
@@ -1,4 +1,4 @@
-// Copyright 2019-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2019-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
@@ -376,8 +376,8 @@ EnsembleContext::EnsembleContext(
     auto ver_it = it->second.find(step_info.model_version_);
     if (ver_it == it->second.end()) {
       std::shared_ptr<Model> model = nullptr;
-      ensemble_status_ = is_->GetModel(
-          step_info.model_id_, step_info.model_version_, &model);
+      ensemble_status_ =
+          is_->GetModel(step_info.model_id_, step_info.model_version_, &model);
       if (!ensemble_status_.IsOk()) {
         break;
       }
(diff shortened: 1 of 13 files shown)


src/model.h
```diff
@@ -31,6 +31,73 @@
...
+  }
+
+  bool operator==(const ModelIdentifier& rhs) const
+  {
+    return ((namespace_ == rhs.namespace_) && (name_ == rhs.name_));
+  }
+
+  bool operator!=(const ModelIdentifier& rhs) const
+  {
+    return (!(namespace_ != rhs.namespace_) || !(name_ != rhs.name_));
```
kthui, line 55: I am wondering if you mean `(namespace_ != rhs.namespace_) || (name_ != rhs.name_)`? So, `(lhs == rhs) == !(lhs != rhs)`

src/model_repository_manager.cc
```diff
@@ -839,7 +940,17 @@ ModelRepositoryManager::GetModel(
     const std::string& model_name, const int64_t model_version,
     std::shared_ptr<Model>* model)
 {
-  Status status = model_life_cycle_->GetModel(model_name, model_version, model);
+  ModelIdentifier model_id("", model_name);
+  RETURN_IF_ERROR(find_identifier_fn_(model_name, &model_id));
+  return GetModel(model_id, model_version, model);
```
dyastremsky, line 945: Why is the behavior different for here versus the other GetModel below this? That one resets the model if there is an error, this one does not.

src/model_repository_manager.cc
```diff
@@ -848,24 +959,54 @@ ModelRepositoryManager::GetModel(
...
+  {
+  case 0:
+    return Status(Status::Code::NOT_FOUND, "Identifier of model '" + model_name + "' is not found in global map");
+    break;
+  case 1: {
+    *model_id = *git->second.begin();
+    break;
+  }
+  default:
+    return Status(Status::Code::INVALID_ARG, "There are multiple identifiers of model '" + model_name + "' in global map, model namespace must be provided to resolve ambiguity.");
```
dyastremsky, line 979: Optional, but it could be good to include the number of identifiers (git->second.size()) in the error message.

src/model_repository_manager.cc
```diff
@@ -1623,4 +1607,284 @@ ModelRepositoryManager::CheckNode(
...
+  }
+}
+
+std::pair<std::set<ModelIdentifier>, std::set<ModelIdentifier>>
+ModelRepositoryManager::DependencyGraph::RemoveNode(const ModelIdentifier& model_id)
+{
+  std::set<ModelIdentifier> upstreams, downstreams;
+  auto it = nodes_.find(model_id);
+  // no-op if not found: "node" has been removed
+  if (it != nodes_.end()) {
```
dyastremsky, line 1797: May be able to simplify code by early-exiting if not found. To avoid indenting if block below. Totally up to you, feel free to hit resolve if you'd rather not do this.

src/model.h
GuanLuo: Yes, missed the cleanup after expanding "not" operation

src/model_repository_manager.cc
kthui: I think the former one gets model by `model_name`, which could collide between different model namespaces. In the event of a collision, returning an error seems appropriate. The latter one gets model by `model_id`, which uniquely identifies the model. If the model is not found (or not ready), then the `shared_ptr` holding the `Model` gets reset. This should prevent the caller from interacting with the `Model` object, but I am not sure if this is @GuanLuo's intent for the reset.

src/model_repository_manager.cc
```diff
@@ -612,57 +677,71 @@ ModelRepositoryManager::LoadUnloadModels(
   auto status = Status::Success;
   *all_models_polled = true;
   // Update ModelInfo related to file system accordingly
-  std::set<std::string> added, deleted, modified, unmodified;
+  std::set<ModelIdentifier> added, deleted, modified, unmodified;
   {
     if (type == ActionType::UNLOAD) {
       for (const auto& model : models) {
-        deleted.insert(model.first);
+        for (const auto& info : infos_) {
```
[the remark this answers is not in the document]
GuanLuo: Yes, I think it can be check from the global map

src/model_repository_manager.h
```diff
@@ -316,33 +404,62 @@ class ModelRepositoryManager {
...
+  // malformed => re-load failure. Marking info "dirty" in this case to ensure
+  // the model to be polled again even though the storage model hasn't been
+  // changed, for operation idempotence.
+  // (https://github.com/triton-inference-server/server/issues/3802)
+  ModelInfoMap infos_;
+  std::set<std::string> repository_paths_;
   // Mappings from (overridden) model names to a pair of their repository and
   // absolute path
+  // [FIXME] key should be updated to contain namespace as well, need to work with
+  // enable namespace. Need to revisit with repository side of things
```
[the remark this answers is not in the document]
GuanLuo: Basically when a mapping (i.e. `name_a -> name_b`) is added during repo registration `repo_1`, if we track it with model identifier instead of just the name, we can have repeated mapped name. For example, on top of `name_a -> name_b`, you register another repo (`repo_2`) that maps `name_c -> name_b`, currently it doesn't work because `name_b` is already a key, using model identifier will allow this because there are two keys `repo_1::name_b`, `repo_2::name_b`

src/model_repository_manager.cc
GuanLuo: This function calls the `GetModel()` overload below, so it adds additional global lookup on top of the other overload. The `reset()` below is sanity check that we don't care anything over `model_life_cycle_->GetModel()` in case of error (i.e. accidentally set by lifecycle).

src/model.h
```diff
@@ -31,6 +31,73 @@
 #include "scheduler.h"
 #include "status.h"
 
+namespace triton { namespace core {
+struct ModelIdentifier {
+  ModelIdentifier(const std::string& model_namespace, const std::string& name)
+   : namespace_(model_namespace), name_(name) {}
+
+  bool operator<(const ModelIdentifier& rhs) const
```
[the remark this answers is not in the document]
kthui: I think the `ModelIdentifier` needs to behave like a `std::string`, to be used in place where "model_name" was used.

src/model_repository_manager.h
```diff
@@ -73,20 +74,106 @@ class ModelRepositoryManager {
...
+
+    // Passing reference of 'global_map'.
+    // There is coupling between the dependency graph and global map: global map
+    // will be used to resolve dependency (node connectivity), and global map
+    // will be updated to reflect node changes.
+    DependencyGraph(std::unordered_map<std::string, std::set<ModelIdentifier>>& global_map)
+     : global_map_ref_(global_map) {}
+
+    std::unordered_map<ModelIdentifier, std::unique_ptr<DependencyNode>>*
+    MutableNodes() {return &nodes_; }
```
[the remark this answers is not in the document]
GuanLuo: Thanks for catching that, it turns out that the pre-commit hook is not working and I fixed that. Let me know if the format still seems off in here and other places

kthui (dismissed):

src/ensemble_scheduler.cc
```diff
@@ -327,7 +327,7 @@ class EnsembleContext {
   std::unordered_map<std::string, TensorData> tensor_data_;
```
dyastremsky, line 327: Update copyright year in this file.

src/ensemble_utils.cc
```diff
@@ -42,9 +42,9 @@ namespace {
 /// of the ensemble tensor and which model they are inferred from.
```
dyastremsky, line 42: Update copyright year here. Please check the other files too.

tanmayv25 (dismissed):

dyastremsky (approved):

kthui (approved):

rmccorm4 (approved):

Tabrizian (approved):

tanmayv25 (approved):

PR #170, opened 2023-02-23
Fix certain race conditions encountered in deferred requests

tanmayv25:
There was a race condition where the payload can be executed by a backend thread, before a callback was attached to the payload for releasing the instance resources. This will prevent allocation of resources to other instances from making a progress. 
The case seems like to be limited to the scenario of using rate limiter with constrained resources.

### DIFF
--- a/src/rate_limiter.cc
+++ b/src/rate_limiter.cc
@@ -1,4 +1,4 @@
-// Copyright 2020-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2020-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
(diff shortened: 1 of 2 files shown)


GuanLuo (dismissed):
Add some documentation on function declaration.

rmccorm4:
Any chance there was a similar race in Sequence Batcher calling these related functions that might now be fixed? https://github.com/triton-inference-server/core/blob/17e764190317635144787381ba182e030f5ebb1d/src/sequence_batch_scheduler.cc#L1493-L1498

tanmayv25:
> Any chance there was a similar race in Sequence Batcher calling these related functions that might now be fixed? 

I don't think these fixes affect any pipelines that has rate limiter disabled.

dyastremsky (approved):

PR #172, opened 2023-02-27
Add the ability for backends to add entries to the trace

Tabrizian:
The API is tested by being used in the frontends.

### DIFF
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -1525,7 +1525,7 @@ TRITONBACKEND_BackendAttributeSetParallelModelInstanceLoading(
 ///
 /// \param batcher User-defined placeholder for backend to store and
 /// retrieve information about the batching strategy for this
-/// model.RITONBACKEND_ISPEC return a TRITONSERVER_Error indicating success or
+/// model. TRITONBACKEND_ISPEC returns a TRITONSERVER_Error indicating success or
 /// failure. \param model The backend model for which Triton is forming a batch.
 /// \return a TRITONSERVER_Error indicating success or failure.
 TRITONBACKEND_ISPEC TRITONSERVER_Error* TRITONBACKEND_ModelBatcherInitialize(
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -786,6 +786,16 @@ TRITONSERVER_DECLSPEC struct TRITONSERVER_Error* TRITONSERVER_InferenceTraceNew(
     TRITONSERVER_InferenceTraceActivityFn_t activity_fn,
     TRITONSERVER_InferenceTraceReleaseFn_t release_fn, void* trace_userp);
 
+/// Get the name associated with a trace. The caller does
+/// not own the returned string and must not modify or delete it. The
+/// lifetime of the returned string extends only as long as 'trace'.
+///
+/// \param trace The trace.
+/// \param trace_name Returns the name associated with the trace.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_InferenceTraceName(
+    TRITONSERVER_InferenceTrace* trace, const char** trace_name);
+
 /// Create a new inference trace object. The caller takes ownership of
 /// the TRITONSERVER_InferenceTrace object and must call
 /// TRITONSERVER_InferenceTraceDelete to release the object.
@@ -819,14 +829,16 @@ TRITONSERVER_InferenceTraceTensorNew(
     TRITONSERVER_InferenceTraceTensorActivityFn_t tensor_activity_fn,
     TRITONSERVER_InferenceTraceReleaseFn_t release_fn, void* trace_userp);
 
-/// Report a trace activity.
+/// Report a trace activity. All the traces reported using this API will be
+/// using TRITONSERVER_TRACE_CUSTOM_ACTIVITY type.
 ///
 /// \param trace The trace object.
 /// \param timestamp The timestamp associated with the trace activity.
+/// \param name The trace activity name.
 /// \return a TRITONSERVER_Error indicating success or failure.
 TRITONSERVER_DECLSPEC TRITONSERVER_Error*
 TRITONSERVER_InferenceTraceReportActivity(
-    TRITONSERVER_InferenceTrace* trace, uint64_t timestamp);
+    TRITONSERVER_InferenceTrace* trace, uint64_t timestamp, const char* name);
 
 /// Delete a trace object.
 ///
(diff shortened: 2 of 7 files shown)


include/triton/core/tritonserver.h
```diff
@@ -847,6 +857,34 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_InferenceTraceParentId(
 TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_InferenceTraceModelName(
     TRITONSERVER_InferenceTrace* trace, const char** model_name);
 
+/// Set the name associated with a trace.
```
rmccorm4, line 860:
```suggestion
/// Set the name of the model associated with the trace.
```

include/triton/core/tritonserver.h
```diff
@@ -847,6 +857,34 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_InferenceTraceParentId(
...
+///
+/// \param trace The trace.
+/// \param name The name associated with the trace.
```
rmccorm4, line 863:
```suggestion
/// \param model_name The name of the model associated with the trace.
```

include/triton/core/tritonserver.h
```diff
@@ -847,6 +857,34 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_InferenceTraceParentId(
...
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error*
+TRITONSERVER_InferenceTraceSetModelName(
+    TRITONSERVER_InferenceTrace* trace, const char* name);
```
rmccorm4, line 867: `model_name` to match with the Getter API
```suggestion
    TRITONSERVER_InferenceTrace* trace, const char* model_name);
```

include/triton/core/tritonserver.h
```diff
@@ -847,6 +857,34 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_InferenceTraceParentId(
...
+
+/// Get the name associated with a trace. The caller does
+/// not own the returned string and must not modify or delete it. The
+/// lifetime of the returned string extends only as long as 'trace'.
+///
+/// \param trace The trace.
+/// \param name Returns the name associated with the trace.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_InferenceTraceName(
+    TRITONSERVER_InferenceTrace* trace, const char** name);
```
rmccorm4, line 877: `trace_name` to match Setter API
```suggestion
/// \param trace_name Returns the name associated with the trace.
/// \return a TRITONSERVER_Error indicating success or failure.
TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_InferenceTraceName(
    TRITONSERVER_InferenceTrace* trace, const char** trace_name);
```

include/triton/core/tritonserver.h
```diff
@@ -847,6 +857,34 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_InferenceTraceParentId(
...
+
+/// Set the name associated with a trace.
+///
+/// \param trace The trace.
+/// \param trace_name The anme associated with a trace.
```
rmccorm4, line 882:
```suggestion
/// \param trace_name The name associated with a trace.
```

include/triton/core/tritonbackend.h
```diff
@@ -530,6 +530,15 @@ TRITONBACKEND_RequestOutputBufferProperties(
 TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_RequestRelease(
     TRITONBACKEND_Request* request, uint32_t release_flags);
 
+/// Get the trace associted with a request. The returned trace is owned by the
```
rmccorm4, line 533:
```suggestion
/// Get the trace associated with a request. The returned trace is owned by the
```

include/triton/core/tritonserver.h
Tabrizian: This API was added by mistake. I have removed it.

include/triton/core/tritonbackend.h
```diff
@@ -530,6 +530,15 @@ TRITONBACKEND_RequestOutputBufferProperties(
 TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_RequestRelease(
     TRITONBACKEND_Request* request, uint32_t release_flags);
 
+/// Get the trace associated with a request. The returned trace is owned by the
+/// request, not the caller, and so should not be modified or freed.
```
GuanLuo, line 534: Mention `nullptr` may be returned if request is not being traced

include/triton/core/tritonserver.h
```diff
@@ -810,6 +811,15 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_InferenceTraceTensorNew(
...
     TRITONSERVER_InferenceTraceReleaseFn_t release_fn, void* trace_userp);
 
+/// Report a trace activity.
+///
+/// \param trace The trace object.
+/// \param timestamp The timestamp associated with the trace activity.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error*
+TRITONSERVER_InferenceTraceReportActivity(
+    TRITONSERVER_InferenceTrace* trace, uint64_t timestamp);
```
GuanLuo, line 821: Don't need to specify what kind of activity?

include/triton/core/tritonserver.h
```diff
@@ -847,6 +857,24 @@ TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_InferenceTraceParentId(
...
+
+/// Get the name associated with a trace. The caller does
+/// not own the returned string and must not modify or delete it. The
+/// lifetime of the returned string extends only as long as 'trace'.
+///
+/// \param trace The trace.
+/// \param trace_name Returns the name associated with the trace.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_InferenceTraceName(
+    TRITONSERVER_InferenceTrace* trace, const char** trace_name);
```
GuanLuo, line 876: Looking at the usage, this is actually the name of the custom activity? I think you should let `TRITONSERVER_InferenceTraceReportActivity` accept the name as another parameter so the backend doesn't have to call the setter before calling the report function (implicitly set by Triton). Also make it clear that if the tracer receive custom activity, should call the getter function(s) for additional information regarding the activity

Tabrizian: > Looking at the usage, this is actually the name of the custom activity?

That's correct.

> I think you should let TRITONSERVER_InferenceTraceReportActivity accept the name as another parameter so the backend doesn't have to call the setter before calling the report function (implicitly set by Triton).

I think the way it is currently implemented would allow us to add more fields to the trace without having to modify the signature of `TRITONSERVER_InferenceTraceReportActivity`. Let me know if this makes sense to you.

> Also make it clear that if the tracer receive custom activity, should call the getter function(s) for additional information regarding the activity

👍

include/triton/core/tritonserver.h
Tabrizian: I didn't want to allow the C-API to use any activity other than the CUSTOM_ACTIVTY to avoid overriding the values provided by the core tracing.

include/triton/core/tritonserver.h
Tabrizian: > I think you should let TRITONSERVER_InferenceTraceReportActivity accept the name as another parameter so the backend doesn't have to call the setter before calling the report function (implicitly set by Triton).


I actually see the problem now. How about adding a `TRITONSERVER_InferenceTrace* trace_attributes` that currently only has a name?

GuanLuo: I don't know if there is a need to add another struct for this, when capturing a timestamp, what else do you need to know beside the timestamp itself and the activity it indicates? Also note that those should be properties that are activity specific (i.e. may be different on different capture of the timestamp), if they are properties of the entire trace, they should not be tied to this function (i.e. request ID / parent ID is always the same for the entire trace).

If you perceive more capture specific attribute may be added into the future, I think you should make it generic (string key value pair to be attached to the capture) so all parties are not tied to the specific getter/setter exposed in Triton API. For example, Triton user is likely going to define their own attributes for tracking backend model execution, which is going to be not useful if they have to go through the API change so that the tracer can extract the information. On the other hand, the tracer implemented can simply dump all attached key-value if any, without worrying to modify the implementation whenever a new attribute is added.

rmccorm4:
What's the state of this PR @Tabrizian ? Do we intend to finalize it at some point? I think a new backend API like this could enable a cleaner way to report  nested spans (create traces) for OpenTelemetry tracing closer to the site, rather than the current behavior of reporting everything all at the end in the TRITONBACKEND_ReportStatistics callback. CC @oandreeva-nv

Tabrizian:
@rmccorm4 Sorry for the delay. This must have fallen through the cracks. Yeah, I think we probably want to have this API at some point for creating custom traces. Right now I think this is just blocked on testing.

PR #173, opened 2023-03-02
Remove C++17 features/syntax

rmccorm4:
Fixes issues with `google-cloud-cpp` / `abseil` libraries with weird string / ABI handling. Ultimately it appears that some `std::string` internals were failing in GCS/Curl libraries when core was compiled with C++17, resulting in `Segmentation fault` or `bad_alloc`, etc.

It may be possible to update all dependent packages to be built with C++17 as well, but I think there is a larger risk/time sink in doing so as other issues will likely pop up and ABI compatibility is tricky. 

## Future Revisiting

This may be worth revisiting in the future when the default `-std` is >=C++17.

The default gcc version in Ububtu 20.04 is gcc 9.X, which has a default C++ `-std=c++14`:
```
$ gcc -v
gcc version 9.4.0 (Ubuntu 9.4.0-1ubuntu1~20.04.1) 
$ gcc -dM -E -x c++  /dev/null | grep -F __cplusplus
#define __cplusplus 201402L
```

The default gcc version in Ubuntu 22.04 is 11.X, which has a default C++ `-std=c++17`, so maybe it can be revisited when we bump to Ubuntu 22.04.
```
$ gcc -v
gcc version 11.3.0 (Ubuntu 11.3.0-1ubuntu1~22.04) 
$ gcc -dM -E -x c++  /dev/null | grep -F __cplusplus
#define __cplusplus 201703L
```

Some things/references I looked into:
- https://github.com/googleapis/google-cloud-cpp/blob/main/doc/packaging.md
- https://github.com/googleapis/google-cloud-cpp/blob/main/doc/packaging.md#abseil-2
- https://stackoverflow.com/questions/34571583/understanding-gcc-5s-glibcxx-use-cxx11-abi-or-the-new-abi
- Updating GCS from 1.42.0 to latest 2.7.0
- https://github.com/abseil/abseil-cpp/issues/1116

### DIFF
--- a/src/cache_entry.h
+++ b/src/cache_entry.h
@@ -35,6 +35,10 @@
 
 namespace triton { namespace core {
 
+// Part of std in C++17
+// enum class Byte : unsigned char {};
+using Byte = unsigned char;
+
 struct CacheOutput {
   // Inference Response output name
   std::string name_ = "";
@@ -62,7 +66,7 @@ class CacheEntry {
   const std::vector<Buffer>& Buffers();
   std::vector<Buffer>& MutableBuffers();
   size_t BufferCount();
-  void AddBuffer(boost::span<std::byte> buffer);
+  void AddBuffer(boost::span<Byte> buffer);
   void AddBuffer(void* base, size_t byte_size);
 
   /* Insert helpers */
@@ -81,7 +85,7 @@ class CacheEntry {
   // and sets entry buffer sizes
   Status SetBufferSizes(boost::span<InferenceResponse*> responses);
   // Directly sets entry buffer sizes from provided buffers
-  Status SetBufferSizes(std::vector<boost::span<std::byte>> buffers);
+  Status SetBufferSizes(std::vector<boost::span<Byte>> buffers);
 
   /* Lookup helpers */
   Status DeserializeBuffers(boost::span<InferenceResponse*> responses);
@@ -95,7 +99,7 @@ class CacheEntry {
   // Insert helpers
   Status SerializeResponse(InferenceResponse* response, Buffer& buffer);
   Status SerializeResponseOutput(
-      const InferenceResponse::Output& output, std::byte* buffer,
+      const InferenceResponse::Output& output, Byte* buffer,
       size_t* output_size);
   Status SetBufferSize(InferenceResponse* response);
   // Calculates total byte size required to serialize response output and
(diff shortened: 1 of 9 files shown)


src/cache_entry.h
```diff
@@ -35,6 +35,9 @@
 
 namespace triton { namespace core {
 
+// Part of std in C++17
```
dyastremsky, line 38: May want to clarify this comment. Does this mean something along the lines of TODO: Change to std::byte when upgrade to C++17?

GuanLuo:
So which C++ standard is the depending libraries to be compiled with? I don't see we explicitly control that when building third-party libraries and I wonder why the ABI issue wasn't shown before.

rmccorm4:
> So which C++ standard is the depending libraries to be compiled with? I don't see we explicitly control that when building third-party libraries and I wonder why the ABI issue wasn't shown before.

@GuanLuo It's not clear to me how compiler features or minimum standards get propogated through `ExternalProject_add`. From what I can tell, each of the external projects related to this issue (gcs, absl, etc.) that we pull in is compiling with whatever its own defaults are as it didn't look like we were setting any CXX standards on them.

I spent quite some time looking into it but struggled to find anything definitive. Supposedly `target_compile_features` is the [clean way to handle this](https://hsf-training.github.io/hsf-training-cmake-webpage/aio/index.html#example-2-c-standard) type of thing, but I wasn't sure how to use it on the external projects added in third_party cmake files. After several failed attempts taking different routes, I just reverted to what we had that was working.

Trying to set `CMAKE_CXX_STANDARD 17` and pass it everywhere seemed like it *may work* for external projects, but it didn't seem like the clean/idiomatic approach. Even still, I tried it and ran into other compilation issues in GCS related to unused variables that didn't go away after removing `-Wall -Werror` in our CMake files, and decided to not spend too much more time on it.

src/cache_entry.h
rmccorm4: done

src/cache_entry.h
```diff
@@ -35,6 +35,9 @@
 
 namespace triton { namespace core {
 
+// Note: This can be replaced with std::byte starting in c++17
+using Byte = unsigned char;
```
GuanLuo, line 39: `uint8_t`? move to `common.h` maybe

rmccorm4: Kept as `unsigned char` from offline discussion, moved to `constants.h`

GuanLuo (approved):

PR #174, opened 2023-03-10
Add request parameters API

Tabrizian:
Look and feel of the request parameters API. It is implemented based on the inference response parameters.

### DIFF
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -429,12 +429,12 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_RequestParameterCount(
 /// \param index The index of the parameter. Must be 0 <= index <
 /// count, where count is the value returned by
 /// TRITONBACKEND_RequestParameterCount.
-/// \param name Returns the name of the parameter.
+/// \param key Returns the key of the parameter.
 /// \param type Returns the type of the parameter.
 /// \param vvalue Returns a pointer to the parameter value.
 /// \return a TRITONSERVER_Error indicating success or failure.
 TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_RequestParameter(
-    TRITONBACKEND_Request* request, const uint32_t index, const char** name,
+    TRITONBACKEND_Request* request, const uint32_t index, const char** key,
     TRITONSERVER_ParameterType* type, const void** vvalue);
 
 /// Get the number of input tensors specified in the request.
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -27,6 +27,7 @@
 
 #include <stddef.h>
 #include <stdint.h>
+
 #include "triton/core/tritonserver.h"
 
 #ifdef __cplusplus
@@ -93,7 +94,7 @@ struct TRITONBACKEND_Batcher;
 ///   }
 ///
 #define TRITONBACKEND_API_VERSION_MAJOR 1
-#define TRITONBACKEND_API_VERSION_MINOR 11
+#define TRITONBACKEND_API_VERSION_MINOR 12
 
 /// Get the TRITONBACKEND API version supported by Triton. This value
 /// can be compared against the TRITONBACKEND_API_VERSION_MAJOR and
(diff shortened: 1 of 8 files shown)


GuanLuo (dismissed):

dyastremsky (dismissed):

tanmayv25 (approved):

include/triton/core/tritonbackend.h
```diff
@@ -406,6 +407,36 @@ TRITONBACKEND_RequestCorrelationIdString(
...
+/// count, where count is the value returned by
+/// TRITONBACKEND_RequestParameterCount.
+/// \param name Returns the name of the parameter.
+/// \param type Returns the type of the parameter.
+/// \param vvalue Returns a pointer to the parameter value.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_RequestParameter(
+    TRITONBACKEND_Request* request, const uint32_t index, const char** name,
+    TRITONSERVER_ParameterType* type, const void** vvalue);
+
```
nv-kmcgill53, line 439: As this is on the hot path of inferencing, I'm wondering if it's better to have a function which returns a structured list back to the backend and allow the backend/user to iterate through it themselves? You point out in your description that the request parameters are not guaranteed to be in the same order every request which means the user will have to iterate through the entire list to find paramters they want to use/ignore. If we gave them a structured list, they may be able to search through their list faster than we can.

```
// Current API usage
TRITONSERVER_Error* TRITONBACKEND_ModelInstanceExecute(
    TRITONBACKEND_ModelInstance* instance, TRITONBACKEND_Request** requests,
    const uint32_t request_count) {

    // The user is constrained to use this paradigm to get the request parameters 
    // even if they know they can do something different.
    size_t count;
    TRITONBACKEND_RequestParameterCount(instance, &count);
    for (int i = 0; i < count; ++i) {
        char *name;
        TRITONSERVER_ParameterType type;
        uint8_t *value;
        TRITONBACKEND_RequestParameter(request, i, &name, &type, &value); 
        switch (type) {
            case STRING:
                //convert *value to string obj
            case INTEGER:
                //convert *value to string obj
            case BOOLEAN:
                //convert *value to string obj
            default:
                // unsupported type
        }
    }
    
    //... do more work
}

// Proposed API usage
TRITONSERVER_Error* TRITONBACKEND_ModelInstanceExecute(
    TRITONBACKEND_ModelInstance* instance, TRITONBACKEND_Request** requests,
    const uint32_t request_count) {

    // Depending on the structure we give back, the user can implement their
    // own logic to iterate through the parameters.
    char *structured_parameters; // or we define a new structure for them to use.
    TRITONBACKEND_RequestParameters(request, &structured_parameters);
    // parse values
    // for each value
        // key
        // value
        // convert value based on the key

    //... do more work
}
```

nv-kmcgill53 (changes requested):
How does this API interact with batching requests? Do the request parameters get merged or mapped in some way?

include/triton/core/tritonbackend.h
Tabrizian: I think the problem is that in order to provide the keys separately we would have to copy the keys from `dequeue` in a separate list which would add a data copy.

Tabrizian:
> How does this API interact with batching requests? Do the request parameters get merged or mapped in some way?

No, the parameters need to be retrieved on a per-request basis.

include/triton/core/tritonbackend.h
GuanLuo: I don't think the backend will access the parameters by index, I would imagine the backend already knows what are the parameters that it cares about and thus a named look up will be the most convenient.

nv-kmcgill53: tldr: I think there is only utility in my suggestion if we are able to provide the request parameters as a structured blob back to the user.

This might be a bit long but I think it's useful to go through the thought experiment. We could provide pointers to the request parameters in a linked list so that there doesn't have to be a copy.
```
struct RequestParameterNode {
    char *key;
    char *value;
    TRITONSERVER_ParameterType *type
    RequestParameterNode *next;
}
```

The side effect of this is the user has to wait for us to build the whole linked list every time rather than being able to iterate through the list with the proposed set of APIs and (potentially) randomly early out-ing due to some parsing criteria.  Here the utility of having a simpler API is lost as the user will still have to iterate through the linked list in order to determine their actions.

So the value for giving back all request parameters to the user is in an iteratable blob of some kind (think a concatenated string of text)  which the user can search for themselves to find the information they need. 

>I think the problem is that in order to provide the keys separately we would have to copy the keys from dequeue in a separate list which would add a data copy.

You are right here in that there is a data copy. I think the only way of knowing for sure of the performance hit would be to bench mark the two ideas as we could alter the API to take a pre-allocated buffer which the backend manages.

tanmayv25 (approved):

tanmayv25:
@Tabrizian  What about Guan's suggestion to retrieve parameter by name? If we want to support both the cases, we can split the API to retrieve the parameter keys by index, and other for retrieving parameter value by provided name. If the name is already known by backend, they can directly access latter API.

include/triton/core/tritonbackend.h
Tabrizian: >I don't think the backend will access the parameters by index, I would imagine the backend already knows what are the parameters that it cares about and thus a named look up will be the most convenient.

@GuanLuo / @tanmayv25  Our current use-case (python backend) wants to retrieve all the parameters but I can imagine that some backends may want to access the parameters by name. I think we can add this (similar to the inputs that have both index-based and name-based retrieval) but this would be separate ticket.

@nv-kmcgill53 I still don't understand the exact suggestion here. We could do a linked list but I don't see the value compared to current API. Also, since Triton Server API is a C-API we never can return a non-opaque struct for ABI backward compatibility. The struct fields should always be hidden and we should use methods for for retrieving the fields of the opaque struct. I think if we make those changes the API would be similar to what we have right now.

nv-kmcgill53: My comment on the linked list was to show that returning any structure other than a blob would not be an improvement. Rereading my comment I don't think I communicated this correctly. I agree with @tanmayv25's suggestion for a get-by-name API, that would be just as good.

nv-kmcgill53 (approved):

GuanLuo (approved):

dyastremsky (approved):

tanmayv25 (approved):

PR #175, opened 2023-03-15
Fixing issue in ResolveBackendConfigs

oandreeva-nv:
This PR fixes [issue](https://github.com/triton-inference-server/server/issues/5277) when not all backend configurations were properly resolved.

### DIFF
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -306,15 +306,15 @@ TritonModel::ResolveBackendConfigs(
     }
 
     // add the rest of the global configs
-    if (global_index < global_backend_config.size()) {
-      auto& current_global_setting = global_backend_config.at(global_index);
+    while (global_index < global_backend_config.size()) {
+      auto& current_global_setting = global_backend_config.at(global_index++);
       config.push_back(current_global_setting);
     }
 
     // add the rest of the specific settings
-    if (specific_index < specific_backend_config.size()) {
+    while (specific_index < specific_backend_config.size()) {
       auto& current_specific_setting =
-          specific_backend_config.at(specific_index);
+          specific_backend_config.at(specific_index++);
       config.push_back(current_specific_setting);
     }
   }  // else empty config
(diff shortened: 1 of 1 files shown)


nv-kmcgill53 (dismissed):

GuanLuo (approved):
I think the resolution logic can be simplified to use map as local variable to resolve the "specific setting overrides global setting":
```c++
map<string, string> lconfig;
for (setting : global) {
    lconfig[setting.first] = setting.second;
}
for (setting : specific) {
    lconfig[setting.first] = setting.second;
}
for (map_item : lconfig) {
    config.emplace_back(map_item);
}
...
```

src/backend_model.cc
```diff
@@ -305,15 +305,15 @@ TritonModel::ResolveBackendConfigs(
     }
 
     // add the rest of the global configs
-    if (global_index < global_backend_config.size()) {
-      auto& current_global_setting = global_backend_config.at(global_index);
+    while (global_index < global_backend_config.size()) {
```
GuanLuo, line 308: use `for (;global_index < global_backend_config.size(); ++global_index)` for clarity, same for below

src/backend_model.cc
```diff
@@ -276,45 +277,19 @@ TritonModel::ResolveBackendConfigs(
...
-          0) {
-        config.push_back(current_specific_setting);
-        ++specific_index;
-      } else {
-        config.push_back(current_global_setting);
-        ++global_index;
-      }
+    std::map<std::string, std::string> lconfig;
+    // Accumulate all global settings
+    for (auto& setting : global_backend_config){
```
GuanLuo, line 282: You can get rid of line 262-279 based on the cases already captured in your new changes
```suggestion
    if (global_itr != backend_cmdline_config_map.end()) {
      // Accumulate all global settings
      for (auto& setting : global_itr->second){
      ...
```

oandreeva-nv: that is true, missed that. Thank you

GuanLuo (approved):

PR #178, opened 2023-03-21
Support initializer as inference request input

taoisu:
Background of this PR can be found here: https://github.com/triton-inference-server/server/issues/5489

Essentially, in onnxruntime, initializers (model weights) can be specified as input to override the default weights. The initializers are different in the sense they don't have batch dimension, so changes need to be made in this repo to make sure the shape treatment is correct. In addition, because model weights will be overridden, dynamic batching should be disabled when initializers are to be fed.

### DIFF
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -94,7 +94,7 @@ struct TRITONBACKEND_Batcher;
 ///   }
 ///
 #define TRITONBACKEND_API_VERSION_MAJOR 1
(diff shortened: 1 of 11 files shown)


taoisu:
Hi @GuanLuo thanks again for your suggestion! This is one of the two PRs for supporting the initializer as input feature I was discussing with you earlier, appreciate your help in reviewing the PRs!

taoisu:
Drop the PR per the suggestion.

PR #179, opened 2023-03-28
Prevent infinity loop if cloud client falsely report file as directory

kthui:
This is to prevent infinity loop like:
```
I0327 17:58:53.944686 1 http_server.cc:3372] HTTP request: 2 /v2/repository/models/debug/load
I0327 17:58:53.944736 1 filesystem.cc:2354] Using credential    for path  as://mystorage/triton/model_repo/debug
...
I0327 17:58:54.150433 1 filesystem.cc:2354] Using credential    for path  as://mystorage/triton/model_repo/debug/1/model.py
I0327 17:58:54.241710 1 filesystem.cc:2354] Using credential    for path  as://mystorage/triton/model_repo/debug/1/model.py/
I0327 17:58:54.268449 1 filesystem.cc:2354] Using credential    for path  as://mystorage/triton/model_repo/debug/1/model.py/
I0327 17:58:54.292619 1 filesystem.cc:2354] Using credential    for path  as://mystorage/triton/model_repo/debug/1/model.py/
I0327 17:58:54.330611 1 filesystem.cc:2354] Using credential    for path  as://mystorage/triton/model_repo/debug/1/model.py/
I0327 17:58:54.361050 1 filesystem.cc:2354] Using credential    for path  as://mystorage/triton/model_repo/debug/1/model.py/
... (infinity loop)
```
Ref: https://github.com/triton-inference-server/server/issues/5533#issuecomment-1485631131

When the cloud storage client falsely reports a file as a directory, probably due to permission, the current logic can infinitely loop on an invalid directory path like `as://mystorage/triton/model_repo/debug/1/model.py/` on [`GetModifiedTime()`](https://github.com/triton-inference-server/core/blob/r23.02/src/model_repository_manager.cc#L209-L253). This change prevents the infinity loop by heuristically checking if the name of the contents inside a directory is "". If the name is "", then it is invalid and will stop calling on the recursive loop. However, it is still the ultimate responsibility of the cloud storage client to correctly report if a content is a file or directory.

### DIFF
--- a/src/model_repository_manager.cc
+++ b/src/model_repository_manager.cc
@@ -256,11 +256,6 @@ GetModifiedTime(const std::string& path)
   }
 
   for (const auto& child : contents) {
-    if (child.empty()) {
-      LOG_ERROR << "Failed to determine modification time for '" << path
-                << "': Directory has content with empty name";
-      return 0;
-    }
     const auto full_path = JoinPath({path, child});
     mtime = std::max(mtime, GetModifiedTime(full_path));
   }
--- a/src/model_repository_manager.cc
+++ b/src/model_repository_manager.cc
@@ -256,6 +256,11 @@ GetModifiedTime(const std::string& path)
   }
 
   for (const auto& child : contents) {
+    if (child.empty()) {
+      LOG_ERROR << "Failed to determine modification time for '" << path
+                << "': Directory has content with empty name";
+      return 0;
+    }
     const auto full_path = JoinPath({path, child});
     mtime = std::max(mtime, GetModifiedTime(full_path));
   }
--- a/src/filesystem.cc
+++ b/src/filesystem.cc
@@ -721,6 +721,13 @@ GCSFileSystem::GetDirectoryContents(
     // Let set take care of subdirectory contents
     std::string item = name.substr(item_start, item_end - item_start);
     contents->insert(item);
+
+    // Fail-safe check to ensure the item name is not empty
+    if (item.empty()) {
+      return Status(
+          Status::Code::INTERNAL,
+          "Cannot handle item with empty name at " + path);
+    }
   }
   return Status::Success;
 }
@@ -1135,6 +1142,12 @@ ASFileSystem::GetDirectoryContents(
   auto func = [&](const as::list_blobs_segmented_item& item,
                   const std::string& dir) {
     contents->insert(dir);
+    // Fail-safe check to ensure the item name is not empty
+    if (dir.empty()) {
+      return Status(
+          Status::Code::INTERNAL,
+          "Cannot handle item with empty name at " + path);
+    }
     return Status::Success;
   };
   std::string container, dir_path;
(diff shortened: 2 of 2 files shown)


src/model_repository_manager.cc
```diff
@@ -256,6 +256,11 @@ GetModifiedTime(const std::string& path)
   }
 
   for (const auto& child : contents) {
+    if (child.empty()) {
```
tanmayv25, line 259: Why does a misrepresented file as a directory would have an empty string as a content?
I would have assumed that contents to be an empty set. 
Usually, an actual directory will hold "." and "..".
I am concerned if we are missing some important cases.

tanmayv25: Can there be cases if a directory contains a file with name "" ?

kthui: > Why does a misrepresented file as a directory would have an empty string as a content? I would have assumed that contents to be an empty set.

I think for most cloud storage systems, there is no directory structures which each "file" is stored with its full path "/dir1/dir2/file" (i.e. on a flat table). If for any reason "/dir1/dir2/file" is misinterpreted as a directory, then I imagine any resources nested within "/dir1/dir2/file/" would be "" and it would be invalid, because "/dir1/dir2/file/" is simply "/dir1/dir2/file" which is somewhat an empty set.

> Usually, an actual directory will hold "." and "..".

Yes on Linux, but I think the layout on cloud storages can be different.

> Can there be cases if a directory contains a file with name "" ?

I think it is invalid, because if a parent directory contains a child file with name "", then the child file will have a full path of "/dir1/dir2/", which is equivalent to a full path of "/dir1/dir2" that matches the full path of the parent directory. A full path must be deterministic which cannot reference both the parent directory and the child file at the same time, so it leads to a contradiction. Thus, the child file cannot have name "".

I found some discussions online that some [file systems might allow empty file name](https://stackoverflow.com/questions/68531715/can-an-empty-string-be-a-legal-file-name), but I think it would be invalid on Triton because we use the "/dir1/dir2/file" structure that will leads to the above contradiction if a file name is empty.

tanmayv25, line 259: Shouldn't we add this test in `GetDirectoryContents` implementation? and directly return an error from `GetDirectoryContents` function? 
This way this failing case will be handled by all the callers of this function.

tanmayv25 (changes requested):

src/model_repository_manager.cc
kthui: Yes, thanks for pointing out. Moved the check into each of the three cloud storage implementation. [Move check into GetDirectoryContents](https://github.com/triton-inference-server/core/pull/179/commits/2ac27283e63b4222926df29bec06ee5b08fad766)

tanmayv25 (approved):

GuanLuo (approved):

jbkyang-nvi (approved):

PR #180, opened 2023-03-30
Add request timeout for sequence batcher

GuanLuo:
server PR: https://github.com/triton-inference-server/server/pull/5571

### DIFF
--- a/src/sequence_batch_scheduler.cc
+++ b/src/sequence_batch_scheduler.cc
@@ -580,6 +580,7 @@ SequenceBatchScheduler::Enqueue(std::unique_ptr<InferenceRequest>& irequest)
         "inference requests");
   }
 
+  bool wake_reaper_thread = false;
   std::unique_lock<std::mutex> lock(mu_);
 
   auto sb_itr = sequence_to_batcherseqslot_map_.find(correlation_id);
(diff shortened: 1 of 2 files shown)


src/sequence_batch_scheduler.cc
```diff
@@ -798,85 +835,140 @@ SequenceBatchScheduler::ReaperThread(const int nice)
...
-      // null request. The scheduler thread will interpret the null
-      // request as meaning it should release the sequence slot but
-      // otherwise do nothing with the request.
-      std::unique_ptr<InferenceRequest> null_request;
-      batchers_[batcher_idx]->Enqueue(
-          seq_slot, idle_correlation_id, null_request);
+    // Reap timed out backlog sequence
+    if (now_us >= timeout_timestamp_) {
+      timeout_timestamp_ = std::numeric_limits<uint64_t>::max();
+      std::deque<std::shared_ptr<BacklogQueue>> expired_backlogs;
```
dyastremsky, line 927: Would using a vector be preferable here? You don't seem to be using any properties where deques are superior here (e.g. insertion/removal at the start), so the additional bookkeeping of deques (e.g. associated with them being stored non-contiguously) seems like unnecessary overhead.

tanmayv25 (approved):

src/sequence_batch_scheduler.cc
GuanLuo: I don't think the container type matters too much here. Using vector does help when iterating the element given the contiguity, but rejecting the request is done at the end of reaper loop and the mutex is released so the performance is not critical. Another thing for using vector is that there can be resizing of the container on insertion and causing overhead.

dyastremsky (approved):

PR #183, opened 2023-04-06
Add support for summaries in latency metrics

rmccorm4:


src/metric_model_reporter.cc
```diff
@@ -76,53 +76,123 @@ MetricModelReporter::MetricModelReporter(
...
-      CreateCounterMetric(Metrics::FamilyCacheHitCount(), labels);
-  metric_cache_hit_duration_us_ =
-      CreateCounterMetric(Metrics::FamilyCacheHitDuration(), labels);
-  metric_cache_miss_count_ =
-      CreateCounterMetric(Metrics::FamilyCacheMissCount(), labels);
-  metric_cache_miss_duration_us_ =
-      CreateCounterMetric(Metrics::FamilyCacheMissDuration(), labels);
+  // Global config only for now in config map
+  auto metrics_config_map_ = Metrics::ConfigMap();
+  const auto& metrics_config_ = metrics_config_map_[""];
```
tanmayv25, line 81: Why is their an empty string in the map? I am guessing it is meant to be the configuration group name that you were describing. 
Can you create a global constant in the file and use that instead?

src/metric_model_reporter.cc
```diff
@@ -76,53 +76,123 @@ MetricModelReporter::MetricModelReporter(
...
+
+  // Default behavior is counters for most latency metrics if no types specified
```
tanmayv25, line 83: I believe having MetricsConfig as a struct with all these default values initialized will be better than a standard vector of key value pair.

tanmayv25 (approved):

PR #185, opened 2023-04-08
Disable request_duration summary when caching is enabled

rmccorm4:
When caching is enabled for a model, cache misses spend additional time inserting into the cache (invoked by scheduler) after a backend has returned the response. This additional time has to be considered in the end-to-end request_duration. 
- For `counter` metrics, this is simply added on top of the backend-reported request_duration with an additional `Counter::Increment` API call as it is jts a monotonically increasing number:
  ```
  counter->Increment(backend_duration)
  counter->Increment(cache_miss_duration)
  ```
- For `summary` metrics, the underlying behavior is more complex, and the final request duration (including the cache miss time) should only be reported once:
  ```
  summary->Observe(backend_duration + cache_miss_duration)
  ```
  If the two values were reported separately, they'd be considered 2 separate request durations and skew the derived summary statistics reported (quantiles).

This is rather trick to correctly get both values `backend_duration + cache_miss_duration` so the `request_duration` summary will be disabled for now.

---

For the non-decoupled case, I think per-response statistics would be the cleanest way of reporting the metrics internally, as we can do any internal business logic that we need, and when the response is finally ready, we can report the statistics just once. However, I think there were some issues with the per-response stats feature and it was shelved, so I will discuss some alternatives for now below.

The request may not be valid when the scheduler is handling cache-misses, as a backend can release the request before that happens - so we cannot store the request_duration intermediate values within the request object. 

It may be possible to store a short-lived map of request_durations based on request_id or similar identifier. So when a backend reports stats:
```cpp
// Called from TRITONBACKEND
UpdateSuccess(request_start_ns, request_end_ns, ...) {
  // Cache enabled, should store request_start now and 
  // substract it from cache_end in UpdateCacheMiss()
  if (cache_enabled) {
    request_start_map[identifier] = request_start_ns;
  }
  // No caching, can report duration right away 
  else {
    request_summary.Observe(request_end_ns  - request_start_ns)
  }

  queue_summary.Observe(queue_duration_ns);
  compute_input_summary.Observe(compute_input_duration_ns);
  ...
}

UpdateCacheMiss(cache_miss_start_ns, cache_miss_end_ns) {
  request_start_ns = request_start_map[identifier];
  request_summary.Observe(cache_miss_end_ns - request_start_ns);
  cache_miss_summary.Observe(cache_miss_end_ns - cache_miss_start_ns)
}
```

However I'm not sure what `identifier` will be yet, as `request->Id()` appears to be usually unset unless explicitly set by the user or application?

### DIFF
--- a/src/infer_stats.cc
+++ b/src/infer_stats.cc
@@ -111,8 +111,15 @@ InferenceStatsAggregator::UpdateSuccessWithDuration(
     metric_reporter->IncrementCounter(
         "compute_output_duration", compute_output_duration_ns / 1000);
     // Summary Latencies
-    metric_reporter->ObserveSummary(
-        "request_duration", request_duration_ns / 1000);
+    const auto& reporter_config = metric_reporter->Config();
+    // NOTE: If caching is enabled for this model, don't update request duration
+    // summary with the duration reported from the backend. Instead, the
+    // UpdateSuccessCacheMiss() will add the extra time spent on cache miss
+    // into the request duration and report it there.
+    if (!reporter_config.cache_enabled_) {
+      metric_reporter->ObserveSummary(
+          "request_duration", request_duration_ns / 1000);
+    } 
     metric_reporter->ObserveSummary("queue_duration", queue_duration_ns / 1000);
     metric_reporter->ObserveSummary(
         "compute_input_duration", compute_input_duration_ns / 1000);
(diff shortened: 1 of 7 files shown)


src/infer_stats.cc
```diff
@@ -111,8 +111,12 @@ InferenceStatsAggregator::UpdateSuccessWithDuration(
     metric_reporter->IncrementCounter(
         "compute_output_duration", compute_output_duration_ns / 1000);
     // Summary Latencies
-    metric_reporter->ObserveSummary(
-        "request_duration", request_duration_ns / 1000);
+    const auto& reporter_config = metric_reporter->Config();
+    // FIXME [DLIS-4762]: request summary is disabled when cache is enabled.
```
tanmayv25, line 115: Instead of silently skipping the request_duration in Summary, shouldn't we throw an error when parsing the CLI options? 
"summary metric config not supported when caching is enabled."
I guess we also depend upon whether model_config specifies caching setting, but for now I think it will be okay to be more conservative and throw this error.

src/metric_model_reporter.cc
```diff
@@ -230,13 +242,16 @@ MetricModelReporter::InitializeSummaries(
         &Metrics::FamilyInferenceComputeInferSummary();
     summary_families_["compute_output_duration"] =
         &Metrics::FamilyInferenceComputeOutputSummary();
-    // Cache: Note that counts are included in summary
-    summary_families_["cache_hit_duration"] = &Metrics::FamilyCacheHitSummary();
-    summary_families_["cache_miss_duration"] =
-        &Metrics::FamilyCacheMissSummary();
+    // Only create cache metrics if cache is enabled to reduce metric output
+    if (config_.cache_enabled_) {
```
tanmayv25, line 246: Should we include the other cache metrics in the summary? I think this might be confusing. We can just have summary not work with caching enabled as before.

src/infer_stats.cc
rmccorm4: The request_duration summary is the only metric with an issue around caching. The queue, compute, and cache hit/miss summaries are all still correct and could be reported as summaries.

Since the metrics are not yet individually toggle-able, I only disabled the broken request_duration summary rather than all summary metrics. I figured this would still give some useful information to users that want this feature with cache-enabled models for now.

tanmayv25 (approved):

PR #189, opened 2023-04-12
Add model instance update

kthui:
Related PR: https://github.com/triton-inference-server/server/pull/5635

### DIFF
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -252,7 +252,7 @@ TritonModel::Create(
   }
 
   // Create and initialize the model instances for this model.
-  RETURN_IF_ERROR(TritonModelInstance::CreateInstances(
+  RETURN_IF_ERROR(TritonModelInstance::SetInstances(
       raw_local_model, backend_cmdline_config_map, host_policy_map,
       model_config));
   RETURN_IF_ERROR(local_model->CommitInstances());
@@ -279,9 +279,9 @@ TritonModel::UpdateInstanceGroup(
       &model_config));
   RETURN_IF_ERROR(ValidateInstanceGroup(model_config, min_compute_capability_));
 
-  // Create new instances if needed.
+  // Update the instances to the new config.
   caller_lock->unlock();  // allow inference while creating instances
-  Status status = TritonModelInstance::CreateInstances(
+  Status status = TritonModelInstance::SetInstances(
       this, backend_cmdline_config_map_, host_policy_map_, model_config);
   caller_lock->lock();
   if (!status.IsOk()) {
@@ -290,7 +290,7 @@ TritonModel::UpdateInstanceGroup(
 
   // At this point, the new model config is ready but not yet written into this
   // object. The 'caller_lock' is held, so 'model_lifecycle' will pause any new
-  // inference request. It is safe to move forward and commit the update.
+  // inference request. It is safe to move forward and commit the change.
   RETURN_IF_ERROR(SetModelConfig(model_config));
   RETURN_IF_ERROR(CommitInstances());
   RETURN_IF_ERROR(SetConfiguredScheduler());
@@ -632,6 +632,20 @@ TritonModel::SetSchedulerMutable(std::unique_ptr<Scheduler> scheduler)
   scheduler_ = std::move(scheduler);
 
   return Status::Success;
+}
+
+std::vector<std::shared_ptr<TritonModelInstance>>
+TritonModel::GetInstancesByDevice(int32_t device_id) const
+{
+  std::vector<std::shared_ptr<TritonModelInstance>> result;
+  for (auto* instances : {&instances_, &bg_instances_}) {
+    for (auto& instance : (*instances)) {
+      if (instance->DeviceId() == device_id) {
+        result.push_back(instance);
+      }
+    }
+  }
+  return result;
 }
 
 extern "C" {
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -533,26 +533,6 @@ TritonModel::SetBatchingStrategy(const std::string& batch_libpath)
   return Status::Success;
 }
 
-Status
-TritonModel::Initialize()
-{
-  for (const auto& instance : instances_) {
-    RETURN_IF_ERROR(instance->Initialize());
-  }
-
-  return Status::Success;
-}
-
-Status
-TritonModel::WarmUp()
-{
-  for (const auto& instance : instances_) {
-    RETURN_IF_ERROR(instance->WarmUp());
-  }
-
-  return Status::Success;
-}
-
 TritonModel::TritonModel(
     InferenceServer* server,
     const std::shared_ptr<LocalizedPath>& localized_model_dir,
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -198,7 +198,7 @@ TritonModel::Create(
   // Initialize the model for Triton core usage
   RETURN_IF_ERROR(local_model->Init(is_config_provided));
 
-  // Set 'device_blocking_', initialized to false.
+  // Set 'device_blocking_', it is initialized to false.
   if (local_model->backend_->ExecutionPolicy() ==
       TRITONBACKEND_EXECUTION_DEVICE_BLOCKING) {
     if (model_config.has_sequence_batching()) {
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -165,7 +165,8 @@ TritonModel::Create(
   // Create and initialize the model.
   std::unique_ptr<TritonModel> local_model(new TritonModel(
       server, localized_model_dir, backend, min_compute_capability, version,
-      model_config, auto_complete_config));
+      model_config, auto_complete_config, backend_cmdline_config_map,
+      host_policy_map));
 
   TritonModel* raw_local_model = local_model.get();
 
@@ -197,7 +198,7 @@ TritonModel::Create(
   // Initialize the model for Triton core usage
   RETURN_IF_ERROR(local_model->Init(is_config_provided));
 
-  bool device_blocking = false;
+  // Set 'device_blocking_', initialized to false.
   if (local_model->backend_->ExecutionPolicy() ==
       TRITONBACKEND_EXECUTION_DEVICE_BLOCKING) {
     if (model_config.has_sequence_batching()) {
@@ -205,7 +206,7 @@ TritonModel::Create(
                   "\"TRITONBACKEND_EXECUTION_BLOCKING\" for sequence model \""
                << model_config.name() << "\"";
     } else {
-      device_blocking = true;
+      local_model->device_blocking_ = true;
     }
   }
 
@@ -253,11 +254,46 @@ TritonModel::Create(
   // Create and initialize the model instances for this model.
   RETURN_IF_ERROR(TritonModelInstance::CreateInstances(
       raw_local_model, backend_cmdline_config_map, host_policy_map,
-      model_config, device_blocking));
+      model_config));
+  RETURN_IF_ERROR(local_model->CommitInstances());
 
   RETURN_IF_ERROR(local_model->SetConfiguredScheduler());
 
   *model = std::move(local_model);
+  return Status::Success;
+}
+
+Status
+TritonModel::UpdateInstanceGroup(
+    const inference::ModelConfig& new_model_config,
+    std::unique_lock<std::mutex>* caller_lock)
+{
+  // Generate normalized model config with new instance group.
+  inference::ModelConfig model_config = config_;
+  model_config.clear_instance_group();
+  model_config.mutable_instance_group()->Add(
+      new_model_config.instance_group().begin(), new_model_config.instance_group().end());
+  RETURN_IF_ERROR(NormalizeInstanceGroup(
+      min_compute_capability_, backend_->BackendAttributes().preferred_groups_, &model_config));
+  RETURN_IF_ERROR(
+      ValidateInstanceGroup(model_config, min_compute_capability_));
+
+  // Create new instances if needed.
+  caller_lock->unlock();  // allow inference while creating instances
+  Status status = TritonModelInstance::CreateInstances(
+      this, backend_cmdline_config_map_, host_policy_map_, model_config);
+  caller_lock->lock();
+  if (!status.IsOk()) {
+    return status;
+  }
+
+  std::lock_guard<std::mutex> update_lock(update_mu_);
+
+  // Commit the update.
+  RETURN_IF_ERROR(SetModelConfig(model_config));
+  RETURN_IF_ERROR(CommitInstances());
+  RETURN_IF_ERROR(SetConfiguredScheduler());
+
   return Status::Success;
 }
 
@@ -324,15 +360,13 @@ TritonModel::SetBackendConfigDefaults(
 }
 
 Status
-TritonModel::AddInstance(
-    std::unique_ptr<TritonModelInstance>&& instance, const bool passive)
+TritonModel::RegisterInstance(
+    std::shared_ptr<TritonModelInstance>&& instance, const bool passive)
 {
   if (passive) {
-    passive_instance_group_map_[instance->GroupName()].emplace_back(
-        std::move(instance));
+    bg_passive_instances_.emplace_back(std::move(instance));
   } else {
-    instance_group_map_[instance->GroupName()].emplace_back(
-        std::move(instance));
+    bg_instances_.emplace_back(std::move(instance));
   }
 
   return Status::Success;
@@ -443,7 +477,7 @@ TritonModel::SetConfiguredScheduler()
         0 /* max_queue_delay_microseconds */, &scheduler));
   }
 
-  return SetScheduler(std::move(scheduler));
+  return SetSchedulerMutable(std::move(scheduler));
 }
 
 Status
@@ -502,10 +536,8 @@ TritonModel::SetBatchingStrategy(const std::string& batch_libpath)
 Status
 TritonModel::Initialize()
 {
-  for (const auto& pair : instance_group_map_) {
-    for (const auto& instance : pair.second) {
-      RETURN_IF_ERROR(instance->Initialize());
-    }
+  for (const auto& instance : instances_) {
+    RETURN_IF_ERROR(instance->Initialize());
   }
 
   return Status::Success;
@@ -514,10 +546,8 @@ TritonModel::Initialize()
 Status
 TritonModel::WarmUp()
 {
-  for (const auto& pair : instance_group_map_) {
-    for (const auto& instance : pair.second) {
-      RETURN_IF_ERROR(instance->WarmUp());
-    }
+  for (const auto& instance : instances_) {
+    RETURN_IF_ERROR(instance->WarmUp());
   }
 
   return Status::Success;
@@ -528,11 +558,15 @@ TritonModel::TritonModel(
     const std::shared_ptr<LocalizedPath>& localized_model_dir,
     const std::shared_ptr<TritonBackend>& backend,
     const double min_compute_capability, const int64_t version,
-    const inference::ModelConfig& config, const bool auto_complete_config)
+    const inference::ModelConfig& config, const bool auto_complete_config,
+    const triton::common::BackendCmdlineConfigMap& backend_cmdline_config_map,
+    const triton::common::HostPolicyCmdlineConfigMap& host_policy_map)
     : Model(
           min_compute_capability, localized_model_dir->Path(), version, config),
       server_(server), min_compute_capability_(min_compute_capability),
       auto_complete_config_(auto_complete_config),
+      backend_cmdline_config_map_(backend_cmdline_config_map),
+      host_policy_map_(host_policy_map), device_blocking_(false),
       localized_model_dir_(localized_model_dir), backend_(backend),
       state_(nullptr)
 {
@@ -556,8 +590,10 @@ TritonModel::~TritonModel()
 
   // Explicitly delete/finalize all model instances before finalizing
   // the model itself.
-  instance_group_map_.clear();
-  passive_instance_group_map_.clear();
+  instances_.clear();
+  passive_instances_.clear();
+  bg_instances_.clear();
+  bg_passive_instances_.clear();
 
   // Unregister itself from the rate limiter. Note this should happen
   // after all instances are destructed. Destrucing instances ensures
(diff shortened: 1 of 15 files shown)


src/model_config_utils.h
```diff
@@ -279,4 +279,20 @@ TRITONSERVER_DataType DataTypeToTriton(const inference::DataType dtype);
 /// \return The data type.
 inference::DataType TritonToDataType(const TRITONSERVER_DataType dtype);
 
+/// Check if the model configs are only differ in instance group.
+/// \param old_config The old model config.
+/// \param new_config The new model config.
+/// \return The comparison result.
```
nv-kmcgill53, line 285: Add an explanation of what a return value of `true` vs. `false` is in the comment. 

```suggestion
/// \return TRUE if model configurations are the same not checking instance_group; FALSE if one or more fields, excluding instance_group, are not the same.
```

src/model_config_utils.h
```diff
@@ -279,4 +279,20 @@ TRITONSERVER_DataType DataTypeToTriton(const inference::DataType dtype);
...
+bool IsOnlyInstanceGroupDiffers(
+    const inference::ModelConfig& old_config,
+    const inference::ModelConfig& new_config);
+
+/// Check if both model instance config are equivalent.
```
nv-kmcgill53, line 290: Add to comment that the `name` and `count` fields are excluded from checking.

src/model_config_utils.h
```diff
@@ -279,4 +279,20 @@ TRITONSERVER_DataType DataTypeToTriton(const inference::DataType dtype);
...
+/// \param instance_config_lhs The left hand side instance config.
+/// \param instance_config_rhs The right hand side instance config.
+/// \return True if they are equivalent. False if otherwise.
```
nv-kmcgill53, line 293:
```suggestion
/// \return True if instance_groups are equivalent, not checking 'name' and 'count' fields; False if instance_group fields, excluding 'name' and 'count', are not the same.
```

src/model_config_utils.cc
```diff
@@ -2291,4 +2292,28 @@ TritonToDataType(const TRITONSERVER_DataType dtype)
...
+IsInstanceEquivalent(
+    const inference::ModelInstanceGroup& instance_config_lhs,
+    const inference::ModelInstanceGroup& instance_config_rhs)
+{
+  ::google::protobuf::util::MessageDifferencer pb_diff;
+  pb_diff.IgnoreField(
+      instance_config_lhs.descriptor()->FindFieldByLowercaseName("name"));
+  pb_diff.IgnoreField(
+      instance_config_lhs.descriptor()->FindFieldByLowercaseName("count"));
+  return pb_diff.Compare(instance_config_lhs, instance_config_rhs);
```
nv-kmcgill53, line 2316: In the case of multiple instance groups, what does this function return if the order  of the groups is swapped? i.e. instance_group: [ A, B ] --> instance_group: [ B, A ]

nv-kmcgill53 (changes requested):

src/model_config_utils.h
kthui: Update commit: [Improve model_config_utils comments](https://github.com/triton-inference-server/core/pull/189/commits/97f7c89c3e06e26d15ce7e05589723f6ac97c0c5)

src/model_config_utils.h
kthui: Update commit: [Improve model_config_utils comments](https://github.com/triton-inference-server/core/pull/189/commits/97f7c89c3e06e26d15ce7e05589723f6ac97c0c5)

src/model_config_utils.h
kthui: Update commit: [Improve model_config_utils comments](https://github.com/triton-inference-server/core/pull/189/commits/97f7c89c3e06e26d15ce7e05589723f6ac97c0c5)

src/model_config_utils.cc
kthui: The type naming is a bit confusing, `const inference::ModelInstanceGroup& instance_config_lhs` is the config of an instance, not the entire instance group array.

> i.e. instance_group: [ A, B ] --> instance_group: [ B, A ]

This function will be comparing i.e. `A` vs `A` or `A` vs `B`

The order swapping is included in the test case: [shuffle the instance](https://github.com/triton-inference-server/server/pull/5635/files#diff-40ca2af021f4288f3fc6ae6071a244bca6350a0e657ab2822b61650cac266931R137-R143), which ensures no instance is created or destructed after swapping the order.

src/backend_model_instance.h
```diff
@@ -48,16 +49,46 @@ class InferenceRequest;
...
+  struct SecondaryDevice {
+    SecondaryDevice(const std::string kind, const int64_t id)
+        : kind_(kind), id_(id)
+    {
+    }
+    const std::string kind_;
+    const int64_t id_;
+  };
+
+  class Signature {
```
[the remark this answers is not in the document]
tanmayv25: Why expose the functionality of EnableMatching and DisableMatching is not very clear to me?

src/rate_limiter.cc
```diff
@@ -538,24 +568,39 @@ RateLimiter::ModelContext::AllocateInstanceIfAvailable()
...
 void
-RateLimiter::ModelContext::RequestRemoval()
+RateLimiter::ModelContext::RemoveInstance(ModelInstanceContext* instance)
 {
-  removal_in_progress_ = true;
+  std::lock_guard<std::recursive_mutex> lk1(sched_request_queue_mtx_);
+  std::lock_guard<std::recursive_mutex> lk2(avbl_instances_mtx_);
+
+  PriorityQueue new_avbl_instances;
+  while (!avbl_instances_.empty()) {
```
tanmayv25, line 594: What if the instance is not in the avbl_instance_ when the RemoveInstance is invoked?

src/backend_model_instance.cc
```diff
@@ -358,35 +371,32 @@ TritonModelInstance::CreateInstance(
     RETURN_IF_ERROR(model->Server()->GetRateLimiter()->RegisterModelInstance(
         local_instance.get(), rate_limiter_config));
     RETURN_IF_ERROR(local_instance->SetBackendThread(
-        kind, device_id, device_blocking, device_to_thread_map));
+        kind, device_id, model->DeviceBlocking()));
   }
 
-  RETURN_IF_ERROR(model->AddInstance(std::move(local_instance), passive));
```
tanmayv25, line 364: I think it would be better to decouple the creation and registration of the instances.

src/model_config_utils.h
```diff
@@ -279,4 +279,24 @@ TRITONSERVER_DataType DataTypeToTriton(const inference::DataType dtype);
 /// \return The data type.
 inference::DataType TritonToDataType(const TRITONSERVER_DataType dtype);
 
+/// Check if the model configs may only differ in instance group.
+/// \param old_config The old model config.
+/// \param new_config The new model config.
+/// \return True if the model configs are the same without checking instance
+/// group. False if they are different without checking instance group.
+bool IsOnlyInstanceGroupMayDiffers(
```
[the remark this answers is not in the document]
kthui: I think we can rename it to `DifferInNonInstanceGroupConfig`?

| non instance group config changed | instance group config changed | -> | Can update? |
| --- | --- | --- | --- |
| No | No | -> | Yes |
| No | Yes | -> | Yes |
| Yes | No | -> | No |
| Yes | Yes | -> | No |

From the table above, we do not care if instance group is altered or not (since it is excluded), the only thing matter is whether config other than instance group is changed or not.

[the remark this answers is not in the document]
tanmayv25: DifferInNonInstanceGroupConfig is good.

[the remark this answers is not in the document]
kthui: Eventually, `EquivalentInNonInstanceGroupConfig(...)` is used, because it avoids double negating like `!DifferInNonInstanceGroupConfig(...)`.

Update commit: [Improve model config util naming](https://github.com/triton-inference-server/core/pull/189/commits/c01ea3ba8f037068b384c38b51074a1b266caf57)

src/model_config_utils.h
```diff
@@ -279,4 +279,24 @@ TRITONSERVER_DataType DataTypeToTriton(const inference::DataType dtype);
 /// \return The data type.
 inference::DataType TritonToDataType(const TRITONSERVER_DataType dtype);
 
+/// Check if the model configs may only differ in instance group.
```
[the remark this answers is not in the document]
kthui: The comments are adjusted.

Update commit: [Improve model config util naming](https://github.com/triton-inference-server/core/pull/189/commits/c01ea3ba8f037068b384c38b51074a1b266caf57)

src/model_config_utils.h
```diff
@@ -279,4 +279,24 @@ TRITONSERVER_DataType DataTypeToTriton(const inference::DataType dtype);
...
+/// \param old_config The old model config.
+/// \param new_config The new model config.
+/// \return True if the model configs are the same without checking instance
```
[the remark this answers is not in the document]
kthui: The comments are adjusted: [Improve model config util naming](https://github.com/triton-inference-server/core/pull/189/commits/c01ea3ba8f037068b384c38b51074a1b266caf57)

src/model_config_utils.h
```diff
@@ -279,4 +279,24 @@ TRITONSERVER_DataType DataTypeToTriton(const inference::DataType dtype);
...
+/// group. False if they are different without checking instance group.
```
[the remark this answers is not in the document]
kthui: The comments are adjusted: [Improve model config util naming](https://github.com/triton-inference-server/core/pull/189/commits/c01ea3ba8f037068b384c38b51074a1b266caf57)

src/backend_model.cc
```diff
@@ -251,16 +251,52 @@ TritonModel::Create(
   }
 
   // Create and initialize the model instances for this model.
```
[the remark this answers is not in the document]
kthui: Thanks for catching, updated: [Fix outdated comment](https://github.com/triton-inference-server/core/pull/189/commits/67bcec93f8f805ad0f424be4bbb0286b43e9fe8e)

src/backend_model.cc
```diff
@@ -197,15 +197,15 @@ TritonModel::Create(
   // Initialize the model for Triton core usage
   RETURN_IF_ERROR(local_model->Init(is_config_provided));
 
-  bool device_blocking = false;
+  // Set 'device_blocking_', it is initialized to false.
```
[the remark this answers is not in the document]
kthui: Update commit: [Refactor setting device blocking](https://github.com/triton-inference-server/core/pull/189/commits/1ece4394246899512d9e3876884401304d3a718c)

src/backend_model.h
```diff
@@ -99,10 +114,18 @@ class TritonModel : public Model {
...
       const std::shared_ptr<TritonBackend>& backend,
       const double min_compute_capability, const int64_t version,
-      const inference::ModelConfig& config, const bool auto_complete_config);
+      const inference::ModelConfig& config, const bool auto_complete_config,
+      const triton::common::BackendCmdlineConfigMap& backend_cmdline_config_map,
+      const triton::common::HostPolicyCmdlineConfigMap& host_policy_map);
 
-  // Set the scheduler based on the model configuration. The scheduler
-  // can only be set once for a backend.
+  // Replace the foreground instances with background instances.
```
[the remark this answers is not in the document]
kthui: Actually, the background instance is for holding instances that will become foreground later. An instance can exist in both foreground and background at the same time, when it is about to be reused during an update. I think [this comment](https://github.com/triton-inference-server/core/pull/189#discussion_r1170629174) also supplements this part.

Some example flow/interaction between foreground and background:

When the model is initially created:
  - The background instance is empty.
  - `SetInstance` cannot match any foreground instance, because foreground is also empty.
  - All requested instances are created, into the background.
  - Commit the instances, so they are swapped into foreground (background <-- swap --> foreground).
  - All background instances are removed. (empty in this case)
  - Load complete.

An instance is removed during an update:
  - `SetInstance` is able to match all requested instance in foreground (all existing instances except the removed one as it is not requested). so the shared pointers are copied into background.
  - Commit the instances, so they are swapped into foreground (background <-- swap --> foreground).
  - All background instances are removed. The removed instance only has one reference count and it is in the background, so it is removed. The non-removed instances have reference count in the foreground, so they are not removed.
  - Update complete.

An instance is added during an update:
  - `SetInstance` is able to match all requested instance in foreground (all existing instances). so the shared pointers are copied into background.
  - `SetInstance` cannot find the added instance, so it is created into the background.
  - Commit the instances, so they are swapped into foreground (background <-- swap --> foreground).
  - All background instances are removed. All instances (except the added one) are in the background, but they have reference count in the foreground, so they are not removed. (no instances are removed in fact)
  - Update complete.

src/backend_model_instance.cc
kthui: Update commit: [Decouple registration from instance creation](https://github.com/triton-inference-server/core/pull/189/commits/0565adcaf6e6b7b7600d60777189468a5eb07c5e)

src/backend_model_instance.h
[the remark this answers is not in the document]
kthui: > Why expose the functionality of EnableMatching and DisableMatching is not very clear to me?

When an instance is reused during an update, the model config could specify multiple instances that are equivalent to each other (i.e. `count = 5`). Once an instance is matched and reused, the `RegisterInstance()` will `DisableMatching` on the instance, so it cannot be reused more than once during an instance update.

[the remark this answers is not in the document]
kthui: Update commit: [Add more doc on Signature](https://github.com/triton-inference-server/core/pull/189/commits/47258625bd7ba02bb1d6519b478eed237507afe4)

src/backend_model.h
```diff
@@ -75,15 +74,31 @@ class TritonModel : public Model {
...
       TRITONSERVER_Message* updated_config_message);
   const std::shared_ptr<TritonBackend>& Backend() const { return backend_; }
-  const std::unordered_map<std::string, TritonInstanceGroup>& InstanceGroups()
-      const
+  const std::vector<std::shared_ptr<TritonModelInstance>>& Instances() const
   {
-    return instance_group_map_;
+    return instances_;
   }
+  std::shared_ptr<TritonModelInstance> FindInstance(
```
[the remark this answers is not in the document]
kthui: Yes.
Update commit: [Add comments to backend_model.h functions](https://github.com/triton-inference-server/core/pull/189/commits/2b8a25ab00cebc7d0203f55cd6e1f2bcf1a3dec6)

src/rate_limiter.cc
kthui: Update commit: [Use function instead of macro for ShareBackendThread](https://github.com/triton-inference-server/core/pull/189/commits/16d0054d7f5c7e63da636d180a1eeb4af61e1ae9)

src/backend_model_instance.cc
```diff
@@ -57,6 +57,10 @@
 
 namespace triton { namespace core {
 
+// Condition when a backend thread is shared.
```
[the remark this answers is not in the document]
kthui: Update commit: [Use function instead of macro for ShareBackendThread](https://github.com/triton-inference-server/core/pull/189/commits/16d0054d7f5c7e63da636d180a1eeb4af61e1ae9)

src/rate_limiter.cc
kthui: I think inference will stop and remain stopped on the removing instance before it is removed from the rate limiter, but I will verify it with the [new rate limiter test](https://github.com/triton-inference-server/server/pull/5635#issuecomment-1512273360) that will vary the resource while adding/removing instances.

kthui:
Rebased

kthui:
New commit: [Reorder create instance error return and register instance](https://github.com/triton-inference-server/core/pull/189/commits/103940c541303ed6d4fdc2ea78481d0148343dbb)
Rebased and apply PR #194

src/backend_model.cc
```diff
@@ -250,17 +241,71 @@ TritonModel::Create(
...
-  // Create and initialize the model instances for this model.
-  RETURN_IF_ERROR(TritonModelInstance::CreateInstances(
+  // Create or update the model instances for this model.
+  RETURN_IF_ERROR(TritonModelInstance::SetInstances(
       raw_local_model, backend_cmdline_config_map, host_policy_map,
-      model_config, device_blocking));
+      model_config));
+  RETURN_IF_ERROR(local_model->CommitInstances());
 
   RETURN_IF_ERROR(local_model->SetConfiguredScheduler());
```
tanmayv25, line 250: Swapping the scheduler without proper check for ongoing sequences in sequence batcher can cause some of the requests for allocated streams to run into first request doesn't specify START flag.

The future requests for a sequence allocated in previous scheduler should still be routed through the same scheduler. I believe the only accurate solution for scaling instances up/down should be to dynamically adjust it within the same sequence_batcher object. @GuanLuo feel free to provide your suggestions/correction here.

src/backend_model.cc
```diff
@@ -250,17 +241,71 @@ TritonModel::Create(
...
+  if (!status.IsOk()) {
+    return status;
+  }
+
+  // At this point, the new model config is ready but not yet written into this
+  // object. The 'caller_lock' is held, so 'model_lifecycle' will pause any new
+  // inference request. It is safe to move forward and commit the change.
+  RETURN_IF_ERROR(SetModelConfig(model_config));
+  RETURN_IF_ERROR(CommitInstances());
+  RETURN_IF_ERROR(SetConfiguredScheduler());
```
tanmayv25, line 286: As described above, might not work for sequence batcher models.

src/backend_model.h
```diff
@@ -99,10 +124,21 @@ class TritonModel : public Model {
...
+      const triton::common::BackendCmdlineConfigMap& backend_cmdline_config_map,
+      const triton::common::HostPolicyCmdlineConfigMap& host_policy_map);
 
-  // Set the scheduler based on the model configuration. The scheduler
-  // can only be set once for a backend.
+  // Replace the foreground instances with background instances.
+  Status CommitInstances();
+
+  // Set the 'device_blocking_' flag.
+  Status SetExecutionPolicy(const inference::ModelConfig& model_config);
```
tanmayv25, line 135: GetExecutionPolicy()

src/backend_model.h
```diff
@@ -99,10 +124,21 @@ class TritonModel : public Model {
...
+      const inference::ModelConfig& config, const bool auto_complete_config,
+      const triton::common::BackendCmdlineConfigMap& backend_cmdline_config_map,
+      const triton::common::HostPolicyCmdlineConfigMap& host_policy_map);
 
-  // Set the scheduler based on the model configuration. The scheduler
-  // can only be set once for a backend.
+  // Replace the foreground instances with background instances.
+  Status CommitInstances();
+
+  // Set the 'device_blocking_' flag.
```
tanmayv25, line 134: // Gets the execution policy setting from the backend.

src/backend_model_instance.cc
```diff
@@ -247,6 +253,21 @@ TritonModelInstance::CreateInstances(
         const auto& kind = std::get<1>(is);
         const auto& id = std::get<2>(is);
 
+        const Signature signature(group, id);
+        // Check if an existing instance can be re-used.
+        if (!ShareBackendThread(model->DeviceBlocking(), kind)) {
+          auto existing_instance = model->FindInstance(signature);
```
tanmayv25, line 259: Add the link to the jira ticket to improve the search process as a TO-DO.

tanmayv25: One solution could be to use hash key.

src/rate_limiter.cc
```diff
@@ -91,6 +92,44 @@ RateLimiter::RegisterModelInstance(
...
+    i_it->second->WaitForRemoval();
+    if (!ignore_resources_and_priority_) {
+      resource_manager_->RemoveModelInstance(i_it->second.get());
+    }
+    model_context.RemoveInstance(i_it->second.get());
+    model_instances.erase(i_it);
+  }
+
+  if (!ignore_resources_and_priority_) {
+    RETURN_IF_ERROR(resource_manager_->UpdateResourceLimits());
```
tanmayv25, line 116: We can skip these update if there is only a model instance count change.  UpdateResourceLimits will not generate new results in such cases.
Add DLIS-4821 as the proposed optimization as a TO-DO.

src/backend_model.cc
```diff
@@ -197,17 +197,8 @@ TritonModel::Create(
...
-      TRITONBACKEND_EXECUTION_DEVICE_BLOCKING) {
-    if (model_config.has_sequence_batching()) {
-      LOG_INFO << "Overriding execution policy to "
-                  "\"TRITONBACKEND_EXECUTION_BLOCKING\" for sequence model \""
-               << model_config.name() << "\"";
-    } else {
-      device_blocking = true;
-    }
-  }
+  // Set 'device_blocking_'.
```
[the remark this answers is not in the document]
kthui: Updated: [Rename SetExecutionPolicy to GetExecutionPolicy](https://github.com/triton-inference-server/core/pull/189/commits/fe6eab259c11d2f078b716be515ff2e32d0e322c)

src/backend_model.h
kthui: Updated: [Rename SetExecutionPolicy to GetExecutionPolicy](https://github.com/triton-inference-server/core/pull/189/commits/fe6eab259c11d2f078b716be515ff2e32d0e322c)

src/backend_model.h
kthui: Updated: [Rename SetExecutionPolicy to GetExecutionPolicy](https://github.com/triton-inference-server/core/pull/189/commits/fe6eab259c11d2f078b716be515ff2e32d0e322c)

src/backend_model_instance.cc
kthui: Updated: [Add links to JIRA](https://github.com/triton-inference-server/core/pull/189/commits/9f65d1267f872ff863c7b0d66910a159e6a3f561)

src/rate_limiter.cc
kthui: Updated: [Add links to JIRA](https://github.com/triton-inference-server/core/pull/189/commits/9f65d1267f872ff863c7b0d66910a159e6a3f561)

src/backend_model.cc
kthui: Yes, but please correct me if I am wrong. I think the same issue currently exists with the `model_lifecycle` model update, because it will swap from pointing to the background model to the foreground, once the new foreground model loading is complete, without knowing anything about the scheduler. Then, the same problem can happen if the swap happens in the middle of a sequence.

I think another possible fix is to block any model update (including instance update) on the model with ongoing sequence(s) until the ongoing sequence(s) completes or times out. At the same time, any other inference requests, not related to the ongoing sequence(s), will pause until the update completes.

If the issue is fixed for model instance update now, we will introduce a behavior change for the type of updates that was previously handled by the `model_lifecycle` update logic. Do you think we might just disable model instance update on sequence model for now, and fix both the `model_lifecycle` and 'instance update' sequence model issue in another ticket?

tanmayv25: > Do you think we might just disable model instance update on sequence model for now, and fix both the model_lifecycle and 'instance update' sequence model issue in another ticket?

Yes. I will let @GuanLuo  comment on the model reloads with sequence batcher. As per my current reading, I too think we will run into same issue when re-loading any model with sequence batching on.

nv-kmcgill53 (approved):

PR #191, opened 2023-04-14
Add device memory tracker. Report memory usage in statistic

GuanLuo:


src/device_memory_tracker.h
```diff
@@ -0,0 +1,265 @@
...
+void
+DeviceMemoryTracker::UntrackThreadMemoryUsage(ScopedMemoryUsage* usage)
+{
+  uint64_t id;
+  if (tracker_ == nullptr) {
+    InitTrace();
+  }
+  CUPTI_CALL(cuptiActivityPopExternalCorrelationId(
+      CUPTI_EXTERNAL_CORRELATION_KIND_UNKNOWN, &id));
+  CUPTI_CALL(cuptiActivityFlushAll(0));
```
tanmayv25, line 180: Can you confirm that this is accurate? Didn't went into details of CUPTI tracing, but I would assume flushing out activities for the popped id above.

src/device_memory_tracker.h
```diff
@@ -0,0 +1,265 @@
...
+      ScopedMemoryUsage* usage = nullptr;
+      {
+        std::lock_guard<std::mutex> lk(mtx_);
+        auto it = activity_to_memory_usage_.find(memory_record->correlationId);
+        if (it != activity_to_memory_usage_.end()) {
+          usage = reinterpret_cast<ScopedMemoryUsage*>(it->second);
+          activity_to_memory_usage_.erase(it);
+        }
+      }
+      // Igore memory record that is not associated with a ScopedMemoryUsage
```
tanmayv25, line 199: Ignore

src/device_memory_tracker.h
```diff
@@ -0,0 +1,265 @@
...
+      // object
+      if (usage != nullptr) {
+        const bool is_allocation =
+            (memory_record->memoryOperationType ==
+             CUPTI_ACTIVITY_MEMORY_OPERATION_TYPE_ALLOCATION);
+        if (is_allocation || (memory_record->memoryOperationType ==
```
tanmayv25, line 205: How about adding another variable `is_release` just for the consistency. It will make the code more readable.

tanmayv25 (commented):
Did you check loading a single instance of model (say X MB) vs loading multiple instances of the same model in parallel? Does the memory consumption scales evenly?

src/device_memory_tracker.h
GuanLuo: What do you mean? From what I understand it just flush out the activities recorded, by the time this function is called, all activities has been done and recorded. Let me know if that is not the case.

src/device_memory_tracker.h
```diff
@@ -0,0 +1,265 @@
...
+    bool tracked_{false};
+  };
+  static void InitTrace();
+
+  // Currently can distinguish activity by correlation id which is
+  // thread specific, which implies that switching threads to handle
+  // activities may not be tracked?
+  // The memory usage will be updated until it's untracked, usage must
+  // be valid until untrack is called.
+  // [WIP] Document thread-local nature..
```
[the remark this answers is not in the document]
GuanLuo: If they are processed in a separate thread, so warmup yes, most of the instance initialization no except like Python backend.

src/device_memory_tracker.h
```diff
@@ -0,0 +1,265 @@
+
+#pragma once
+
+#include <cupti.h>
```
rmccorm4, line 4: Does this work on a CPU only build? Or a container with no GPU driver? Might need to be protected `#ifdef TRITON_ENABLE_GPU`

src/backend_model.cc
```diff
@@ -162,6 +163,8 @@ TritonModel::Create(
         ValidateInstanceGroup(model_config, min_compute_capability));
   }
 
+  DeviceMemoryTracker::ScopedMemoryUsage usage;
+  DeviceMemoryTracker::TrackThreadMemoryUsage(&usage);
```
Tabrizian, line 167: Do you think we can make the memory usage tracking a self-reported metric by the backend using the backend API instead of automatically tracking it in the core? This would ensure that the value reported can be customized if needed. For example, the Python backend can report the value for the stub process.

src/device_memory_tracker.h
```diff
@@ -0,0 +1,265 @@
...
+}  // namespace
+
+void
+DeviceMemoryTracker::InitTrace()
+{
+  if (tracker_ == nullptr) {
+    tracker_.reset(new DeviceMemoryTracker());
+    CUPTI_CALL(
+        cuptiActivityRegisterCallbacks(bufferRequested, bufferCompleted));
+    CUPTI_CALL(cuptiActivityEnable(CUPTI_ACTIVITY_KIND_RUNTIME));
```
Tabrizian, line 153: Do we need to disable the activity to make sure it doesn't add overhead for actual inferences?

GuanLuo: I believe enable/disable tracing is a global property so switching that will interfere with other memory usage collection, I have done some small performance comparison with the feature enable and the performance impact is not obvious (on the simple test models). But I do agree that it will add overhead to inference, I would like to see the impact on small model but with large number of kernel launches (speed on launching CUDA activities is bottleneck), but I don't have such a model in hand.

src/backend_model.cc
GuanLuo: I was thinking about that but I think this will be an alternative if the backend does provide the usage info. The current way require no changes in the backend implementation so all backends get it for free if they are allocating / freeing memory on demand. We can later on add new field to "backend attribute" API so if the backend is implemented to provide more accurate usage, we will collect through the backend instead of through cupti

Tabrizian: > The current way require no changes in the backend implementation so all backends get it for free

I think that is great but I think reporting an inaccurate number for Python/FT backends would be bad UX. Maybe we can make the memory tracking a backend utility that the backend can opt-in?

GuanLuo: That is probably a better idea, but I need to do some experiments on how the CUPIT API usage may need to be changed if moved to backends (that shouldn't need to consider each other). At the minimum, will be backend provided callback to provide memory usage (API for backend to prepare the struct) when requested by core. And depending on CUPIT usage, Triton wrapped API like the memory pool.

include/triton/core/tritonbackend.h
```diff
@@ -1411,6 +1411,36 @@ TRITONBACKEND_ISPEC TRITONSERVER_Error* TRITONBACKEND_GetBackendAttribute(
...
+/// 128 bytes on CUDA device 1. Then 'usage' should have first two entries set,
+/// one has the buffer attributes of "type GPU, id 0, 64 bytes" and the
+/// other has "type GPU, id 1, 256 bytes". The list and the list entries are
+/// owned by Triton.
+/// \param usage_size This parameter is both input and output of the function.
+/// As input, it represents the number of entries in the list that the backend
+/// may write memory usage into. As output, it returns the number of entries
+/// that the backend has written the memory usage into. If the backend needs
+/// more entries to write the memory info, -1 should be set for 'usage_size' and
+/// error should be returned with TRITONSERVER_ERROR_INVALID_ARG.
```
rmccorm4, line 1438: 1. How does Triton core know the correct number of entries to prepare and send? via model config instance counts and such? Does this apply to backends that handle things like instance count differently like FasterTransformer or maybe KIND_MODEL in TF backend?
2. If Triton is wrong on the number of entries, should the backend instead correct Triton core on the necessary number of entries rather than just sending -1? Or is it expected via (1) that Triton core will eventually get the right number, and it can only be wrong due to some timing sensitive things like a new instance being loaded on demand while querying for the info?

GuanLuo: Each entry corresponds to the total memory usage of a physical device. I don't think there will be too many (~10x?). Usually if the API is designed in this way (caller owns the array), caller will expand the size and try again. We can expose a linked list structure if this is a concern.

The query is invoked after the initialize/finalize when Triton thinks the memory usage is changed. Sounds like the documentation isn't really clear on how it works?

rmccorm4: I haven't looked at any implementation/usage of this API yet if there is any, but just from reading these docs and your comment, my interpretation is that Triton either knows how many devices are in-use and trys with that number, or has no idea and just has some default guess, such as 10, and retries until success?:
```cpp
int32_t assumed_buffers = 10;
int32_t num_buffers = 10;

TRITONSERVER_BufferAttributes** buffer_list = new BufferList(num_buffers);
// Query memory usage with an assumed sufficient number of buffers for backend to fill
auto err = TRITONBACKEND_ModelMemoryUsage(model, buffer_list, &num_buffers);
// num_buffers == -1 or err==TRITONSERVER_INVALID_ARG, but err could be raised from other args
while (num_buffers == -1) {
  // retry with more buffers?
  num_buffers += 1
  // or num_buffers *= 2 , etc. depending on pros/cons of retries vs unnecessary allocs
}

if (assumed_buffers != num_buffers) {
  std::cout << "triton assumed " << assumed_buffers << " but backend needed: " << num_buffers << std::endl
}
```

GuanLuo: yes

include/triton/core/tritonbackend.h
```diff
@@ -1411,6 +1411,36 @@ TRITONBACKEND_ISPEC TRITONSERVER_Error* TRITONBACKEND_GetBackendAttribute(
...
+/// called multiple times within the lifetime of the model (between
+/// TRITONBACKEND_ModelInitialize and TRITONBACKEND_ModelFinalize).
+/// The backend may return error to indicate failure to get the latest memory
+/// usage.
+///
+/// \param model The model.
+/// \param usage The list of buffer attributes that the backend can record the
+/// memory usage, each entry should record the total memory usage of a given
+/// memory type and id. For example, if the model itself occupies 64 bytes on
+/// CUDA device 0 and it has two model instances that each instance occupies
```
nnshah1, line 1428: Question: Should the api be in terms of model instances or models? Do we need to specify this is for GPU memory - or does it include cpu memory as well?

include/triton/core/tritonbackend.h
```diff
@@ -1411,6 +1411,36 @@ TRITONBACKEND_ISPEC TRITONSERVER_Error* TRITONBACKEND_GetBackendAttribute(
     TRITONBACKEND_Backend* backend,
     TRITONBACKEND_BackendAttribute* backend_attributes);
 
+/// Query the total memory usage of the model and its associated model
```
Tabrizian, line 1414: Instead of Triton Core calling this function, I was wondering whether we can delegate this responsibility to the backend?

The backend can call `TRITONBACKEND_ReportModelMemoryUsage` whenever it wants and we would update the usage statistics for this model. For example, this would allow the backend to update the usage during actual execution if it notices any memory changes.

include/triton/core/tritonbackend.h
```diff
@@ -1411,6 +1411,36 @@ TRITONBACKEND_ISPEC TRITONSERVER_Error* TRITONBACKEND_GetBackendAttribute(
...
+/// \param model The model.
+/// \param usage The list of buffer attributes that the backend can record the
+/// memory usage, each entry should record the total memory usage of a given
+/// memory type and id. For example, if the model itself occupies 64 bytes on
+/// CUDA device 0 and it has two model instances that each instance occupies
+/// 128 bytes on CUDA device 1. Then 'usage' should have first two entries set,
+/// one has the buffer attributes of "type GPU, id 0, 64 bytes" and the
+/// other has "type GPU, id 1, 256 bytes". The list and the list entries are
+/// owned by Triton.
+/// \param usage_size This parameter is both input and output of the function.
```
Tabrizian, line 1433: Why not delegate this responsibility completely to the backend and make this only an output parameter? For example, it can create any number of buffer attributes it needs for reporting memory usage.

include/triton/core/tritonbackend.h
```diff
@@ -1437,6 +1467,25 @@ TRITONBACKEND_BackendAttributeAddPreferredInstanceGroup(
...
 
+/// Set whether the backend enables memory tracker. If set to true, Triton
+/// will expect the backend to track the memory allocated for individual model
+/// and assoicated model instances, and the backend must implement
+/// TRITONBACKEND_ModelMemoryUsage callback to report total memory usage of the
+/// model. Triton will invoke the callabck after model memory usage may have
+/// changed, for example, after instance initialization or finalization.
+///
+/// \param backend_attributes The backend attributes object.
+/// \param kind The kind of the instance group.
```
nnshah1, line 1478: comments seem to not match signature

include/triton/core/tritonbackend.h
GuanLuo: This is a good idea

rmccorm4: I was thinking the same based on other thread of back and forth with core and num_usage==-1 🚀

include/triton/core/tritonbackend.h
```diff
@@ -1206,6 +1227,28 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_ModelInstanceState(
...
+/// each entry should record the total memory usage of a given memory type and
+/// id. For example, if the instance itself occupies 64 bytes on each of
+/// CUDA device 0 and CUDA device 1. Then 'usage' should have first two entries
+/// set, one has the buffer attributes of "type GPU, id 0, 64 bytes" and the
+/// other has "type GPU, id 1, 64 bytes". 'usage' is owned by the backend and
+/// may be released after the function returns.
+/// \param usage_size The number of entries in 'usage'.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONBACKEND_DECLSPEC TRITONSERVER_Error*
+TRITONBACKEND_ModelInstanceReportMemoryUsage(
```
Tabrizian, line 1248: I was wondering why do we need two separate APIs to report the usage (ModelInstanceReportMemoryUsage and ModelReportMemoryUsage)? Is it because the CUPTI APIs track the memory per thread and the thread that loaded the model would be different than the thread that loads the model instance?

Tabrizian: How about having only a single model-level memory reporting that would account for the total memory usage? I understand that we can provide this breakdown because of the CUPTI thread level support but I think having a single total memory usage would be less confusing for the users. Also, it would create a lot of confusion specially when model sharing is enabled as you point out.

nnshah1: After some thought - I agree that keeping it at model level with the expectation that all memory usage for all instances are combined there is simpler to specify. Initially I thought model instance made more sense but after a little more experience with the backendapi - I think it might be tricky to specify the correct behavior. We have to be explicit if model_size = sum(model_instances) + delta - since we don't have metrics at instance level (yet?) - maybe we defer this to future and leave it out for now.

GuanLuo: >but I think having a single total memory usage would be less confusing for the users.

What do you refer to for "users"?
- For user of Triton, they are somewhat far from the backend API, we can collect the memory usage as finest granularity and report the total usage for them, which is the current implementation at server frontend.
- For user of the backend API, I would say it is easier for them to manage the usage records. The issue with single model-level reporting is that they need to keep the summary and modify it as instances are being created / destroyed. Not to mention that the change of model instances may be in parallel in the future (i.e. RFE on concurrent instance loading) which further complicates the management. With this two-level reporting, the scope is constrained that model reports usage w.r.t. TritonModel and instance reports usage w.r.t. TritonModelInstance, and the usage record is not being shared (can even use it as local).

i.e. before
```
ModelState {
   ...
   usage_;
} 

ModelInit() {
  track and set "usage_"
}

InstanceInit() {
  declare lusage;
  ... track in lusage;
  get "usage_" from model state
  // thread-safety ..
  update "usage_"
  report model usage
}

InstanceFini() {
  // similar to above
}
```
After
```
ModelInit() {
  declare and track in lusage;
  report model usage;
}

InstanceInit() {
  declare and track in lusage;
  report instant usage;
}

InstanceFini() {
  // No extra handling, core already knows the instance usage
}
```

nnshah1: So I understand: Is the suggestion that the core aggregates? So the "model" memory usage is "model + instance" as reported by the backend? As long as the semantics are clear and not left to the backend to decide - then ok. I want to make sure its clear what the backend implementation needs to be - i.e. model is aggregate or model is in addition to instances.

GuanLuo: > So the "model" memory usage is "model + instance" as reported by the backend?

Yes if `"model" memory usage` refers to what will be reported by statistic endpoint.

The backend API has mentioned what are the usage to be reported, let me know if it needs to be improved:
```
/// Report the memory usage of the model that will be released on
/// TRITONBACKEND_ModelFinalize. ...
TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_ModelReportMemoryUsage
```
```
/// Report the memory usage of the model instnace that will be released on
/// TRITONBACKEND_ModelInstanceFinalize. ...
TRITONBACKEND_DECLSPEC TRITONSERVER_Error*
TRITONBACKEND_ModelInstanceReportMemoryUsage(
```

nnshah1: I think just giving the example / overall description that the total usage is model + model instances  (so backend doesn't double count by accident). Model Finalize could also release all instances etc. - so I think we should just be extra clear.

src/CMakeLists.txt
```diff
@@ -497,6 +498,7 @@ if(${TRITON_ENABLE_GPU})
     PRIVATE
       ${CNMEM_LIBRARY}
       CUDA::cudart
+      CUDA::cupti
```
nnshah1, line 501: is this needed here if memory tracker moved to backend?

include/triton/core/tritonbackend.h
```diff
@@ -1066,6 +1066,27 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_ModelState(
 TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_ModelSetState(
     TRITONBACKEND_Model* model, void* state);
 
+/// Report the memory usage of the model that will be released on
+/// TRITONBACKEND_ModelFinalize. The backend may call this function within the
+/// lifecyle of the TRITONBACKEND_Model object (between
```
rmccorm4, line 1071:
```suggestion
/// lifecycle of the TRITONBACKEND_Model object (between
```

include/triton/core/tritonbackend.h
```diff
@@ -1206,6 +1227,28 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_ModelInstanceState(
 TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_ModelInstanceSetState(
     TRITONBACKEND_ModelInstance* instance, void* state);
 
+/// Report the memory usage of the model instnace that will be released on
```
rmccorm4, line 1230:
```suggestion
/// Report the memory usage of the model instance that will be released on
```

include/triton/core/tritonbackend.h
```diff
@@ -1206,6 +1227,28 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_ModelInstanceState(
...
+/// TRITONBACKEND_ModelInstanceFinalize. The backend may call this function
+/// within the lifecyle of the TRITONBACKEND_Model object (between
```
rmccorm4, line 1232:
```suggestion
/// within the lifecycle of the TRITONBACKEND_Model object (between
```

src/backend_model_instance.h
```diff
@@ -125,6 +125,27 @@ class TritonModelInstance {
 
   MetricModelReporter* MetricReporter() const { return reporter_.get(); }
 
+  // Directly call from C API, so arguments are in the same style
+  void SetMemoryUsage(BufferAttributes** memory_usage, uint32_t usage_size)
```
rmccorm4, line 129: This could probably be implicitly converted to a `boost::span<BufferAttributes*>` if you wanted to not use the C style in implementation.

src/model.h
```diff
@@ -146,6 +146,48 @@ class Model {
     return stats_aggregator_;
   }
 
+  // Directly call from C API, so arguments are in the same style
+  void SetMemoryUsage(BufferAttributes** memory_usage, uint32_t usage_size)
```
rmccorm4, line 150: This could probably be implicitly converted to a `boost::span<BufferAttributes*>` if you wanted to not use the C style in implementation.

src/tritonserver.cc
```diff
@@ -2755,6 +2755,38 @@ TRITONSERVER_ServerModelStatistics(
...
+          }
+          case TRITONSERVER_MEMORY_CPU_PINNED: {
+            type = "CPU_PINNED";
+            break;
+          }
+          case TRITONSERVER_MEMORY_GPU: {
+            type = "GPU";
+            break;
+          }
+        }
```
rmccorm4, line 2781: Can use `TRITONSERVER_MemoryTypeString`?

src/model.h
```diff
@@ -146,6 +146,48 @@ class Model {
...
+    auto lusage = AccumulatedInstanceMemoryUsage();
+    // accumulate model memory usage
+    {
+      std::lock_guard<std::mutex> lk(usage_mtx_);
+      for (const auto& mem_type_map : memory_usage_) {
+        const auto& mem_type = mem_type_map.first;
+        for (const auto& mem_id_map : mem_type_map.second) {
+          const auto& mem_id = mem_id_map.first;
+          const auto& byte_size = mem_id_map.second;
+          lusage[mem_type][mem_id] = byte_size;
```
nnshah1, line 174: qq: would this be += to add the instance usage to the model usage?

src/tritonserver.cc
```diff
@@ -2755,6 +2755,38 @@ TRITONSERVER_ServerModelStatistics(
     for (const auto& version : mv_pair.second) {
       std::shared_ptr<tc::Model> model;
       RETURN_IF_STATUS_ERROR(lserver->GetModel(mv_pair.first, version, &model));
+
+      // Add memory usage
+      triton::common::TritonJson::Value memory_usage(
```
nnshah1, line 2760: Question: If a backend doesn't support memory usage collection - do we return all 0 - or leave the section out - or indicate via some other value that it is not applicable?

nnshah1, line 2760: Question: In case memory tracking isn't enabled - we include an empty array is that correct? Is it possible to tell the difference between tracking enabled and value being 0 (i.e. no memory reported?)

include/triton/core/tritonbackend.h
Tabrizian: Sorry for the delayed response. I think my only concern is that we may not able to bucket the model's memory usage to "model" and "model instance" buckets. For example, if we add support for memory usage tracking for activations (i.e., during the actual inference) we may not able "Untrack" it since it may not be returned to the system on model instance unloading.

On a related note, I was wondering how would this feature work with PyTorch and TF backend. Since they would not release the memory back to the system, we would probably untrack only a small portion (if any during model unloading). If another model ends up re-using the same memory pool, we wouldn't track any CUDA activity since it is using old allocations. Can this happen for TF and PyTorch? Is my understanding correct?

GuanLuo: For your first point, this is why the API is really generic that it is up to the backend to report the memory usage, if the backend is not going to release the memory on model instance unloading, then those memory should not be reported to the API.

For your second point, this is exactly what happened in TF and PyTorch, CUPTI collects no additional allocation and thus reports nothing. Ideally we should figure out what is the proper way for handling TF / PyTorch models and use corresponding implementation for those backends instead of the backend utils. One good example of this will be Python backend where it is not track-able by current implementation and the alternative is present that we can map the process-level usage to the model / instance usage.

After some thought, maybe the function name can be `XxxReportReleasableMemoryUsage` to show the intent of the API.

Tabrizian: I think the API naming is fine. It is probably more an issue with TF and PyTorch. We may be able to support the same API using Rapids RMM instead of Cupti.

src/tritonserver.cc
GuanLuo: Good question, the backend can report 0 value to be explicit that no memory will be released for the corresponding memory type. It is probably safer to say that the memory type not being included / reported only means it is not tracked, and it may be released somehow ("pessimistic" estimation). In other words, the value reported is what the backend is confident to release on model / instance unloading.

src/backend_model.h
```diff
@@ -137,6 +137,27 @@ class TritonModel : public Model {
   // Set or update the scheduler.
   Status SetSchedulerMutable(std::unique_ptr<Scheduler> scheduler);
 
+  std::map<TRITONSERVER_MemoryType, std::map<int64_t, size_t>>
+  AccumulatedInstanceMemoryUsage() const override
+  {
+    std::map<TRITONSERVER_MemoryType, std::map<int64_t, size_t>> res;
+    // [FIXME] thread-safety on encountering instance change
```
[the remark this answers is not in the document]
GuanLuo: I would defer this to @kthui to address it, the concern is introduced by the instance update and I am not sure what is the thread-safety guarantee implemented there.

src/backend_model_instance.h
```diff
@@ -125,6 +125,27 @@ class TritonModelInstance {
...
+  {
+    std::map<TRITONSERVER_MemoryType, std::map<int64_t, size_t>> lusage;
+    for (uint32_t idx = 0; idx < usage_size; ++idx) {
+      const auto mem_type = memory_usage[idx]->MemoryType();
+      const auto mem_id = memory_usage[idx]->MemoryTypeId();
+      const auto byte_size = memory_usage[idx]->ByteSize();
+      lusage[mem_type][mem_id] = byte_size;
```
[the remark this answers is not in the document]
GuanLuo: On first use, map will implicitly construct the value which is 0 by default, this is typo that should have been "+=".

include/triton/core/tritonbackend.h
nnshah1: I think the API and metrics will be interpreted as memory_usage and releasable memory are the same. I think the expectation will be that the memory usage reflects the size of the model / instance and how many you could add / how much memory can be reclaimed. If there is a mismatch it may be confusing - for the cases that CUPTI works - my understanding is that they do match? (Overall usage and releasable usage per model?)

src/backend_model_instance.h
```diff
@@ -126,10 +126,10 @@ class TritonModelInstance {
   MetricModelReporter* MetricReporter() const { return reporter_.get(); }
 
   // Directly call from C API, so arguments are in the same style
-  void SetMemoryUsage(BufferAttributes** memory_usage, uint32_t usage_size)
+  void SetMemoryUsage(boost::span<BufferAttributes*> memory_usage)
   {
     std::map<TRITONSERVER_MemoryType, std::map<int64_t, size_t>> lusage;
-    for (uint32_t idx = 0; idx < usage_size; ++idx) {
+    for (uint32_t idx = 0; idx < memory_usage.size(); ++idx) {
```
rmccorm4: You should also be able to do `for (const auto& usage : memory_usage)` with the span so no need for `idx`, if you want to.

src/model.h
```diff
@@ -146,11 +147,10 @@ class Model {
     return stats_aggregator_;
   }
 
-  // Directly call from C API, so arguments are in the same style
-  void SetMemoryUsage(BufferAttributes** memory_usage, uint32_t usage_size)
+  void SetMemoryUsage(boost::span<BufferAttributes*> memory_usage)
   {
     std::map<TRITONSERVER_MemoryType, std::map<int64_t, size_t>> lusage;
-    for (uint32_t idx = 0; idx < usage_size; ++idx) {
+    for (uint32_t idx = 0; idx < memory_usage.size(); ++idx) {
```
rmccorm4, line 153: same as other comment, can use `for each` loop with span.

nnshah1 (approved):

nnshah1 (approved):

tanmayv25 (approved):

Tabrizian (approved):

tanmayv25 (approved):

src/backend_model.h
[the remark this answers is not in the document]
rmccorm4: @kthui any comments?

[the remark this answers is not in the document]
kthui: Sorry, just saw it now. I think the only concurrency issue is when [`instances_` and `passive_instances_` are swapped](https://github.com/triton-inference-server/core/blob/main/src/backend_model.cc#L404-L405) and when they are accessed below `for (const auto& instances : {&instances_, &passive_instances_}) {`. I think a simple instance swap mutex belonging to the 'backend_model.h' can avoid the potential problem.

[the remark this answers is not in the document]
kthui: >instance update and I am not sure what is the thread-safety guarantee implemented there

The safety guarantee is all changes to the instances are to be made into background instances (active or passive), which the current foreground instances are untouched, so they can be safely accessed (but not modified without other protection). Once the background instances are ready (the complete set of instances after the update), they will be swapped and become the current foreground instances. Thus, during swap, the instances cannot be accessed.

PR #192, opened 2023-04-17
update c api to properly use the struct keyword

ryanolson:
Rust will automatically create bindings when the C API is properly structured. Unfortunately, we were missing the `struct` keyword in a number of places.  If you are using this header in a C++ setting, you can get away with it, but the way Rust is using LLMV to create the bindings, [the C API must properly conform to the standards](https://learn.microsoft.com/en-us/cpp/cpp/struct-cpp?view=msvc-170#using-a-structure).

> In C, you must explicitly use the struct keyword to declare a structure. In C++, you do not need to use the struct keyword after the type has been defined.
>
> You have the option of declaring variables when the structure type is defined by placing one or more comma-separated variable names between the closing brace and the semicolon.

### DIFF
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -104,7 +104,7 @@ struct TRITONSERVER_MetricFamily;
 /// \param minor Returns the TRITONSERVER API minor version supported
 /// by Triton.
 /// \return a TRITONSERVER_Error indicating success or failure.
-TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_ApiVersion(
+TRITONSERVER_DECLSPEC struct TRITONSERVER_Error* TRITONSERVER_ApiVersion(
     uint32_t* major, uint32_t* minor);
 
 /// TRITONSERVER_DataType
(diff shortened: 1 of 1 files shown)


include/triton/core/tritonserver.h
```diff
@@ -317,20 +317,20 @@ typedef enum TRITONSERVER_errorcode_enum {
...
-TRITONSERVER_DECLSPEC void TRITONSERVER_ErrorDelete(TRITONSERVER_Error* error);
+TRITONSERVER_DECLSPEC void TRITONSERVER_ErrorDelete(struct TRITONSERVER_Error* error);
 
 /// Get the error code.
 ///
 /// \param error The error object.
 /// \return The error code.
-TRITONSERVER_DECLSPEC TRITONSERVER_Error_Code
-TRITONSERVER_ErrorCode(TRITONSERVER_Error* error);
+TRITONSERVER_DECLSPEC struct TRITONSERVER_Error_Code
```
ryanolson, line 332: i goofed here.  shoudl shouldn't be a struct

```suggestion
TRITONSERVER_DECLSPEC TRITONSERVER_Error_Code
```

ryanolson (commented):
Error_Code doesn't need to have a struct on it.

PR #193, opened 2023-04-17
Reduce default model load thread count to 4

rmccorm4:
CLI change corresponding to https://github.com/triton-inference-server/server/pull/5650

### DIFF
--- a/src/server.cc
+++ b/src/server.cc
@@ -104,8 +104,7 @@ InferenceServer::InferenceServer()
   exit_timeout_secs_ = 30;
   pinned_memory_pool_size_ = 1 << 28;
   buffer_manager_thread_count_ = 0;
-  model_load_thread_count_ =
-      std::max(2u, 2 * std::thread::hardware_concurrency());
+  model_load_thread_count_ = 4;
   enable_model_namespacing_ = false;
 
 #ifdef TRITON_ENABLE_GPU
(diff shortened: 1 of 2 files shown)


Tabrizian (approved):

kthui (approved):

kthui:
@rmccorm4 Do you think it might be a better balance if we use a slightly higher default concurrency? Like `std::min(16, std::thread::hardware_concurrency())`, because the low value could be more limiting on hardware that can easily have more than 32 threads, but I think we can see how does it impact the L0_infer load time first.

rmccorm4:
> Do you think it might be a better balance if we use a slightly higher default concurrency?  ... the low value could be more limiting on hardware that can easily have more than 32 threads, but I think we can see how does it impact the L0_infer load time first.

I think the low value is a better default, as I think the average case is for models to be loaded at startup and not frequently changed. There's no need for so many threads to be sitting idle by default in such case. Users that don't expect this flow can set their necessary concurrency. 

re: `L0_infer` we should manually set `--model-load-thread-count` with the necessary values for expected speedup. Maybe we can also improve the test to see the model repository setup times for `--model-load-thread-count=4` < `--model-load-thread-count=8` < 16 < ... , etc.

dyastremsky (approved):

PR #194, opened 2023-04-20
Serialize the model instance registration in resource manager

tanmayv25:
Fixes the intermittence in L0_rate_limiter.

### DIFF
--- a/src/rate_limiter.cc
+++ b/src/rate_limiter.cc
@@ -72,8 +72,17 @@ RateLimiter::RegisterModelInstance(
     model_context.AddSpecificRequestQueue();
 
     if (!ignore_resources_and_priority_) {
+      // As there can be mulitple models being loaded concurrently, need
+      // to hold a lock to protect the resource counts.
+      // Without this serialization instances of other models might fail
+      // to load because of the resource constraints in other models.
+      std::lock_guard<std::mutex> lk(resource_manager_mtx_);
       resource_manager_->AddModelInstance(model_instances.back().get());
-      RETURN_IF_ERROR(resource_manager_->UpdateResourceLimits());
+      const auto& status = resource_manager_->UpdateResourceLimits();
+      if(!status.IsOk()) {
+        resource_manager_->RemoveModelInstance(model_instances.back().get());
+	return status;
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


kthui (approved):

src/rate_limiter.cc
```diff
@@ -72,8 +72,17 @@ RateLimiter::RegisterModelInstance(
...
+      // Without this serialization instances of other models might fail
+      // to load because of the resource constraints in other models.
+      std::lock_guard<std::mutex> lk(resource_manager_mtx_);
       resource_manager_->AddModelInstance(model_instances.back().get());
-      RETURN_IF_ERROR(resource_manager_->UpdateResourceLimits());
+      const auto& status = resource_manager_->UpdateResourceLimits();
+      if(!status.IsOk()) {
+        resource_manager_->RemoveModelInstance(model_instances.back().get());
+	return status;
+      }
```
rmccorm4, line 85: Can you explain this new behavior? It doesn't look like the instance was removed on failure previously. Was this a separate bug/issue?

tanmayv25: Earlier the model instance were removed from resource manager when the model failed to load and its destructor was called. https://github.com/triton-inference-server/core/blob/main/src/backend_model.cc#L562-L566
This was not a problem because the loading of the next model would start only after the destruction/unregistration was  complete for a failed model load.

tanmayv25: Now this call will not find the model instance in the resource manager and do nothing.
https://github.com/triton-inference-server/core/blob/main/src/rate_limiter.cc#L98

rmccorm4 (dismissed):
Nice find :tada:

src/rate_limiter.cc
oandreeva-nv: Am I right that since that call will not find the model instance, it will report "Can not find an instance to remove", as in here: https://github.com/triton-inference-server/core/blob/main/src/rate_limiter.cc#L747

kthui (approved):

oandreeva-nv (approved):

src/rate_limiter.cc
tanmayv25: In the second call yes. But the status is ignored.

oandreeva-nv (changes requested):
Could you please change copyright in rate_limiter.h

dyastremsky (approved):

oandreeva-nv (approved):

PR #195, opened 2023-04-26
Update README and versions for 23.04 branch

mc-nv:


Tabrizian (approved):
It is good that we're updating these versions but it is ok if we don't. It is just an example.

PR #196, opened 2023-05-02
Return full error when custom batcher init fails

dyastremsky:
- Autoformatting
- Returning full error (type, message) when custom batcher init fails. This was previously being lost due to the error being deleted before the string was copied.

Related server PR: https://github.com/triton-inference-server/server/pull/5729


krishung5 (approved):
LGTM!

PR #197, opened 2023-05-05
Add fix for rate limiter hang

kthui:
The first commit [Fix rate limiter deadlock](https://github.com/triton-inference-server/core/pull/197/commits/a75cf3228ddc2e784ad2a6d677d2b7afbc254aa6) reorders the mutex acquiring order inside the rate limiter, to prevent deadlock.

The second commit [Add temporary fix for the issue](https://github.com/triton-inference-server/core/pull/197/commits/2ecc36b735ebc3b2b27b72299efec2d6875f323f) adds a temporary fix to discard any pending inference request after the assigned backend thread has stopped, to prevent the rate limiter hang while waiting for `ModelInstanceContext` to unload. This part of the fix should be replaced with a more permanent fix that prevent any inference request from queueing on the rate limiter before the stop payload is added into the backend thread.

### DIFF
--- a/src/rate_limiter.cc
+++ b/src/rate_limiter.cc
@@ -27,6 +27,7 @@
 #include "rate_limiter.h"
 
 #include <limits>
+#include "server.h"
 #include "triton/common/logging.h"
 
 namespace triton { namespace core {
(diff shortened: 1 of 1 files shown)


kthui:
The fix will be implemented by: https://github.com/triton-inference-server/core/pull/199

PR #200, opened 2023-05-10
Fix possible rate limiter hang

kthui:
This PR is to add https://github.com/triton-inference-server/core/pull/199 fix into r23.05.


tanmayv25:
@kthui I believe the process is to cherry-pick the same exact commit from the main branch to the release branch.

PR #201, opened 2023-05-12
Avoid possible rate limiter hang on r23.05

kthui:
…stance removal (#199)

* [WIP] Rate limiter instance remove and exit payload

* Adjust update to scheduler and model config

### DIFF
--- a/src/rate_limiter.cc
+++ b/src/rate_limiter.cc
@@ -104,7 +104,6 @@ RateLimiter::UnregisterModelInstance(TritonModelInstance* triton_model_instance)
   auto& model_instances = model_instance_ctxs_[model];
   auto i_it = model_instances.find(triton_model_instance);
   if (i_it != model_instances.end()) {
-    i_it->second->WaitForRemoval();
     if (!ignore_resources_and_priority_) {
       resource_manager_->RemoveModelInstance(i_it->second.get());
     }
@@ -143,7 +142,6 @@ RateLimiter::UnregisterModel(const TritonModel* model)
 
     model_context.RequestRemoval();
     for (const auto& instance : model_instance_ctxs_[model]) {
-      instance.second->WaitForRemoval();
       if (!ignore_resources_and_priority_) {
         resource_manager_->RemoveModelInstance(instance.second.get());
       }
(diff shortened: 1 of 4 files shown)


src/rate_limiter.cc
```diff
@@ -319,6 +317,19 @@ RateLimiter::GetPayload(
...
 RateLimiter::PayloadRelease(std::shared_ptr<Payload>& payload)
 {
+  // If this is an exit payload, the instance must not be staged once marked as
+  // available, so mark the instance as removing.
+  if (payload->GetOpType() == Payload::Operation::EXIT) {
+    std::lock_guard<std::mutex> lk(model_instance_ctx_mtx_);
+    auto& instances = model_instance_ctxs_[payload->GetInstance()->Model()];
+    auto it = instances.find(payload->GetInstance());
+    if (it == instances.end()) {
+      LOG_INFO << "Should not print this ";
```
nv-kmcgill53, line 327: This should be an error stating we could not find the instance in question.

src/rate_limiter.cc
```diff
@@ -319,6 +317,19 @@ RateLimiter::GetPayload(
 void
 RateLimiter::PayloadRelease(std::shared_ptr<Payload>& payload)
 {
+  // If this is an exit payload, the instance must not be staged once marked as
+  // available, so mark the instance as removing.
```
nv-kmcgill53, line 321: Just for my understanding, this is done after the response has been sent?

nv-kmcgill53 (changes requested):

src/rate_limiter.cc
kthui: This should never be printed. It is a failsafe check to leave some clue for debugging, if the user is to report some strange failure in the future. On the surface, the trigger for the print is "we could not find the instance in question", but in reality the instance should always be found, which "Should not print this" is a stronger indication on something is deeply wrong rather than something seems acceptable.

src/rate_limiter.cc
kthui: No, this is called immediately after the inference request(s) inside the payload is sent to the backend for inferencing, so the response is likely not yet available when this happens.

nv-kmcgill53 (approved):

PR #204, opened 2023-05-19
Fix nullptr dereference on OldestSequenceBatch

kthui:
Reword `LOG_VERBOSE(1)` due to `irequest` null pointer dereference. This bug will result in segmentation fault if the reaper thread is to remove a timed-out sequence and the user sets log verbose 1 or above.


rmccorm4 (approved):
Nice find

nnshah1 (approved):

PR #207, opened 2023-05-22
Add ability to update existing schedulers

kthui:
Tests: https://github.com/triton-inference-server/server/pull/5831

### DIFF
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -244,9 +244,8 @@ TritonModel::Create(
   RETURN_IF_ERROR(TritonModelInstance::SetInstances(
       raw_local_model, backend_cmdline_config_map, host_policy_map,
       model_config));
-  RETURN_IF_ERROR(local_model->CommitInstances());
-
   RETURN_IF_ERROR(local_model->SetConfiguredScheduler());
+  RETURN_IF_ERROR(local_model->CommitInstances());
 
   *model = std::move(local_model);
   return Status::Success;
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -253,9 +253,7 @@ TritonModel::Create(
 }
 
 Status
-TritonModel::UpdateInstanceGroup(
-    const inference::ModelConfig& new_model_config,
-    std::unique_lock<std::mutex>* caller_lock)
+TritonModel::UpdateInstanceGroup(const inference::ModelConfig& new_model_config)
 {
   // Generate normalized model config with new instance group.
   inference::ModelConfig model_config = config_;
@@ -268,25 +266,17 @@ TritonModel::UpdateInstanceGroup(
       &model_config));
   RETURN_IF_ERROR(ValidateInstanceGroup(model_config, min_compute_capability_));
 
-  // Update the instances to the new config.
-  caller_lock->unlock();  // allow inference while creating instances
-  Status status = TritonModelInstance::SetInstances(
-      this, backend_cmdline_config_map_, host_policy_map_, model_config);
-  caller_lock->lock();
-  if (!status.IsOk()) {
-    // Remove any pending instances if created.
-    bg_instances_.clear();
-    bg_passive_instances_.clear();
-    return status;
-  }
+  // Prepare the new instances on the new config.
+  RETURN_IF_ERROR(TritonModelInstance::SetInstances(
+      this, backend_cmdline_config_map_, host_policy_map_, model_config));
 
-  // At this point, the new model config is ready but not yet written into this
-  // object. The 'caller_lock' is held, so 'model_lifecycle' will pause any new
-  // inference request. It is safe to move forward and commit the change.
+  // At this point, the new model config and instances are ready but not yet
+  // written into this object. Update the scheduler and retrieve the lock
+  // holding off new inference requests, and then commit the changes.
+  std::unique_ptr<std::lock_guard<std::mutex>> lock;
+  RETURN_IF_ERROR(scheduler_->Update(&lock));
   config_.mutable_instance_group()->Swap(model_config.mutable_instance_group());
   RETURN_IF_ERROR(CommitInstances());
-  // Only model owned dynamic batch scheduler can be updated currently, so there
-  // is no need to update the scheduler.
 
   return Status::Success;
 }
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -274,7 +274,7 @@ TritonModel::UpdateInstanceGroup(const inference::ModelConfig& new_model_config)
   // holding off new inference requests, and then commit the changes.
   std::unique_ptr<std::lock_guard<std::mutex>> lock;
   RETURN_IF_ERROR(scheduler_->Update(&lock));
-  config_.mutable_instance_group()->Swap(model_config.mutable_instance_group());
+  *config_.mutable_instance_group() = model_config.instance_group();
   RETURN_IF_ERROR(CommitInstances());
 
   return Status::Success;
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -264,17 +264,16 @@ TritonModel::UpdateInstanceGroup(const inference::ModelConfig& new_model_config)
       min_compute_capability_, backend_->BackendAttributes().preferred_groups_,
       &model_config));
   RETURN_IF_ERROR(ValidateInstanceGroup(model_config, min_compute_capability_));
+  // Updating the 'config_' has to be an atomic operation, because it can be
+  // accessed by other threads concurrently.
+  *config_.mutable_instance_group() = model_config.instance_group();
 
   // Prepare the new instances on the new config.
   RETURN_IF_ERROR(TritonModelInstance::SetInstances(
       this, backend_cmdline_config_map_, host_policy_map_, model_config));
-
-  // At this point, the new model config and instances are ready but not yet
-  // written into this object. Update the scheduler and retrieve the lock
-  // holding off new inference requests, and then commit the changes.
-  std::unique_ptr<std::lock_guard<std::mutex>> lock;
-  RETURN_IF_ERROR(scheduler_->Update(&lock));
-  *config_.mutable_instance_group() = model_config.instance_group();
+  // Update the scheduler while the union of the old and new set of instances
+  // are presence.
+  RETURN_IF_ERROR(scheduler_->Update());
   RETURN_IF_ERROR(CommitInstances());
 
   return Status::Success;
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -245,7 +245,7 @@ TritonModel::Create(
       raw_local_model, backend_cmdline_config_map, host_policy_map,
       model_config));
   RETURN_IF_ERROR(local_model->SetConfiguredScheduler());
-  RETURN_IF_ERROR(local_model->CommitInstances());
+  local_model->CommitInstances();
 
   *model = std::move(local_model);
   return Status::Success;
@@ -264,17 +264,30 @@ TritonModel::UpdateInstanceGroup(const inference::ModelConfig& new_model_config)
       min_compute_capability_, backend_->BackendAttributes().preferred_groups_,
       &model_config));
   RETURN_IF_ERROR(ValidateInstanceGroup(model_config, min_compute_capability_));
-  // Updating the 'config_' has to be an atomic operation, because it can be
-  // accessed by other threads concurrently.
-  *config_.mutable_instance_group() = model_config.instance_group();
 
   // Prepare the new instances on the new config.
-  RETURN_IF_ERROR(TritonModelInstance::SetInstances(
-      this, backend_cmdline_config_map_, host_policy_map_, model_config));
-  // Update the scheduler while the union of the old and new set of instances
-  // are presence.
-  RETURN_IF_ERROR(scheduler_->Update());
-  RETURN_IF_ERROR(CommitInstances());
+  Status status = TritonModelInstance::SetInstances(
+      this, backend_cmdline_config_map_, host_policy_map_, model_config);
+  if (!status.IsOk()) {
+    ClearBackgroundInstances();
+    return status;
+  }
+
+  // Update the scheduler if this is a sequence model.
+  if (model_config.has_sequence_batching()) {
+    SequenceBatchScheduler* sched = (SequenceBatchScheduler*)scheduler_.get();
+    if (sched == nullptr) {
+      ClearBackgroundInstances();
+      return Status(
+          Status::Code::INTERNAL,
+          "Unable to downcast from Scheduler to SequenceBatchScheduler");
+    }
+    sched->Update();
+  }
+
+  // Commit the instance update.
+  CommitInstances();
+  *config_.mutable_instance_group() = model_config.instance_group();
 
   return Status::Success;
 }
@@ -394,15 +407,19 @@ TritonModel::RegisterInstance(
   return Status::Success;
 }
 
-Status
+void
 TritonModel::CommitInstances()
 {
   instances_.swap(bg_instances_);
   passive_instances_.swap(bg_passive_instances_);
+  ClearBackgroundInstances();
+}
+
+void
+TritonModel::ClearBackgroundInstances()
+{
   bg_instances_.clear();
   bg_passive_instances_.clear();
-
-  return Status::Success;
 }
 
 std::vector<std::shared_ptr<TritonModelInstance>>
@@ -621,8 +638,7 @@ TritonModel::~TritonModel()
   // the model itself.
   instances_.clear();
   passive_instances_.clear();
-  bg_instances_.clear();
-  bg_passive_instances_.clear();
+  ClearBackgroundInstances();
 
   // Unregister itself from the rate limiter. Note this should happen
   // after all instances are destructed. Destrucing instances ensures
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -425,10 +425,21 @@ TritonModel::PrepareInstances(
     for (int32_t c = 0; c < group.count(); ++c) {
       std::string instance_name{group.name() + "_" + std::to_string(c)};
       const bool passive = group.passive();
-      std::vector<std::tuple<
-          std::string, TRITONSERVER_InstanceGroupKind, int32_t,
-          const inference::ModelRateLimiter*>>
-          instance_settings;
+      struct InstanceSetting {
+        InstanceSetting(
+            const std::string& policy_name, TRITONSERVER_InstanceGroupKind kind,
+            int32_t device_id,
+            const inference::ModelRateLimiter* rate_limiter_config)
+            : policy_name_(policy_name), kind_(kind), device_id_(device_id),
+              rate_limiter_config_(rate_limiter_config)
+        {
+        }
+        const std::string policy_name_;
+        const TRITONSERVER_InstanceGroupKind kind_;
+        const int32_t device_id_;
+        const inference::ModelRateLimiter* rate_limiter_config_;
+      };
+      std::vector<InstanceSetting> instance_settings;
       if (group.kind() == inference::ModelInstanceGroup::KIND_CPU) {
         instance_settings.emplace_back(
             group.host_policy().empty() ? "cpu" : group.host_policy(),
@@ -453,19 +464,14 @@ TritonModel::PrepareInstances(
             std::string("instance_group kind ") +
                 ModelInstanceGroup_Kind_Name(group.kind()) + " not supported");
       }
-      for (const auto& instance_setting : instance_settings) {
-        const std::string& policy_name = std::get<0>(instance_setting);
-        const auto& kind = std::get<1>(instance_setting);
-        const auto& device_id = std::get<2>(instance_setting);
-        const inference::ModelRateLimiter* rate_limiter_config =
-            std::get<3>(instance_setting);
-
+      for (const auto& is : instance_settings) {
         // All the information for the requested instance is ready. Create a
         // signature that identifies the requested instance.
-        const TritonModelInstance::Signature signature(group, device_id);
+        const TritonModelInstance::Signature signature(group, is.device_id_);
 
         // Check if the requested instance can reuse an existing instance.
-        if (!TritonModelInstance::ShareBackendThread(DeviceBlocking(), kind)) {
+        if (!TritonModelInstance::ShareBackendThread(
+                DeviceBlocking(), is.kind_)) {
           auto itr = existing_instances.find(signature);
           if (itr != existing_instances.end() && !itr->second.empty()) {
             auto existing_instance = itr->second.back();
@@ -483,11 +489,11 @@ TritonModel::PrepareInstances(
         // new instance.
         std::shared_ptr<TritonModelInstance> new_instance;
         LOG_VERBOSE(2) << "Creating model instance named '" << instance_name
-                       << "' with device id '" << device_id << "'";
+                       << "' with device id '" << is.device_id_ << "'";
         RETURN_IF_ERROR(TritonModelInstance::CreateInstance(
-            this, instance_name, signature, kind, device_id, profile_names,
-            passive, policy_name, *rate_limiter_config, secondary_devices,
-            &new_instance));
+            this, instance_name, signature, is.kind_, is.device_id_,
+            profile_names, passive, is.policy_name_, *is.rate_limiter_config_,
+            secondary_devices, &new_instance));
         added_instances->push_back(new_instance);
         RegisterBackgroundInstance(std::move(new_instance), passive);
       }
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -406,6 +406,9 @@ TritonModel::PrepareInstances(
       std::vector<std::shared_ptr<TritonModelInstance>>>
       existing_instances = IndexInstances();
 
+  // Iterates over all the requested instances on the model config, and decides
+  // if each requested instance can reuse an existing instance or a new instance
+  // is needed.
   for (const auto& group : model_config.instance_group()) {
     std::vector<std::string> profile_names;
     for (const auto& profile_name : group.profile()) {
@@ -666,7 +669,9 @@ TritonModel::UpdateConfiguredScheduler(
     return sched->Update(added_instances, removed_instances);
   }
 
-  // Non-sequence scheduler does not need to be updated.
+  // Non-sequence scheduler does not need to be updated, because other
+  // schedulers do not require the information on model instances to function,
+  // and only interact with the rate limiter.
   return Status::Success;
 }
 
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -460,7 +460,7 @@ TritonModel::PrepareInstances(
         const inference::ModelRateLimiter* rate_limiter_config =
             std::get<3>(instance_setting);
 
-        // All the information for the reqested instance is ready. Create a
+        // All the information for the requested instance is ready. Create a
         // signature that identifies the requested instance.
         const TritonModelInstance::Signature signature(group, device_id);
 
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -282,7 +282,11 @@ TritonModel::UpdateInstanceGroup(const inference::ModelConfig& new_model_config)
           Status::Code::INTERNAL,
           "Unable to downcast from Scheduler to SequenceBatchScheduler");
     }
-    sched->Update();
+    Status status = sched->Update();
+    if (!status.IsOk()) {
+      ClearBackgroundInstances();
+      return status;
+    }
   }
 
   // Commit the instance update.
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -275,13 +275,8 @@ TritonModel::UpdateInstanceGroup(const inference::ModelConfig& new_model_config)
 
   // Update the scheduler if this is a sequence model.
   if (model_config.has_sequence_batching()) {
-    SequenceBatchScheduler* sched = (SequenceBatchScheduler*)scheduler_.get();
-    if (sched == nullptr) {
-      ClearBackgroundInstances();
-      return Status(
-          Status::Code::INTERNAL,
-          "Unable to downcast from Scheduler to SequenceBatchScheduler");
-    }
+    SequenceBatchScheduler* sched =
+        reinterpret_cast<SequenceBatchScheduler*>(scheduler_.get());
     Status status = sched->Update();
     if (!status.IsOk()) {
       ClearBackgroundInstances();
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -633,6 +633,9 @@ TritonModel::~TritonModel()
   // Clear library handles.
   ClearHandles();
 
+  // Explicitly delete/finalize the scheduler before the model instances.
+  scheduler_.reset(nullptr);
+
   // Explicitly delete/finalize all model instances before finalizing
   // the model itself.
   instances_.clear();
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -241,10 +241,11 @@ TritonModel::Create(
   }
 
   // Create or update the model instances for this model.
-  RETURN_IF_ERROR(TritonModelInstance::SetInstances(
-      raw_local_model, backend_cmdline_config_map, host_policy_map,
-      model_config));
-  RETURN_IF_ERROR(local_model->SetConfiguredScheduler());
+  std::vector<std::shared_ptr<TritonModelInstance>> added_instances,
+      removed_instances;
+  RETURN_IF_ERROR(local_model->PrepareInstances(
+      model_config, &added_instances, &removed_instances));
+  RETURN_IF_ERROR(local_model->SetConfiguredScheduler(added_instances));
   local_model->CommitInstances();
 
   *model = std::move(local_model);
@@ -266,22 +267,20 @@ TritonModel::UpdateInstanceGroup(const inference::ModelConfig& new_model_config)
   RETURN_IF_ERROR(ValidateInstanceGroup(model_config, min_compute_capability_));
 
   // Prepare the new instances on the new config.
-  Status status = TritonModelInstance::SetInstances(
-      this, backend_cmdline_config_map_, host_policy_map_, model_config);
+  std::vector<std::shared_ptr<TritonModelInstance>> added_instances,
+      removed_instances;
+  Status status =
+      PrepareInstances(model_config, &added_instances, &removed_instances);
   if (!status.IsOk()) {
     ClearBackgroundInstances();
     return status;
   }
 
-  // Update the scheduler if this is a sequence model.
-  if (model_config.has_sequence_batching()) {
-    SequenceBatchScheduler* sched =
-        reinterpret_cast<SequenceBatchScheduler*>(scheduler_.get());
-    Status status = sched->Update();
-    if (!status.IsOk()) {
-      ClearBackgroundInstances();
-      return status;
-    }
+  // Update the scheduler.
+  status = UpdateConfiguredScheduler(added_instances, removed_instances);
+  if (!status.IsOk()) {
+    ClearBackgroundInstances();
+    return status;
   }
 
   // Commit the instance update.
@@ -394,13 +393,108 @@ TritonModel::IndexInstances() const
 }
 
 Status
-TritonModel::RegisterInstance(
-    std::shared_ptr<TritonModelInstance>&& instance, const bool passive)
+TritonModel::PrepareInstances(
+    const inference::ModelConfig& model_config,
+    std::vector<std::shared_ptr<TritonModelInstance>>* added_instances,
+    std::vector<std::shared_ptr<TritonModelInstance>>* removed_instances)
 {
-  if (passive) {
-    bg_passive_instances_.emplace_back(std::move(instance));
-  } else {
-    bg_instances_.emplace_back(std::move(instance));
+  added_instances->clear();
+  removed_instances->clear();
+
+  std::unordered_map<
+      TritonModelInstance::Signature,
+      std::vector<std::shared_ptr<TritonModelInstance>>>
+      existing_instances = IndexInstances();
+
+  for (const auto& group : model_config.instance_group()) {
+    std::vector<std::string> profile_names;
+    for (const auto& profile_name : group.profile()) {
+      profile_names.push_back(profile_name);
+    }
+    std::vector<TritonModelInstance::SecondaryDevice> secondary_devices;
+    for (const auto& secondary_device : group.secondary_devices()) {
+      secondary_devices.emplace_back(
+          inference::
+              ModelInstanceGroup_SecondaryDevice_SecondaryDeviceKind_Name(
+                  secondary_device.kind()),
+          secondary_device.device_id());
+    }
+    for (int32_t c = 0; c < group.count(); ++c) {
+      std::string instance_name{group.name() + "_" + std::to_string(c)};
+      const bool passive = group.passive();
+      std::vector<std::tuple<
+          std::string, TRITONSERVER_InstanceGroupKind, int32_t,
+          const inference::ModelRateLimiter*>>
+          instance_settings;
+      if (group.kind() == inference::ModelInstanceGroup::KIND_CPU) {
+        instance_settings.emplace_back(
+            group.host_policy().empty() ? "cpu" : group.host_policy(),
+            TRITONSERVER_INSTANCEGROUPKIND_CPU, 0 /* device_id */,
+            &group.rate_limiter());
+      } else if (group.kind() == inference::ModelInstanceGroup::KIND_GPU) {
+        for (const int32_t device_id : group.gpus()) {
+          instance_settings.emplace_back(
+              group.host_policy().empty() ? ("gpu_" + std::to_string(device_id))
+                                          : group.host_policy(),
+              TRITONSERVER_INSTANCEGROUPKIND_GPU, device_id,
+              &group.rate_limiter());
+        }
+      } else if (group.kind() == inference::ModelInstanceGroup::KIND_MODEL) {
+        instance_settings.emplace_back(
+            group.host_policy().empty() ? "model" : group.host_policy(),
+            TRITONSERVER_INSTANCEGROUPKIND_MODEL, 0 /* device_id */,
+            &group.rate_limiter());
+      } else {
+        return Status(
+            Status::Code::INVALID_ARG,
+            std::string("instance_group kind ") +
+                ModelInstanceGroup_Kind_Name(group.kind()) + " not supported");
+      }
+      for (const auto& instance_setting : instance_settings) {
+        const std::string& policy_name = std::get<0>(instance_setting);
+        const auto& kind = std::get<1>(instance_setting);
+        const auto& device_id = std::get<2>(instance_setting);
+        const inference::ModelRateLimiter* rate_limiter_config =
+            std::get<3>(instance_setting);
+
+        // All the information for the reqested instance is ready. Create a
+        // signature that identifies the requested instance.
+        const TritonModelInstance::Signature signature(group, device_id);
+
+        // Check if the requested instance can reuse an existing instance.
+        if (!TritonModelInstance::ShareBackendThread(DeviceBlocking(), kind)) {
+          auto itr = existing_instances.find(signature);
+          if (itr != existing_instances.end() && !itr->second.empty()) {
+            auto existing_instance = itr->second.back();
+            itr->second.pop_back();
+            LOG_VERBOSE(2) << "Re-using model instance named '"
+                           << existing_instance->Name() << "' with device id '"
+                           << existing_instance->DeviceId() << "'";
+            RegisterBackgroundInstance(std::move(existing_instance), passive);
+
+            continue;
+          }
+        }
+
+        // The requested instance did not match an existing instance. Create a
+        // new instance.
+        std::shared_ptr<TritonModelInstance> new_instance;
+        LOG_VERBOSE(2) << "Creating model instance named '" << instance_name
+                       << "' with device id '" << device_id << "'";
+        RETURN_IF_ERROR(TritonModelInstance::CreateInstance(
+            this, instance_name, signature, kind, device_id, profile_names,
+            passive, policy_name, *rate_limiter_config, secondary_devices,
+            &new_instance));
+        added_instances->push_back(new_instance);
+        RegisterBackgroundInstance(std::move(new_instance), passive);
+      }
+    }
+  }
+
+  // Any existing instances not reused will be removed.
+  for (auto pair : existing_instances) {
+    removed_instances->insert(
+        removed_instances->end(), pair.second.begin(), pair.second.end());
   }
 
   return Status::Success;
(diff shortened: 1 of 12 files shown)


src/dynamic_batch_scheduler.cc
```diff
@@ -170,9 +170,28 @@ DynamicBatchScheduler::~DynamicBatchScheduler()
...
+}
+
 Status
 DynamicBatchScheduler::Enqueue(std::unique_ptr<InferenceRequest>& request)
 {
+  // Pause if the scheduler is updating.
+  {
+    std::lock_guard<std::mutex> update_lock(update_mu_);
+  }
   if (stop_) {
```
nnshah1, line 195: If I understand correctly - the lock only gurantees that an enqueue won't start while an update is in progress - but doesn't gurantee that an update won't happen during an enqueue - is that expected?

kthui: Yes, it is expected, because the request(s) will be blocked by the rate limiter once the update progresses into removing the model instance, and the request can be consumed by any other instance(s) on the rate limiter after the update completes. The user [cannot specify an instance group of size zero](https://github.com/triton-inference-server/core/blob/main/src/model_config_utils.cc#L1537), so there will always be one or more instances after the update. The lock is only to block requests from rushing in during an update.

kthui: Second thought, holding the lock on `Enqueue()` is not needed for dynamic batch scheduler, since the rate limiter is capable of handling everything, so [Remove lock on dynamic scheduler enqueue](https://github.com/triton-inference-server/core/pull/207/commits/0d426026941578a1f44d6bf09da9e5fa4957a786).

I also need to [Add gap for file timestamp update](https://github.com/triton-inference-server/server/pull/5831/commits/11db1d24af8199b78ea6a0b65c83d29b4015537a), because the code likely ran a bit faster that the 'config.pbtxt' timestamp did not get a chance to update after its content is changed, which the server thinks the model is not updated and resulted in that particular test case to fail.

src/model_lifecycle.cc
```diff
@@ -650,9 +646,12 @@ ModelLifeCycle::UpdateModelConfig(
     return;
   }
 
-  // Update model instance group.
-  Status status =
-      model->UpdateInstanceGroup(new_model_config, &model_info_lock);
+  // Update model instance group. It is the responsibility of the model to
```
tanmayv25, line 649: You should describe why is it okay to release `model_info->mtx_` when calling "UpdateInstanceGroup" with this function's perspective. Why was it held in the first place and why was it released...

src/scheduler.h
```diff
@@ -72,6 +73,14 @@ class Scheduler {
   // Return the number of in-flight inferences tracked by the scheduler.
   virtual size_t InflightInferenceCount() = 0;
 
+  // Update the scheduler during a model update (i.e. instances). This function
```
tanmayv25, line 76: > This function should block any new inference requests into 'Enqueue()' when ready (i.e. sequence models)

The comment seems ambiguous. Can you make it more accurate?

src/sequence_batch_scheduler.cc
```diff
@@ -59,7 +59,8 @@ SequenceBatchScheduler::Create(
     const std::unordered_map<std::string, bool>& enforce_equal_shape_tensors,
     std::unique_ptr<Scheduler>* scheduler)
 {
-  std::unique_ptr<SequenceBatchScheduler> sched(new SequenceBatchScheduler());
+  std::unique_ptr<SequenceBatchScheduler> sched(
+      new SequenceBatchScheduler(model, enforce_equal_shape_tensors));
```
tanmayv25, line 63: This appears concerning... Do you know why SequenceBatchScheduler constructor was not forwarded "model" and "enforce_equal_shape_tensors" parameters before?

src/sequence_batch_scheduler.cc
```diff
@@ -109,12 +110,73 @@ SequenceBatchScheduler::Create(
...
+
+  // Wait until all in-flight and backlog sequence are queued.
+  update_enqueued_cv_.wait(lk, [this] {
+    return sequence_to_batcherseqslot_map_.empty() && backlog_queues_.empty() &&
+           sequence_to_backlog_map_.empty();
+  });
+
+  // Flush all pending requests to the rate limiter.
+  lk.unlock();
+  batchers_.clear();
```
tanmayv25, line 156: Instead of deleting and creating the batcher objects again during the update, shouldn't we handle these case by case basis?
case 1: Increment in instance count: No need to pause/clear out the request. Just prepare the new batchers for the new instance count.
case 2: Decrement in the instance count: Wait only for the requests in sequences allocated to batchers(for the delta instances) to complete before destructing the instances.

In the current workflow, the scheduler is still doing much of the heavy-lifting with each model instance update.

src/sequence_batch_scheduler.cc
kthui: Are they going to be updated across different sequence batch scheduler creation in the past? I think they are not going to be changed across model instance update, unless the model is destructed. Could you elaborate more on why it is concerning?

src/backend_model.cc
```diff
@@ -268,22 +265,17 @@ TritonModel::UpdateInstanceGroup(
...
-  // inference request. It is safe to move forward and commit the change.
-  config_.mutable_instance_group()->Swap(model_config.mutable_instance_group());
+  // Prepare the new instances on the new config.
+  RETURN_IF_ERROR(TritonModelInstance::SetInstances(
+      this, backend_cmdline_config_map_, host_policy_map_, model_config));
+
+  // At this point, the new model config and instances are ready but not yet
+  // written into this object. Update the scheduler and retrieve the lock
+  // holding off new inference requests, and then commit the changes.
+  std::unique_ptr<std::lock_guard<std::mutex>> lock;
```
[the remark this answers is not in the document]
kthui: This is a good point, the remaining steps after updating the scheduler actually does not require any lock. Updated: [Remove lock returned from scheduler update](https://github.com/triton-inference-server/core/pull/207/commits/092a2581d221f8c78d1019a092682a1ada29adce)

src/model_lifecycle.cc
kthui: Yes. Updated: [Remove lock returned from scheduler update](https://github.com/triton-inference-server/core/pull/207/commits/092a2581d221f8c78d1019a092682a1ada29adce)

src/scheduler.h
kthui: Yes. Updated: [Remove lock returned from scheduler update](https://github.com/triton-inference-server/core/pull/207/commits/092a2581d221f8c78d1019a092682a1ada29adce)

src/sequence_batch_scheduler.cc
kthui: Synced offline: there is no conflict for the instance update case.

src/backend_model.cc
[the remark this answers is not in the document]
kthui: Synced offline: the lock is not needed.

src/sequence_batch_scheduler.cc
kthui: Yes. This is implemented in the latest refresh.

src/backend_model.cc
```diff
@@ -268,25 +265,33 @@ TritonModel::UpdateInstanceGroup(
...
-  // At this point, the new model config is ready but not yet written into this
-  // object. The 'caller_lock' is held, so 'model_lifecycle' will pause any new
-  // inference request. It is safe to move forward and commit the change.
-  config_.mutable_instance_group()->Swap(model_config.mutable_instance_group());
-  RETURN_IF_ERROR(CommitInstances());
-  // Only model owned dynamic batch scheduler can be updated currently, so there
-  // is no need to update the scheduler.
+  // Update the scheduler if this is a sequence model.
+  if (model_config.has_sequence_batching()) {
+    SequenceBatchScheduler* sched = (SequenceBatchScheduler*)scheduler_.get();
```
tanmayv25, line 278: Why not use reinterpret_cast?

src/backend_model.cc
```diff
@@ -268,25 +265,33 @@ TritonModel::UpdateInstanceGroup(
...
+    if (sched == nullptr) {
+      ClearBackgroundInstances();
+      return Status(
+          Status::Code::INTERNAL,
+          "Unable to downcast from Scheduler to SequenceBatchScheduler");
```
tanmayv25, line 283: Isn't the code segment triggered when sched is nullptr? Can this error message be made more explicit? Like when are we expecting the sched to be nullptr. 
Is this ever a possibility even?

src/sequence_batch_scheduler.cc
```diff
@@ -1586,12 +1742,12 @@ DirectSequenceBatch::BatcherThread(const int nice)
   }  // end runner loop
 
   LOG_VERBOSE(1) << "Stopping Direct sequence-batch scheduler thread "
-                 << batcher_idx_ << "...";
+                 << (size_t)model_instance_ << "...";
```
tanmayv25, line 1745: Why not model_instance_->Name()? Like everywhere above?

tanmayv25:
Some couple of minor changes. I will defer to @nnshah1  and @GuanLuo for final approvals.

src/sequence_batch_scheduler.cc
kthui: It is because the direct batcher thread is terminated after the model instances are destructed, which accessing the `Name()` could result in segmentation violation.
```
I0628 21:25:41.467416 38923 python_be.cc:2032] TRITONBACKEND_ModelFinalize: delete model state
I0628 21:25:41.467520 38923 sequence_batch_scheduler.cc:1105] Stopping sequence-batch reaper thread...
I0628 21:25:41.467596 38923 sequence_batch_scheduler.cc:1741] Stopping Direct sequence-batch scheduler thread <A1><9B>^Xb<A6>^?^@^@it_del_1...
I0628 21:25:41.467685 38923 sequence_batch_scheduler.cc:1741] Stopping Direct sequence-batch scheduler thread <FA><D9>.r<A4>^?^@^@it_del_0<A0>  ...
I0628 21:25:41.467734 38923 model_lifecycle.cc:602] OnDestroy callback() 'model_init_del' version 1
I0628 21:25:41.467746 38923 model_lifecycle.cc:604] successfully unloaded 'model_init_del' version 1
```

It looks like destructing the scheduler before any model instances solved the issue:
```
I0628 21:37:06.530392 42475 sequence_batch_scheduler.cc:1741] Stopping Direct sequence-batch scheduler thread model_init_del_0_2...
I0628 21:37:06.530561 42475 sequence_batch_scheduler.cc:1741] Stopping Direct sequence-batch scheduler thread model_init_del_0_1...
I0628 21:37:06.530638 42475 backend_model_instance.cc:828] Stopping backend thread for model_init_del_0_1...
I0628 21:37:06.530874 42475 python_be.cc:2193] TRITONBACKEND_ModelInstanceFinalize: delete instance state
I0628 21:37:07.637947 42475 backend_model_instance.cc:828] Stopping backend thread for model_init_del_0_2...
I0628 21:37:07.638081 42475 python_be.cc:2193] TRITONBACKEND_ModelInstanceFinalize: delete instance state
```
It is updated on this commit: [Destruct scheduler before instances on model destructor](https://github.com/triton-inference-server/core/pull/207/commits/10a0b7fd3fac8e5a1154b57625d2dfb66c72f312)

I wonder you could think of anything that might go wrong with the model destruction order change.

src/backend_model.cc
kthui: Yes, reinterpret_cast seems more suitable for the pointer type change. Updated both downcast on the model and scheduler object: [Use reinterpret_cast for pointer type conversion](https://github.com/triton-inference-server/core/pull/207/commits/5adf085fefc9eab8a05bf91b908c842194354182)

src/backend_model.cc
kthui: I read somewhere that a failed cast will return a null pointer, but reinterpret_cast does not. The check is removed: [Use reinterpret_cast for pointer type conversion](https://github.com/triton-inference-server/core/pull/207/commits/5adf085fefc9eab8a05bf91b908c842194354182)

src/sequence_batch_scheduler.cc
tanmayv25: We can avoid explicit thread destruction by storing the Name() value when model_instance is defined.

kthui: I think the explicit destruction order is actually correct, because the destruction is changed from "[TritonModelInstance](https://github.com/triton-inference-server/core/blob/main/src/backend_model.cc#L634-L637) -> Scheduler -> [TritonModel](https://github.com/triton-inference-server/core/blob/main/src/backend_model.cc#L616)" to "[Scheduler](https://github.com/triton-inference-server/core/pull/207/commits/10a0b7fd3fac8e5a1154b57625d2dfb66c72f312#diff-1868cdcb6655ec791faf4b245a3ce92e4cd15983558118450208982edcc5fb5aR638) -> [TritonModelInstance](https://github.com/triton-inference-server/core/pull/207/commits/10a0b7fd3fac8e5a1154b57625d2dfb66c72f312#diff-1868cdcb6655ec791faf4b245a3ce92e4cd15983558118450208982edcc5fb5aR642-R644) -> [TritonModel](https://github.com/triton-inference-server/core/pull/207/commits/10a0b7fd3fac8e5a1154b57625d2dfb66c72f312#diff-1868cdcb6655ec791faf4b245a3ce92e4cd15983558118450208982edcc5fb5aR621)", which is exactly the opposite order of which they are constructed "[TritonModel](https://github.com/triton-inference-server/core/blob/main/src/backend_model.cc#L165-L168) -> [TritonModelInstance](https://github.com/triton-inference-server/core/blob/main/src/backend_model.cc#L244-L247) -> [Scheduler](https://github.com/triton-inference-server/core/blob/main/src/backend_model.cc#L249)".

We can also store the `Name()` value if you think updating the destruction order is too risky.

@GuanLuo @nnshah1 What do you think?

src/sequence_batch_scheduler.cc
```diff
@@ -110,12 +108,143 @@ SequenceBatchScheduler::Create(
...
+      queue_request_cnts_.erase(instance);
+    }
+  }
+
+  return Status::Success;
+}
+
+void
+SequenceBatchScheduler::InstancesDiff(
+    std::vector<std::shared_ptr<TritonModelInstance>>* added_instances,
```
nnshah1, line 198: could use references here?

kthui: Yes, but I think @GuanLuo mentioned to me that we want to use pointers for parameters that will be modified by the function (i.e. [FindModelIdentifier](https://github.com/triton-inference-server/core/blob/main/src/model_repository_manager.h#L469-L470)). Unless this does not apply on the model and scheduler part of the code.

nnshah1: I would just keep consistent with other functions - so sounds good to me. We can reconsider later.

src/sequence_batch_scheduler.cc
tanmayv25: The explicit destruction was removed with some design clean-ups if I remember correctly. I don't see any risks involved.

src/sequence_batch_scheduler.cc
```diff
@@ -110,12 +108,143 @@ SequenceBatchScheduler::Create(
...
+  return Status::Success;
+}
+
+Status
+SequenceBatchScheduler::Update()
+{
+  // Find added and removed instances for this update.
+  std::vector<std::shared_ptr<TritonModelInstance>> added_instances;
+  std::unordered_set<TritonModelInstance*> removed_instances;
+  InstancesDiff(&added_instances, &removed_instances);
```
GuanLuo, line 139: Why the scheduler needs to "re-"discover this information when model which invokes `Update()` already knows what is the changes in the instances

src/sequence_batch_scheduler.cc
```diff
@@ -110,12 +108,143 @@ SequenceBatchScheduler::Create(
...
+  {
+    std::unique_lock<std::mutex> lk(mu_);
+
+    // Add the new batchers.
+    RETURN_IF_ERROR(CreateBatchers(added_instances));
+
+    // All sequence slots of the removed instances are pending to be removed.
+    pending_removal_seq_slots_.clear();
+    for (auto& instance : removed_instances) {
+      pending_removal_seq_slots_.emplace(instance, seq_slot_cnt_);
```
GuanLuo, line 153: Should use the count captured in the corresponding SequenceBatch object. It is probably not the case here, but note that `seq_slot_cnt_` is not proper indicator of the slots associated to the given instance, which can be dangerous if `seq_slot_cnt_` may get updated after the SequenceBatch is constructed.

src/sequence_batch_scheduler.cc
```diff
@@ -110,12 +108,143 @@ SequenceBatchScheduler::Create(
...
+        // The ready slot is removed, so update pending removal instances.
+        ErasePendingRemovalSequenceSlot(ready_batcher_seq_slot);
+      }
+      ready_batcher_seq_slots_.pop();
+    }
+    ready_batcher_seq_slots_.swap(new_ready_batcher_seq_slots);
+
+    // Wait until all removed batchers have completed its assigned sequence.
+    pending_removal_seq_slots_cv_.wait(
+        lk, [this] { return pending_removal_seq_slots_.empty(); });
```
GuanLuo, line 177: So we lose a model loading thread here waiting for sequences to be finished?

src/sequence_batch_scheduler.cc
```diff
@@ -1215,10 +1344,34 @@ DirectSequenceBatch::DirectSequenceBatch(
 
 DirectSequenceBatch::~DirectSequenceBatch()
 {
-  // Signal the scheduler thread to exit...
-  {
-    std::unique_lock<std::mutex> lock(mu_);
-    scheduler_thread_exit_ = true;
+  // Wait until all pending requests are completed.
+  bool waiting_for_requests = true;
+  while (waiting_for_requests) {
```
GuanLuo, line 1349: Why this is needed?

src/backend_model.cc
```diff
@@ -381,7 +381,8 @@ TritonModel::IndexInstances() const
       std::vector<std::shared_ptr<TritonModelInstance>>>
       mapped_instances;
   for (auto* instances : {&instances_, &passive_instances_}) {
-    for (auto& instance : (*instances)) {
+    for (auto ritr = instances->rbegin(); ritr != instances->rend(); ++ritr) {
```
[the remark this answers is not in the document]
kthui: This is not the correct fix, it will be updated on DLIS-5084.

src/sequence_batch_scheduler.cc
kthui: Yes. The model loading thread will be paused, waiting for sequences belonging to the instances that will be removed to finish.

src/sequence_batch_scheduler.cc
kthui: This is to ensure the `queues_`, which holds any pending inference requests, is empty before stopping the direct scheduler thread.

Added a new cv that notifies for checking the queue during destruction, instead of a loop: [Improve direct batcher destruction logic](https://github.com/triton-inference-server/core/pull/207/commits/ee04d8d3f7be2563dca0c02d10e0143a72226968)

src/sequence_batch_scheduler.cc
kthui: This is a good point. Updated: [Use slot count when batcher is created during removal](https://github.com/triton-inference-server/core/pull/207/commits/bde56a383342f6cb57d00e03832f8bfd96575f2e)

src/sequence_batch_scheduler.cc
kthui: "re-"discover is not needed. Updated: [Refactor model reuse code](https://github.com/triton-inference-server/core/pull/207/commits/38ac91008a5d46ba63ca4e4e5ca795d87fd9c723)

src/backend_model.cc
```diff
@@ -268,25 +265,28 @@ TritonModel::UpdateInstanceGroup(
...
 
-  // At this point, the new model config is ready but not yet written into this
-  // object. The 'caller_lock' is held, so 'model_lifecycle' will pause any new
-  // inference request. It is safe to move forward and commit the change.
-  config_.mutable_instance_group()->Swap(model_config.mutable_instance_group());
-  RETURN_IF_ERROR(CommitInstances());
-  // Only model owned dynamic batch scheduler can be updated currently, so there
-  // is no need to update the scheduler.
+  // Update the scheduler if this is a sequence model.
+  if (model_config.has_sequence_batching()) {
```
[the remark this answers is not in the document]
kthui: Yes. It is updated on [Use dynamic cast for model downcast on model lifecycle](https://github.com/triton-inference-server/core/pull/207/commits/2c272148b0ee1664adc33e2b9aa01ea5bc966a27) and [Refactor model reuse code](https://github.com/triton-inference-server/core/pull/207/commits/38ac91008a5d46ba63ca4e4e5ca795d87fd9c723) commits.

src/backend_model.cc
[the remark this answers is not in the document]
GuanLuo: If it is not addressing anything in this PR, maybe you can just revert the change.

src/backend_model.cc
```diff
@@ -394,27 +394,139 @@ TritonModel::IndexInstances() const
...
-  } else {
-    bg_instances_.emplace_back(std::move(instance));
+  added_instances->clear();
+  removed_instances->clear();
+
+  std::unordered_map<
+      TritonModelInstance::Signature,
+      std::vector<std::shared_ptr<TritonModelInstance>>>
+      existing_instances = IndexInstances();
+
```
GuanLuo, line 409: Maybe add comment to summarize how the code below operates..

src/backend_model.cc
```diff
@@ -539,6 +652,27 @@ TritonModel::SetConfiguredScheduler()
...
+    if (sched == nullptr) {
+      return Status(
+          Status::Code::INTERNAL,
+          "Unable to downcast from 'Scheduler' to 'SequenceBatchScheduler' "
+          "during scheduler update");
+    }
+    return sched->Update(added_instances, removed_instances);
+  }
+
+  // Non-sequence scheduler does not need to be updated.
```
GuanLuo, line 672: Add more description on why, something like instances are encapsulated in rate limiter and other scheduler doesn't need to be aware of the change as it only interact with the rate limiter.

src/backend_model_instance.h
```diff
@@ -84,11 +84,14 @@ class TritonModelInstance {
...
-      const triton::common::HostPolicyCmdlineConfigMap& host_policy_map,
-      const inference::ModelConfig& model_config);
+  static Status CreateInstance(
+      TritonModel* model, const std::string& name, const Signature& signature,
+      TRITONSERVER_InstanceGroupKind kind, int32_t device_id,
+      const std::vector<std::string>& profile_names, const bool passive,
+      const std::string& host_policy_name,
+      const inference::ModelRateLimiter& rate_limiter_config,
+      const std::vector<SecondaryDevice>& secondary_devices,
+      std::shared_ptr<TritonModelInstance>* triton_model_instance);
```
GuanLuo, line 94: There is already a `CreateInstance()` static function. Is this a duplication?

GuanLuo: If not, please avoid confusing naming.

src/sequence_batch_scheduler.cc
GuanLuo: So if there are 4 sequence model to be updated and each has 1 active sequences there is a client keep sending requests to keep all the sequences alive. We can't perform any model update as we only have 4 model loading threads by default?

If so, can this be something passive that the scheduler itself will remove the instance when it is done with ongoing sequences? `Update()` here will just mark the instances to be removed if it can't be done immediately?

src/backend_model.cc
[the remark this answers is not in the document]
kthui: Dropped the commit: ab770ee Reuse instances in order

src/sequence_batch_scheduler.cc
kthui: > if there are 4 sequence model to be updated and each has 1 active sequences there is a client keep sending requests to keep all the sequences alive. We can't perform any model update as we only have 4 model loading threads by default?

Yes, but I think the user may specify more model loading threads in such case.

> can this be something passive that the scheduler itself will remove the instance when it is done with ongoing sequences? Update() here will just mark the instances to be removed if it can't be done immediately?

I think it is doable, because the shared ptr of the removed instances can be held by the update function and passed to a background thread.

A point worth noting is test cases involving any sequence instance removal will need to poll for the final removal count, as instances can be removing after the update api returns.

src/backend_model.cc
kthui: Yes. Updated: [Add more descriptions and improve function naming](https://github.com/triton-inference-server/core/pull/207/commits/686ebc95701d383515300cbf784fbba546cb9aef)

src/backend_model.cc
kthui: Yes. Updated: [Add more descriptions and improve function naming](https://github.com/triton-inference-server/core/pull/207/commits/686ebc95701d383515300cbf784fbba546cb9aef)

src/backend_model_instance.h
kthui: Yes. Updated: [Add more descriptions and improve function naming](https://github.com/triton-inference-server/core/pull/207/commits/686ebc95701d383515300cbf784fbba546cb9aef)

src/sequence_batch_scheduler.cc
kthui: The implementation is added, see [Remove batcher pending sequence completion after update return](https://github.com/triton-inference-server/core/pull/207/commits/93f46109d1732e354b568780de0eb475a51d37ab)
, [[WIP] Async destruction of batchers](https://github.com/triton-inference-server/core/pull/207/commits/5524945f4e48b01e52384c0a7371acbc65ef91b6)
 and [Async destruction of resources](https://github.com/triton-inference-server/core/pull/207/commits/00d0ef13e8a10dbb599e4c8f573d2422a3e6c96b)

src/backend_model.cc
```diff
@@ -394,27 +393,140 @@ TritonModel::IndexInstances() const
...
+                  secondary_device.kind()),
+          secondary_device.device_id());
+    }
+    for (int32_t c = 0; c < group.count(); ++c) {
+      std::string instance_name{group.name() + "_" + std::to_string(c)};
+      const bool passive = group.passive();
+      std::vector<std::tuple<
+          std::string, TRITONSERVER_InstanceGroupKind, int32_t,
+          const inference::ModelRateLimiter*>>
+          instance_settings;
```
rmccorm4, line 431: How about a simple struct instead of tuple for readability and less usage of `x = std::get<0>, y = std::get<1>, ...`?
```cpp
struct InstanceSetting {
  // some constructor to use with emplace_back call
  std::string policy_name;
  TRITONSERVER_InstanceGroupKind kind;
  int32_t device_id;
  const inference::ModelRateLimiter* rate_limiter_config;
};

std::vector<InstanceSetting> instance_settings;
// ...

for (const auto& setting : instance_settings) {
    const TritonModelInstance::Signature signature(setting.group, setting.device_id);
    ...
}
```

src/backend_model_instance.cc
```diff
@@ -193,135 +186,62 @@ TritonModelInstance::~TritonModelInstance()
...
-      Signature, std::vector<std::shared_ptr<TritonModelInstance>>>
-      existing_instances = model->IndexInstances();
-
   static triton::common::HostPolicyCmdlineConfig empty_host_policy;
+  const triton::common::HostPolicyCmdlineConfig* host_policy;
+  const auto policy_it = model->HostPolicyMap().find(host_policy_name);
+  if (policy_it != model->HostPolicyMap().end()) {
+    host_policy = &policy_it->second;
+  } else {
+    host_policy = &empty_host_policy;
```
rmccorm4, line 204: Can you just initialize this upfront?
```cpp
  static triton::common::HostPolicyCmdlineConfig empty_host_policy;
  const triton::common::HostPolicyCmdlineConfig* host_policy = &empty_host_policy;
  const auto policy_it = model->HostPolicyMap().find(host_policy_name);
  if (policy_it != model->HostPolicyMap().end()) {    
    host_policy = &policy_it->second;
  }
```

src/backend_model_instance.cc
```diff
@@ -193,135 +186,62 @@ TritonModelInstance::~TritonModelInstance()
...
+    const size_t allow = total * memory_limit;
+    const size_t used = total - free;
+    if (used > allow) {
+      return Status(
+          Status::Code::UNAVAILABLE,
+          std::string("can not create model '") + name +
+              "': memory limit set for " +
+              TRITONSERVER_InstanceGroupKindString(kind) + " " +
+              std::to_string(device_id) +
+              " has exceeded, model loading is rejected.");
```
rmccorm4, line 236: Can you move this to a helper?
```cpp
  if (kind == TRITONSERVER_INSTANCEGROUPKIND_GPU) {
    RETURN_IF_ERROR(VerifyModelLoadGpuFraction(...));
  }

  return Status::Success;
```

src/backend_model.cc
kthui: Sure. Updated on commit: [Use struct for instance setting](https://github.com/triton-inference-server/core/pull/207/commits/7d476256d620b2f5a43eba5217f81c5cfa7ffe55)

src/backend_model_instance.cc
kthui: Sure. Updated on commit: [Adjust host policy init logic](https://github.com/triton-inference-server/core/pull/207/commits/a21e40e101eea409b3e3cb522e011acb79f7b1e4)

src/backend_model_instance.cc
kthui: Sure. Updated on commit: [Move verify model load gpu fraction to helper function](https://github.com/triton-inference-server/core/pull/207/commits/a05d4f9378bc2d8f7f69cd77c0838146f94a9e2d)

GuanLuo (commented):
LGTM, please check with @tanmayv25 @rmccorm4 for their feedback

src/backend_model.cc
```diff
@@ -268,25 +266,26 @@ TritonModel::UpdateInstanceGroup(
...
-  // At this point, the new model config is ready but not yet written into this
-  // object. The 'caller_lock' is held, so 'model_lifecycle' will pause any new
-  // inference request. It is safe to move forward and commit the change.
-  config_.mutable_instance_group()->Swap(model_config.mutable_instance_group());
-  RETURN_IF_ERROR(CommitInstances());
-  // Only model owned dynamic batch scheduler can be updated currently, so there
-  // is no need to update the scheduler.
+  // Commit the instance update.
+  CommitInstances();
+  *config_.mutable_instance_group() = model_config.instance_group();
```
rmccorm4, line 288: I noticed the order of these two operations switched, and that you went from `->Swap()` to `=`. Any implications?

src/model_lifecycle.cc
```diff
@@ -644,17 +640,23 @@ ModelLifeCycle::UpdateModelConfig(
...
-      model->UpdateInstanceGroup(new_model_config, &model_info_lock);
+  // Update model instance group. The 'model_info->mtx_' may be released while
+  // the model is updating its instance group, because no model info from this
+  // model lifecycle object is being accessed by the model during the update,
+  // and model repository manager will ensure no concurrent update on the same
+  // model may take place. Therefore, releasing the lock allows the model to be
+  // accessed which enables inference during update.
+  model_info_lock.unlock();
+  Status status = model->UpdateInstanceGroup(new_model_config);
+  model_info_lock.lock();
```
rmccorm4, line 659: I notice we're now unblocking the lock for a longer section compared to before. Before we passed the lock and only unlocked it for for the call to `SetInstances()`. Can you summarize the difference if any? Or just noticed an improvement to safely unlock for longer?

src/sequence_batch_scheduler.cc
```diff
@@ -110,12 +109,116 @@ SequenceBatchScheduler::Create(
...
   const size_t model_batch_size = std::max(1, config.max_batch_size());
-  size_t seq_slot_cnt = model_batch_size;
+  sched->seq_slot_cnt_ = model_batch_size;
   if (config.sequence_batching().has_oldest()) {
-    seq_slot_cnt =
+    sched->seq_slot_cnt_ =
         config.sequence_batching().oldest().max_candidate_sequences();
   }
 
+  // Create batchers.
```
rmccorm4, line 118:
```suggestion
  // Create a batcher for each instance.
```

src/sequence_batch_scheduler.cc
```diff
@@ -110,12 +109,116 @@ SequenceBatchScheduler::Create(
...
+  RETURN_IF_ERROR(sched->CreateBatchers(new_instances));
+
+
+  SequenceBatchScheduler* raw = sched.release();
+  // Create background threads that watch for different sequence states.
+  raw->StartBackgroundThreads();
+  scheduler->reset(raw);
```
rmccorm4, line 125: Can this be simplified to something like this?
```
  // Create background threads that watch for different sequence states.
  sched->StartBackgroundThreads();
  scheduler->reset(std::move(sched));
```

src/backend_model.cc
kthui: > the order of these two operations switched

The `CommitInstances();` use to be `RETURN_IF_ERROR(CommitInstances());`. If an error is returned and rolled-back during commit, the model config should not be changed. However, `CommitInstances();` should not (and does not) return any error, so the `RETURN_IF_ERROR()` is removed and the order is kept as it is.

I do not think swapping those two lines will make any difference, because updating the instance group is for the next round of model instance update, which cannot happen concurrently with this update, as it is safeguarded by model repository manager.

> went from `->Swap()` to `=`

I saw a segfault in the past that might be attributed to updating the model config protobuf message (swapping while accessing part of the message). From my readings, protobuf is not thread safe unless otherwise noted, and the only note I found is only [allocation methods are thread safe](https://protobuf.dev/reference/cpp/arenas/#thread-safety). Swapping is not an allocation, but assignment involves deep copy which could be regarded as an allocation. This is still more on the experiment side to see if the issue is fixed, but I have not noticed any related failures on this scheduler update branch after switching from swap to assignment.

src/model_lifecycle.cc
kthui: The lock is not necessary, because it cannot stop inference at the first place, and no concurrent model instance update can happen as it is safeguarded by model repository manager. Therefore, the only two places where locking is needed are on the rate limiter and scheduler. They both have their internal lock to safeguard any concurrent access, so the lock is not necessary.

src/sequence_batch_scheduler.cc
kthui: I think we could, but I am not sure about the initial reason for setting the thread on the raw pointer, maybe it has something to do with the `nice` setting.

@tanmayv25 Do you think we can remove the intermediate raw pointer, and set the threads on the unique pointer instead?

src/sequence_batch_scheduler.cc
```diff
@@ -110,12 +109,116 @@ SequenceBatchScheduler::Create(
...
+    return false;
+  }
+
+  LOG_VERBOSE(1) << "Removing slot for batcher "
+                 << seq_slot.model_instance_->Name() << ", slot "
+                 << seq_slot.seq_slot_;
+
+  // Subtract the number of slots, and erase the batcher if no more slots left.
+  if (--it->second.first == 0) {
+    LOG_VERBOSE(1) << "Removing batcher " << seq_slot.model_instance_->Name();
```
rmccorm4, line 188:
```suggestion
    LOG_VERBOSE(1) << "Removing batcher for instance: " << seq_slot.model_instance_->Name();
```

src/sequence_batch_scheduler.cc
```diff
@@ -110,12 +109,116 @@ SequenceBatchScheduler::Create(
...
+    // The sequence slot is not pending removal.
+    return false;
+  }
+
+  LOG_VERBOSE(1) << "Removing slot for batcher "
+                 << seq_slot.model_instance_->Name() << ", slot "
+                 << seq_slot.seq_slot_;
+
+  // Subtract the number of slots, and erase the batcher if no more slots left.
+  if (--it->second.first == 0) {
```
rmccorm4, line 187: Can you assign some names to `auto& descriptive_var = it->first`, `it->second`, and `batcher_it->second`, etc.? It's not easy to scan the following section without going back and checking what each pair consists of.

src/sequence_batch_scheduler.cc
```diff
@@ -813,6 +906,44 @@ SequenceBatchScheduler::DelayScheduler(
...
+      new std::thread([this]() { CleanUpThread(20 /* nice */); }));
+}
+
+void
+SequenceBatchScheduler::StopBackgroundThreads()
+{
+  // Exit the clean-up thread.
+  {
+    std::lock_guard<std::mutex> lock(mu_);
+    clean_up_thread_exit_ = true;
```
rmccorm4, line 929: `std::atomic<bool>` would be simpler here.

src/sequence_batch_scheduler.cc
```diff
@@ -813,6 +906,44 @@ SequenceBatchScheduler::DelayScheduler(
...
+  }
+  clean_up_cv_.notify_one();
+  if (clean_up_thread_ && clean_up_thread_->joinable()) {
+    clean_up_thread_->join();
+  }
+
+  // Exit the reaper thread.
+  {
+    std::lock_guard<std::mutex> lock(mu_);
+    reaper_thread_exit_ = true;
```
rmccorm4, line 939: `std::atomic<bool>` would be simpler here.

src/sequence_batch_scheduler.cc
```diff
@@ -976,8 +1107,49 @@ SequenceBatchScheduler::ReaperThread(const int nice)
...
+                   << "...";
+  } else {
+    LOG_VERBOSE(1) << "Starting sequence-batch clean-up thread at default nice "
+                      "(requested nice "
+                   << nice << " failed)...";
+  }
+#else
+  LOG_VERBOSE(1)
+      << "Starting sequence-batch clean-up thread at default nice...";
+#endif
```
rmccorm4, line 1125: Can this be a helper re-used in both thread functions?

rmccorm4: Looks like `BatcherThread` could reuse it too

src/sequence_batch_scheduler.cc
```diff
@@ -1215,12 +1387,39 @@ DirectSequenceBatch::DirectSequenceBatch(
...
+        return false;
+      }
+      return true;
+    });
   }
 
+  // Signal the scheduler thread to exit.
+  {
+    std::lock_guard<std::mutex> lk(mu_);
+    scheduler_thread_exit_ = true;
```
rmccorm4, line 1421: std::atomic<bool> would be simpler and safer here. These seem to be inconsistently protected.

src/sequence_batch_scheduler.cc
```diff
@@ -1573,7 +1777,7 @@ DirectSequenceBatch::BatcherThread(const int nice)
           std::unique_lock<std::mutex> lk(payload_mu_);
           exec_complete_ = true;
         }
-        payload_cv_.notify_one();
+        payload_cv_.notify_all();
```
rmccorm4, line 1780: Why the move from `notify_one` to `notify_all` based on a single payload's readiness?

rmccorm4 (commented):
Nice refactoring and additions! :rocket: 

Only minor comments.

src/backend_model.cc
rmccorm4: > I saw a segfault in the past that might be attributed to updating the model config protobuf message (swapping while accessing part of the message).

Shall we comment a note that this line may be cause for concern?

src/sequence_batch_scheduler.cc
kthui: The condition variable has two different points where it is waited on, the batcher and destructor both wait until the last enqueued payload is complete. When the batcher thread is being destructed, notify just one might only unblock the batcher thread, but not the destructor, so notify all will ensure everything unblocks.

src/sequence_batch_scheduler.cc
tanmayv25: @kthui  can you elaborate upon the "it has something to do with the nice setting."?

tanmayv25: I think unique_ptr is only a wrapper around the raw pointer. As per my understanding we can make the proposed simplification.

kthui: Updated: [Simplify steps before starting background threads](https://github.com/triton-inference-server/core/pull/207/commits/6cdd4eb7b178bcd7803b1506d64668c5de37b855)

src/sequence_batch_scheduler.cc
kthui: Sure. Updated on commit: [Improve iterator variable naming](https://github.com/triton-inference-server/core/pull/207/commits/f3338a9873683baf899b48b051bc0d37f86add37)

src/sequence_batch_scheduler.cc
kthui: Yes, updated on commit: [Use atomic bool for thread exit vars](https://github.com/triton-inference-server/core/pull/207/commits/d875d79519ccdd3d0877161dea58ea4a02b503ea)

src/sequence_batch_scheduler.cc
kthui: Yes, updated on commit: [Use atomic bool for thread exit vars](https://github.com/triton-inference-server/core/pull/207/commits/d875d79519ccdd3d0877161dea58ea4a02b503ea)

src/sequence_batch_scheduler.cc
kthui: Yes, updated on commit: [Use atomic bool for thread exit vars](https://github.com/triton-inference-server/core/pull/207/commits/d875d79519ccdd3d0877161dea58ea4a02b503ea)

src/sequence_batch_scheduler.cc
kthui: Nice spotting! Updated on commit: [Group set thread priority into a function](https://github.com/triton-inference-server/core/pull/207/commits/db77e3759b34df84bcc9239784995a33ac81f4c3)

src/sequence_batch_scheduler.cc
```diff
@@ -184,23 +200,26 @@ SequenceBatchScheduler::EraseBatcherSequenceSlot(
...
     // Erase batcher.
+    auto batcher_it = batchers_.find(seq_slot.model_instance_);
+    auto& pending_removal_batcher = batcher_it->second;
     std::unique_ptr<SequenceBatch> removed_batcher;
-    auto batcher_it = batchers_.find(it->first);
-    batcher_it->second.swap(removed_batcher);
-    batchers_.erase(std::move(batcher_it));
+    removed_batcher.swap(pending_removal_batcher);
     removed_batchers_.emplace_back(std::move(removed_batcher));
+    batchers_.erase(std::move(batcher_it));
```
rmccorm4, line 213: I don't think the `std::move` needed here, is it?
```cpp
#include <iostream>
#include <unordered_map>

int main() {
  std::unordered_map<std::string, std::string> m;
  m["hello"] = "world";

  auto it = m.find("hello");

  m.erase(it);
  //m.erase(std::move(it));  // move unnecessary

  return 0;
}
```

src/sequence_batch_scheduler.cc
```diff
@@ -184,23 +200,26 @@ SequenceBatchScheduler::EraseBatcherSequenceSlot(
...
-    it->second.second.swap(removed_instance);
+    removed_instance.swap(pending_removal_instance);
     removed_instances_.emplace_back(std::move(removed_instance));
 
     // Erase from debugging/testing info.
-    queue_request_cnts_.erase(it->first);
+    queue_request_cnts_.erase(seq_slot.model_instance_);
 
     // Erase sequence slot from pending.
     pending_removal_seq_slots_.erase(std::move(it));
```
rmccorm4, line 225: same here, is `std::move` needed to just erase the iterator?

src/sequence_batch_scheduler.cc
```diff
@@ -184,23 +200,26 @@ SequenceBatchScheduler::EraseBatcherSequenceSlot(
...
     LOG_VERBOSE(1) << "Removing batcher " << seq_slot.model_instance_->Name();
 
     // Erase batcher.
+    auto batcher_it = batchers_.find(seq_slot.model_instance_);
+    auto& pending_removal_batcher = batcher_it->second;
     std::unique_ptr<SequenceBatch> removed_batcher;
-    auto batcher_it = batchers_.find(it->first);
-    batcher_it->second.swap(removed_batcher);
-    batchers_.erase(std::move(batcher_it));
+    removed_batcher.swap(pending_removal_batcher);
```
rmccorm4, line 211: Similarly, if I'm understanding this section right, do we need the intermediate swap?

Currently an intermediate "removed_batcher" that is swapped then moved:
```cpp
    auto& pending_removal_batcher = batcher_it->second;
    std::unique_ptr<SequenceBatch> removed_batcher;
    removed_batcher.swap(pending_removal_batcher);
    removed_batchers_.emplace_back(std::move(removed_batcher));
```

Can it just be something like:
```cpp
    auto& pending_removal_batcher = batcher_it->second;
    removed_batchers_.emplace_back(std::move(pending_removal_batcher));
```

src/sequence_batch_scheduler.cc
```diff
@@ -184,23 +200,26 @@ SequenceBatchScheduler::EraseBatcherSequenceSlot(
...
     removed_batchers_.emplace_back(std::move(removed_batcher));
+    batchers_.erase(std::move(batcher_it));
 
     // Stop tracking the removed instance.
+    auto& pending_removal_instance = it->second.second;
     std::shared_ptr<TritonModelInstance> removed_instance;
-    it->second.second.swap(removed_instance);
+    removed_instance.swap(pending_removal_instance);
     removed_instances_.emplace_back(std::move(removed_instance));
```
rmccorm4: Same here, do we need an intermediate swap+move? Or can you just directly move `pending_removal_instance`?
```cpp
    auto& pending_removal_instance = it->second.second;
    removed_instances_.emplace_back(std::move(pending_removal_instance));
```

src/backend_model.cc
kthui: The segfault could relate to something else than the protobuf message swapping. I think leaving a note might lead debugging into the wrong way, if that would happen again.

src/sequence_batch_scheduler.cc
kthui: >  elaborate upon the "it has something to do with the nice setting."?

This is just my example on something that could go wrong, if the original code is trying to avoid some problems by initializing the thread with the raw pointer. The CI is passing, so I think there is no problem.

src/sequence_batch_scheduler.cc
kthui: Yes, checking on the docs the `erase` function does not take `&&` parameter. Updated: [Simplify pending removal to removed steps](https://github.com/triton-inference-server/core/pull/207/commits/b6a8a635c9177e5577ec3b806118f1b7914e6e13)

src/sequence_batch_scheduler.cc
kthui: Yes, checking on the docs the `erase` function does not take `&&` parameter. Updated: [Simplify pending removal to removed steps](https://github.com/triton-inference-server/core/pull/207/commits/b6a8a635c9177e5577ec3b806118f1b7914e6e13)

src/sequence_batch_scheduler.cc
kthui: It certainly looks cleaner. I was concerned about moving the object away from the container managing it, but if no one will be accessing the object on the container, I think if is fine. Updated: [Simplify pending removal to removed steps](https://github.com/triton-inference-server/core/pull/207/commits/b6a8a635c9177e5577ec3b806118f1b7914e6e13)

src/sequence_batch_scheduler.cc
kthui: It certainly looks cleaner. I was concerned about moving the object away from the container managing it, but if no one will be accessing the object on the container, I think if is fine. Updated: [Simplify pending removal to removed steps](https://github.com/triton-inference-server/core/pull/207/commits/b6a8a635c9177e5577ec3b806118f1b7914e6e13)

rmccorm4 (approved):
LGTM, but please share a final clean CI run before merging.

PR #208, opened 2023-05-24
Avoid rate limiter resource count update from iterating over all instances

kthui:
This PR optimize the logic on rate limiter instance max resource update, after an instance is added or removed. The previous logic will iterate over all instances once for each instance added or removed. For adding an instance, the new logic will only compare the resource limit between the new instance and the max resource, and update the max resource if the new instance has a higher limit. For removing an instance, the resource on the instance being removed is compared with the max resource. If the max resource is greater than the resource on the instance being removed, then no update will be performed. If the removed instance has higher or equal limit to the max resource, then the max resource is discarded and re-computed, which iterates over the instances once.

Server PR: https://github.com/triton-inference-server/server/pull/5885

### DIFF
--- a/src/rate_limiter.cc
+++ b/src/rate_limiter.cc
@@ -78,8 +78,8 @@ RateLimiter::RegisterModelInstance(
       // Without this serialization instances of other models might fail
       // to load because of the resource constraints in this instance.
       std::lock_guard<std::mutex> lk(resource_manager_mtx_);
-      resource_manager_->AddModelInstance(pair_it.first->second.get());
-      const auto& status = resource_manager_->UpdateResourceLimits();
+      const auto& status =
+          resource_manager_->AddModelInstance(pair_it.first->second.get());
       if (!status.IsOk()) {
         resource_manager_->RemoveModelInstance(pair_it.first->second.get());
         return status;
@@ -111,23 +111,6 @@ RateLimiter::UnregisterModelInstance(TritonModelInstance* triton_model_instance)
     model_instances.erase(i_it);
   }
 
-  if (!ignore_resources_and_priority_) {
-    // The update can be skipped if there is only model instance count update,
-    // which new results will not be generated in such case. [FIXME: DLIS-4821]
-    RETURN_IF_ERROR(resource_manager_->UpdateResourceLimits());
-  }
-
-  {
-    std::lock_guard<std::mutex> lk(payload_queues_mu_);
-    auto p_it = payload_queues_.find(model);
-    if (p_it != payload_queues_.end()) {
-      auto s_it = p_it->second->specific_queues_.find(triton_model_instance);
-      if (s_it != p_it->second->specific_queues_.end()) {
-        p_it->second->specific_queues_.erase(s_it);
-      }
-    }
-  }
-
   return Status::Success;
 }
 
@@ -149,10 +132,6 @@ RateLimiter::UnregisterModel(const TritonModel* model)
 
     model_instance_ctxs_.erase(model);
     model_contexts_.erase(model);
-  }
-
-  if (!ignore_resources_and_priority_) {
-    RETURN_IF_ERROR(resource_manager_->UpdateResourceLimits());
   }
 
   {
@@ -733,13 +712,15 @@ RateLimiter::ModelInstanceContext::ScaledPriority()
 }
 
 void
-RateLimiter::ModelInstanceContext::RequestRemoval() {
+RateLimiter::ModelInstanceContext::RequestRemoval()
+{
   std::unique_lock<std::mutex> lk(state_mtx_);
   removal_in_progress_ = true;
 }
 
 bool
-RateLimiter::ModelInstanceContext::IsRemovalInProgress() {
+RateLimiter::ModelInstanceContext::IsRemovalInProgress()
+{
   std::unique_lock<std::mutex> lk(state_mtx_);
   return removal_in_progress_;
 }
@@ -760,11 +741,12 @@ RateLimiter::ResourceManager::Create(
   return Status::Success;
 }
 
-void
+Status
 RateLimiter::ResourceManager::AddModelInstance(
     const ModelInstanceContext* instance)
 {
-  std::lock_guard<std::mutex> lk(model_resources_mtx_);
+  // Add instance into model resources.
+  std::lock_guard<std::mutex> lk1(model_resources_mtx_);
   auto pr = model_resources_.emplace(std::make_pair(instance, ResourceMap()));
   for (const auto& resource : instance->GetRateLimiterConfig()->resources()) {
     if (resource.global()) {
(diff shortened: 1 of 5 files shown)


rmccorm4:
>  If the removed instance has **higher** or equal limit to the max resource

@kthui can an instance have a higher resource limit than the max resource?

src/rate_limiter.cc
```diff
@@ -775,52 +757,89 @@ RateLimiter::ResourceManager::AddModelInstance(
...
   }
+  // Check if max resources need to be updated.
+  bool update_needed = false;
+  std::lock_guard<std::mutex> lk2(max_resources_mtx_);
+  for (const auto& resource_device_map : itr->second) {
+    auto ditr = max_resources_.find(resource_device_map.first);
+    if (ditr != max_resources_.end()) {
+      for (const auto resource : resource_device_map.second) {
+        auto ritr = ditr->second.find(resource.first);
+        if (ritr != ditr->second.end() && ritr->second >= resource.second) {
```
rmccorm4, line 785: > If the removed instance has higher or equal limit to the max resource

@kthui can an instance have a higher resource limit than the max resource?

src/rate_limiter.cc
```diff
@@ -775,52 +757,89 @@ RateLimiter::ResourceManager::AddModelInstance(
...
+          update_needed = true;
+          break;
+        }
+      }
+    }
+    if (update_needed) {
```
rmccorm4, line 791: At a glance from the loops above, it looks like we need to quite a bit of looping. Does it simplify things if the `max_resources_` object also held some list of `ModelInstance*` that contributed to it? 

So upon removal you can just check if the instance is contained in that max resource list and immediately know it needs to be recomputed? Not sure if that overcomplicates things, or the rough size/impact of the loops above. Let me know what you think.

src/rate_limiter.cc
```diff
@@ -775,52 +757,89 @@ RateLimiter::ResourceManager::AddModelInstance(
...
+    const ResourceMap& instance_resource_map)
+{
+  for (const auto& resource_device_map : instance_resource_map) {
+    auto ditr = max_resources_.find(resource_device_map.first);
+    if (ditr == max_resources_.end()) {
+      ditr = max_resources_
+                 .emplace(resource_device_map.first, resource_device_map.second)
+                 .first;
+    } else {
+      for (const auto resource : resource_device_map.second) {
```
rmccorm4, line 826:
```suggestion
      for (const auto& resource : resource_device_map.second) {
```

src/rate_limiter.cc
```diff
@@ -775,52 +757,89 @@ RateLimiter::ResourceManager::AddModelInstance(
...
     return Status(
         Status::Code::INTERNAL, "Can not find the instance to remove");
   }
+  // Check if max resources need to be updated.
+  bool update_needed = false;
+  std::lock_guard<std::mutex> lk2(max_resources_mtx_);
+  for (const auto& resource_device_map : itr->second) {
+    auto ditr = max_resources_.find(resource_device_map.first);
+    if (ditr != max_resources_.end()) {
+      for (const auto resource : resource_device_map.second) {
```
rmccorm4, line 783:
```suggestion
      for (const auto& resource : resource_device_map.second) {
```

src/rate_limiter.cc
kthui: > can an instance have a higher resource limit than the max resource?

In the case when an instance is being removed, it cannot have higher resource limit than the max resource. For instance, there are 3 instances with resources `{1, 2, 3}`, so the current max resource is `3`. Removing any of the three instances cannot exceed the current max resource of `3`.

If the instance with the max resource is removed, the max resource is recomputed because the second to the max resource is unknown. For any other instances, the current max resource is still valid.

> Shouldn't the instance's value be considered in the current max?

In the case when it is removed, I think it does not hurt to leave it as the max, until some other instances with higher resource limit is added, but it is recomputed to match the previous behavior, which recompute the max resource after every instance add/remove.

rmccorm4: > In the case when it is removed, I think it does not hurt to leave it as the max
I'm not suggesting to leave it as the max when removed, I just wanted to verify my understanding that it shouldn't be possible to be higher than the max.

i.e. `ritr->second >= resource.second` may actually be more like:
```
if (ritr->second > resource.second) {
  // this shouldn't happen, bug?
} else if (ritr->second == resource.second) {
  // this instance held a max, so recompute
} else {
  // this wasn't a max, no action necessary
}
```

kthui: Yes, you are right. The `>` case should never happen, the `>=` is only as a failsafe.

src/rate_limiter.cc
kthui: I like the idea, but there is an added complexity in managing the contents in "some list of `ModelInstance*`". For example, consider adding instances 1 to 3 with cost `[1, 2, 3]` respectively. When instance 2 is being added, it will see a cost of `1` is being replaced, but not knowing which instance contributed to the cost of `1`. Of course, in this simple example, the list of `ModelInstance*` only contains instance 1, which is "known". In the real world, the max resource is a map from multiple `device_id` to multiple `resource_type_name` and its 'resource limit', which is reasonable to have multiple instances in the list of `ModelInstance*` without knowing exactly which 'resource limit' it is specifying. To address this issue, it is possible to put the pointer next to the 'resource limit', but the 'resource limit' is already sufficient at making the decision without needing the pointer.

Please let me know if I am missing something here.

src/rate_limiter.cc
```diff
@@ -78,8 +78,8 @@ RateLimiter::RegisterModelInstance(
       // Without this serialization instances of other models might fail
       // to load because of the resource constraints in this instance.
       std::lock_guard<std::mutex> lk(resource_manager_mtx_);
-      resource_manager_->AddModelInstance(pair_it.first->second.get());
-      const auto& status = resource_manager_->UpdateResourceLimits();
+      const auto& status =
+          resource_manager_->AddModelInstance(pair_it.first->second.get());
       if (!status.IsOk()) {
         resource_manager_->RemoveModelInstance(pair_it.first->second.get());
```
rmccorm4, line 84: @kthui it looks like we throw away the status returned by `RemoveModelInstance`, which in turn calls several other methods that can return detailed status errors. 

Can we propogate the error statuses back up via `LOG_ERROR` in the places where we call this method? It looks like we call it in a few places.

src/rate_limiter.cc
rmccorm4: Can we return an error status in the `>` case to catch if we ever have a bug in our resource limit logic?

src/rate_limiter.cc
kthui: Updated: [Minor fix and return error on instance remove failure](https://github.com/triton-inference-server/core/pull/208/commits/33a5369309a117e216614f91ca96df7b95cadedb)

src/rate_limiter.cc
kthui: Updated: [Minor fix and return error on instance remove failure](https://github.com/triton-inference-server/core/pull/208/commits/33a5369309a117e216614f91ca96df7b95cadedb)

src/rate_limiter.cc
kthui: Yes. Updated: [Minor fix and return error on instance remove failure](https://github.com/triton-inference-server/core/pull/208/commits/33a5369309a117e216614f91ca96df7b95cadedb)

src/rate_limiter.cc
kthui: Yes, also enhanced all other "should not print this" scattered in the rate limiter: [Enhance should not print this message](https://github.com/triton-inference-server/core/pull/208/commits/0cdb4c98b1e4b2e0155f8e16cc3284698f87a1d4)

kthui: Sorry for the confusion, my arrow is at the wrong direction 🤦: [Fix bug](https://github.com/triton-inference-server/core/pull/208/commits/ae3a3efae03e5d522a4aef2256a3a1d99fff9c61)

src/rate_limiter.cc
```diff
@@ -775,52 +773,93 @@ RateLimiter::ResourceManager::AddModelInstance(
...
+  std::lock_guard<std::mutex> lk2(max_resources_mtx_);
+  for (const auto& resource_device_map : itr->second) {
+    auto ditr = max_resources_.find(resource_device_map.first);
+    if (ditr != max_resources_.end()) {
+      for (const auto& resource : resource_device_map.second) {
+        auto ritr = ditr->second.find(resource.first);
+        if (ritr != ditr->second.end() && ritr->second <= resource.second) {
+          update_needed = true;
+          if (ritr->second < resource.second) {
+            LOG_ERROR << "Should not print this! Removing an instance with "
```
tanmayv25, line 804: Why is this case treated as an error? Isn't it quite possible and is being handled below?

src/rate_limiter.cc
```diff
@@ -105,29 +109,13 @@ RateLimiter::UnregisterModelInstance(TritonModelInstance* triton_model_instance)
...
-    RETURN_IF_ERROR(resource_manager_->UpdateResourceLimits());
-  }
-
-  {
-    std::lock_guard<std::mutex> lk(payload_queues_mu_);
-    auto p_it = payload_queues_.find(model);
-    if (p_it != payload_queues_.end()) {
-      auto s_it = p_it->second->specific_queues_.find(triton_model_instance);
-      if (s_it != p_it->second->specific_queues_.end()) {
-        p_it->second->specific_queues_.erase(s_it);
```
tanmayv25, line 126: I wonder why we are no longer removing `specific_queues`  upon unregistering the model?

src/rate_limiter.cc
kthui: > Isn't it quite possible and is being handled below?

Yes, this case [strictly less than] can be handled below.

> Why is this case treated as an error?

From the lifecycle of max resource as a whole, the max resource will always equal to the maximum count specified by one of the instance, for the particular resource identifier on the particular device. Therefore, given any instance and any given time, the resource requirement on the instance should not exceed the max resource. If this is seen [strictly less than case], it should be treated as an error.

The above is true even if explicit resource is set, because the max resource among instances is computed initially, and then the max resource is compared against explicit resource. If the explicit resource is less than the max resource, then [an error is returned](https://github.com/triton-inference-server/core/pull/208/files/eac9e0986569b776fc21cb3c28630e5a3b2e28d4#diff-bf3055556e25cd26a150f7262cf28d8e78733c10ed3f72f0c4fd10d9fe010dbbR949-R957). If the explicit resource equals to or exceeds the max resource, then the [max resource is updated to match the explicit resource](https://github.com/triton-inference-server/core/pull/208/files/eac9e0986569b776fc21cb3c28630e5a3b2e28d4#diff-bf3055556e25cd26a150f7262cf28d8e78733c10ed3f72f0c4fd10d9fe010dbbL920). From the logic, the max resource can only be increased further, and cannot be lowered. Therefore, it would be very unusual to see an instance to have resource count exceeding the max resource.

src/rate_limiter.cc
kthui: Thanks for catching. I was thinking it is removing payloads from the specific queue, but it is removing the queue for holding payloads. It is added back: [Revert specific queue erase removal](https://github.com/triton-inference-server/core/pull/208/commits/8fa3e9a3ebff49cfa206baa4712eb335e2734cd0)

src/rate_limiter.cc
tanmayv25: Thanks for the explanation!

tanmayv25 (approved):

PR #209, opened 2023-05-25
Improve runtime at finding model instances during instance update

kthui:
New test: https://github.com/triton-inference-server/server/pull/5852

### DIFF
--- a/src/backend_model_instance.h
+++ b/src/backend_model_instance.h
@@ -62,25 +62,30 @@ class TritonModelInstance {
   class Signature {
    public:
     Signature(
-        const inference::ModelInstanceGroup& group_config, int32_t device_id)
-        : group_config_(group_config), device_id_(device_id)
+        const inference::ModelInstanceGroup& instance_config, int32_t device_id)
+        : config_signature_(InstanceConfigSignature(instance_config)),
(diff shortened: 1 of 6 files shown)


src/backend_model_instance.h
```diff
@@ -62,28 +62,24 @@ class TritonModelInstance {
...
     }
     bool operator!=(const Signature& rhs) const { return !(*this == rhs); }
-    // Enable/Disable matching. If disabled, on either lhs or rhs or both, then
-    // lhs != rhs under all scenarios, including if they are equivalent.
-    // This feature is intended to filter out signatures that have already been
-    // matched, by disabling matching.
-    void EnableMatching() { can_match_ = true; }
-    void DisableMatching() { can_match_ = false; }
+    bool operator<(const Signature& rhs) const
+    {
```
nnshah1, line 75: how about this instead:

1) use an unordered map
2) define the hash for a signature as the hash of the serialized config (as string)
3) create the hash once when signature is constructed

Unordered map would have better perf - and we don't need an ordered map. creating the hash based on the serialized string (and storing it) - will be more efficient than serializing on each compare -

nnshah1 (approved):

PR #216, opened 2023-06-07
Affix S3 client to use HTTP 1.1 only

kthui:
Server PR: https://github.com/triton-inference-server/server/pull/5911

This PR prevents the s3 sdk curl library from upgrading connections to HTTP 2.

### DIFF
--- a/src/filesystem/implementations/s3.h
+++ b/src/filesystem/implementations/s3.h
@@ -61,7 +61,8 @@ class S3CurlHttpClient : public Aws::Http::CurlHttpClient {
  protected:
   void OverrideOptionsOnConnectionHandle(CURL* connectionHandle) const override
   {
-    curl_easy_setopt(connectionHandle, CURLOPT_HTTP_VERSION, 2L);  // HTTP 1.1
+    curl_easy_setopt(
+        connectionHandle, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_1_1);
   }
 };
 class S3HttpClientFactory : public Aws::Http::HttpClientFactory {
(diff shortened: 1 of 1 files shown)


nnshah1 (dismissed):

src/filesystem/implementations/s3.h
```diff
@@ -39,6 +45,50 @@ namespace triton { namespace core {
...
+  explicit S3CurlHttpClient(
+      const Aws::Client::ClientConfiguration& client_config)
+      : Aws::Http::CurlHttpClient(client_config)
+  {
+  }
+
+ protected:
+  void OverrideOptionsOnConnectionHandle(CURL* connectionHandle) const override
+  {
+    curl_easy_setopt(connectionHandle, CURLOPT_HTTP_VERSION, 2L);  // HTTP 1.1
```
Tabrizian, line 64: Do we want to force HTTP 2 or HTTP 1.1? Does 2L resolve to CURL_HTTP_VERSION_1_1? If so, maybe update it to `CURL_HTTP_VERSION_1_1` to avoid confusion.

nnshah1: @kthui if the constant isn't available - can we define it ourselves at the top - agree with @Tabrizian for readability - having a named constant here is beneficial

kthui: > Do we want to force HTTP 2 or HTTP 1.1?

We want to limit the version to http/1.1 (or http/1.0), but not http/2.0 or above.

> Does 2L resolve to CURL_HTTP_VERSION_1_1?

Yes.

> If so, maybe update it to `CURL_HTTP_VERSION_1_1` to avoid confusion.

Yes, the build is successful with `CURL_HTTP_VERSION_1_1` constant and passing the test locally.

nnshah1 (approved):

Tabrizian (approved):

PR #222, opened 2023-06-09
[WIP] Enabling nested spans for trace mode OpenTelemetry

oandreeva-nv:
This PR enables nested spans for trace mode OpenTelemetry. Major changes are:
- I moved InferenceTraceMode from server to core as TRITONSERVER_InferenceTraceMode
- Added TRITONSERVER_InferenceTraceWithModeNew, to create a trace instance with provided mode + refactored TRITONSERVER_InferenceTraceTensorNew and TRITONSERVER_InferenceTraceNew to remove repeated code. 
- TRITONSERVER_InferenceTraceSetOpenTelemetryTracer and TRITONSERVER_InferenceTraceSetOpenTelemetryContext to pass OpenTelemetry's tracer and context to core's trace instance. These 2 objects are necessary for span creation (tracer) and storing info about currently active span and time_offset (created on the server side).
- I tritonserver.h ran through `cclang`, so it has a lot of `cclang` related changes

Current logic : on the core side I create 2 spans: Request span, which has the same name as the model, this request is sent to, and compute span, which holds compute related timestamps and events and represents the amount of time compute took:

How it looks:
image
image

Note: this PR should be merged with server's PR: https://github.com/triton-inference-server/server/pull/5928


src/tritonserver.cc
```diff
@@ -50,6 +50,12 @@
 #include "triton/common/triton_json.h"
 #include "tritonserver_apis.h"
 
+#ifndef _WIN32
+#include "opentelemetry/sdk/trace/tracer.h"
+namespace otel_trace_api = opentelemetry::trace;
```
rmccorm4, line 55: @oandreeva-nv in your screenshot examples, why are they labeled "unknown_service"? Is there some tag we can put in the source to have these say "Triton" or something in the collector/UI view?

oandreeva-nv: It's due to the jaeger. I was planning to investigate if we can do something with this on our side, or if it is something, that need to be addressed on the collector's side

oandreeva-nv:
Working on the doc for discussions

oandreeva-nv:
Closing: moved this logic to server side

PR #223, opened 2023-06-09
Allow reserved values for platform field when using python backend

tanmayv25:


src/model_config_utils.cc
```diff
@@ -2247,9 +2247,34 @@ GetBackendType(const std::string& backend_name)
...
+
   return BackendType::BACKEND_TYPE_UNKNOWN;
 }
 
+bool
+ValidateBackendAndPlatformFields(const inference::ModelConfig& config)
+{
+  bool valid_pair = false;
+
+  auto backend_type = GetBackendType(config.backend());
```
nnshah1, line 2262: can we instead relax this logic to: "If GetBackendTypeFromPlaform() == UNKNOWN OR GetBackendTypeFromPlatform()==GetBackendType()"

In general it seems like we allow for the possibility that the platform is handled by the backend - so in cases where it isn't known - maybe that is up to the backend to validate.

tanmayv25: Fixed.

nnshah1 (approved):
Future - could add an API to backend to provide list of supported platforms - or remove this check and let backend do the validation.

Tabrizian (approved):

krishung5 (approved):

PR #224, opened 2023-06-16
Load PyTorch backend as a persistent backend

Tabrizian:


src/persistent_backend_manager.h
```diff
@@ -0,0 +1,61 @@
...
+  DISALLOW_COPY_AND_ASSIGN(PersistentBackendManager);
+  PersistentBackendManager(std::shared_ptr<TritonBackendManager> backend_manager) : backend_manager_(backend_manager) {
+  }
+  Status InitPersistentBackends(const triton::common::BackendCmdlineConfigMap& config_map);
+  Status InitPersistentBackend(
+      const std::string& backend_name,
+      const triton::common::BackendCmdlineConfigMap& config_map);
+
+  std::list<std::shared_ptr<TritonBackend>> persist_backends_;
+  std::shared_ptr<TritonBackendManager> backend_manager_;
```
dzier, line 57: Is there a reason we need to keep a point to the backend manager? We should avoid these doubly links when possible.

src/persistent_backend_manager.cc
```diff
@@ -0,0 +1,109 @@
...
+    const triton::common::BackendCmdlineConfig* config;
+    const auto& itr = config_map.find(backend_name);
+    if (itr == config_map.end()) {
+      config = &empty_backend_cmdline_config;
+    } else {
+      config = &itr->second;
+    }
+
+    std::shared_ptr<TritonBackend> persist_backend;
+    RETURN_IF_ERROR(backend_manager_->CreateBackend(
```
dzier, line 100: This is the only case I see where that internal backend_manager_ is being used. Why can we not just pass the backend_manager to this method? This is a private method that is only being called by create.

dzier (changes requested):

src/persistent_backend_manager.h
dzier: Wait, is this the whole reason for this class, to keep and hold onto the pointer? That is a bit of a hack :-/

Tabrizian: This code is mostly ported from an older Triton release: https://github.com/triton-inference-server/server/blob/r22.02/src/core/persistent_backend_manager.cc

I have updated the code to not store the `backend_manager_` as an attribute.

> Wait, is this the whole reason for this class, to keep and hold onto the pointer? That is a bit of a hack :-/

This class is mostly a singleton designed to load a set of backends before any backend/model is loaded.

src/persistent_backend_manager.cc
Tabrizian: Good suggestion, I've the updated the code to include this change.

dzier (approved):

src/persistent_backend_manager.cc
```diff
@@ -0,0 +1,113 @@
...
+{
+  for (const auto& be : {"pytorch"}) {
+    RETURN_IF_ERROR(InitPersistentBackend(be, config_map, backend_manager));
+  }
+
+  return Status::Success;
+}
+
+Status
+PersistentBackendManager::InitPersistentBackend(
```
nnshah1, line 75: To follow on @dzier's suggestions - if we just implemented this method alone and used it to initialize pytorch before other models - does that work as well? We could avoid the mutex, singleton and keep a static method: PreloadBackend()

Tabrizian: We still need the mutex since the user can create multiple TritonServer objects in the C-API and we want to make sure that this library is only dlopened once: https://github.com/triton-inference-server/server/blob/main/qa/L0_multi_server/test.sh

src/persistent_backend_manager.cc
```diff
@@ -0,0 +1,113 @@
...
+    if (itr == config_map.end()) {
+      config = &empty_backend_cmdline_config;
+    } else {
+      config = &itr->second;
+    }
+
+    std::shared_ptr<TritonBackend> persist_backend;
+    RETURN_IF_ERROR(backend_manager->CreateBackend(
+        backend_name, backend_dir, backend_libpath, *config, &persist_backend));
+    persist_backends_.push_back(persist_backend);
```
GuanLuo, line 107: I don't think you need a dedicated class for this, basically you just want the core to hold its own reference to the backend to keep its reference count > 0. The `list<shared_ptr<Backend>>` can be hold within backend manager?

Tabrizian: Fixed.

src/backend_manager.h
```diff
@@ -168,7 +173,12 @@ class TritonBackendManager {
  private:
   DISALLOW_COPY_AND_ASSIGN(TritonBackendManager);
   TritonBackendManager() = default;
+
+  Status PreloadBackend(
+      const std::string& backend_name,
+      const triton::common::BackendCmdlineConfigMap& config_map);
   std::unordered_map<std::string, std::shared_ptr<TritonBackend>> backend_map_;
+  std::list<std::shared_ptr<TritonBackend>> persist_backends_;
```
GuanLuo, line 181: After another look at the BackendManager, actually it doesn't even unload the backend once loaded, probably to avoid re-loading the framework libraries which may cause issue with static initialization. So may not even need this: `CreateBackend()` will put the loaded backend into `backend_map_`.

src/server.h
```diff
@@ -348,6 +348,7 @@ class InferenceServer {
   std::unique_ptr<ModelRepositoryManager> model_repository_manager_;
   std::shared_ptr<TritonBackendManager> backend_manager_;
   std::shared_ptr<TritonCacheManager> cache_manager_;
+  std::shared_ptr<PersistentBackendManager> persist_backend_manager_;
```
GuanLuo, line 351: Remove? I don't think it builds..

Tabrizian: Not sure why it compiled locally. Removed.

src/backend_manager.h
Tabrizian: Good point. Removed.

GuanLuo (dismissed):

nnshah1 (dismissed):

GuanLuo (approved):

PR #225, opened 2023-06-22
Upgrade S3 ListObjects version, allow >1000 model files

dyastremsky:
This PR upgrades the version of ListObjectsRequest and ListObjectsResult, as per AWS recommendations [here](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html) to switch to V2. It also enables Triton to load all files in an S3 bucket, beyond the [page limit of 1000](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html).

### DIFF
--- a/src/filesystem/implementations/s3.h
+++ b/src/filesystem/implementations/s3.h
@@ -501,31 +501,30 @@ S3FileSystem::GetDirectoryContents(
               list_objects_outcome.GetError().GetExceptionName() +
               ", error message: " +
               list_objects_outcome.GetError().GetMessage());
-    } else {
-      const auto& list_objects_result = list_objects_outcome.GetResult();
-      for (const auto& s3_object : list_objects_result.GetContents()) {
-        // In the case of empty directories, the directory itself will appear
-        // here
-        if (s3_object.GetKey().c_str() == full_dir) {
-          continue;
-        }
+    }
+    const auto& list_objects_result = list_objects_outcome.GetResult();
+    for (const auto& s3_object : list_objects_result.GetContents()) {
+      // In the case of empty directories, the directory itself will appear
+      // here
+      if (s3_object.GetKey().c_str() == full_dir) {
+        continue;
+      }
 
-        // We have to make sure that subdirectory contents do not appear here
-        std::string name(s3_object.GetKey().c_str());
-        int item_start = name.find(full_dir) + full_dir.size();
-        // S3 response prepends parent directory name
-        int item_end = name.find("/", item_start);
+      // We have to make sure that subdirectory contents do not appear here
+      std::string name(s3_object.GetKey().c_str());
+      int item_start = name.find(full_dir) + full_dir.size();
+      // S3 response prepends parent directory name
+      int item_end = name.find("/", item_start);
 
-        // Let set take care of subdirectory contents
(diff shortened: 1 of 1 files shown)


src/filesystem/implementations/s3.h
```diff
@@ -485,44 +485,56 @@ S3FileSystem::GetDirectoryContents(
...
+        done_listing = true;
+      }
+    } else {
+      return Status(
+          Status::Code::INTERNAL,
+          "Could not list contents of directory at " + true_path +
+              " due to exception: " +
+              list_objects_outcome.GetError().GetExceptionName() +
+              ", error message: " +
+              list_objects_outcome.GetError().GetMessage());
```
GuanLuo, line 536: Put this before the success handling, so the failure handle is a bit more obvious:
```
if (not_success) {
  return error;
}
// Check content..
```

kthui: I think @GuanLuo means this?
```
while (!done_listing) {
  auto list_objects_outcome = client_->ListObjectsV2(objects_request);

  if (!list_objects_outcome.IsSuccess()) {
    return Status(
        Status::Code::INTERNAL,
        "Could not list contents of directory at " + true_path +
            " due to exception: " +
            list_objects_outcome.GetError().GetExceptionName() +
            ", error message: " +
            list_objects_outcome.GetError().GetMessage());
  }

  const auto& list_objects_result = list_objects_outcome.GetResult();
  for (const auto& s3_object : list_objects_result.GetContents()) {
    // In the case of empty directories, the directory itself will appear here
    if (s3_object.GetKey().c_str() == full_dir) {
      continue;
    }

  // We have to make sure that subdirectory contents do not appear here
  std::string name(s3_object.GetKey().c_str());
  int item_start = name.find(full_dir) + full_dir.size();
  // S3 response prepends parent directory name
  int item_end = name.find("/", item_start);

  // Let set take care of subdirectory contents
  std::string item = name.substr(item_start, item_end - item_start);
  contents->insert(item);

  // Fail-safe check to ensure the item name is not empty
  if (item.empty()) {
    return Status(
        Status::Code::INTERNAL,
        "Cannot handle item with empty name at " + true_path);
    // We have to make sure that subdirectory contents do not appear here
    std::string name(s3_object.GetKey().c_str());
    int item_start = name.find(full_dir) + full_dir.size();
    // S3 response prepends parent directory name
    int item_end = name.find("/", item_start);

    // Let set take care of subdirectory contents
    std::string item = name.substr(item_start, item_end - item_start);
    contents->insert(item);

    // Fail-safe check to ensure the item name is not empty
    if (item.empty()) {
      return Status(
          Status::Code::INTERNAL,
          "Cannot handle item with empty name at " + true_path);
    }
  }
  // If there are more pages to retrieve, set the marker to the next page.
  if (list_objects_result.GetIsTruncated()) {
    objects_request.SetContinuationToken(
        list_objects_result.GetNextContinuationToken());
  } else {
    done_listing = true;
  }
}
```

GuanLuo: Yes, this have one less nesting for the success case

dyastremsky: Got it, that makes sense! Thanks for clarifying, @kthui. Done.

PR #227, opened 2023-06-26
 Add GitHub action to format and lint code

dyastremsky:
This pull request adds a pre-commit to run the following:
- Black for formatting Python
- Flake8 for linting Python for conformance to PEP8, PyFlakes, and circular complexity (max-line-length set to match Black's 88 default/recommended character limit; this and other settings from Black's documentation [here](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length) to avoid enforcement of limit when unnecessary).
- isort for deterministically sorting and organizing Python imports
- Clang for formatting C++, C, Java, TextProto, Proto, and CUDA
- Codespell for finding common spelling mistakes 

In addition, this adds these native pre-commit hooks, details [here](https://pre-commit.com/hooks.html):
- check-case-conflict
- check-executables-have-shebangs
- check-merge-conflict
- check-json
- check-toml
- check-yaml
- check-shebang-scripts-are-executable
- end-of-file-fixer
- mixed-line-ending
- requirements-txt-fixer
- trailing-whitespace

To run these locally, you can go into the repo and run `pip install pre-commit` and `pre-commit install`. After that has been done once, you just need to run `pre-commit run --all-files` any time you want to apply them. Once installed, it should run automatically on commit for files included in a commit. To correct spelling errors found by codespell when possible, you can install codespell (`pip install codespell` and potentially `pip install --upgrade codespell` to use the toml file in the directory) and call `codespell -w <path>`.

This pull request also adds a GitHub action so that these are run for every pull request and push to main. The core changes are in pre-commit.yml, pre-commit-config.yaml, and pyproject.toml. The rest of the changes are the effect of applying the action to the current repo.


.pre-commit-config.yaml
```diff
@@ -0,0 +1,92 @@
...
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+repos:
+- repo: https://github.com/psf/black
```
Tabrizian, line 54: I think we have been using yapf so for. Is it possible to stick with that or we want to update the contributing guide to reflect this style change: https://github.com/triton-inference-server/common/blob/main/tools/format.py

dyastremsky: Thanks for this comment. I put the rationale for the move from yapf to Black in the design doc (visible under Teams->General->Addition of Pre-Commit GitHub Actions). I'd quite emphatically recommend this move, and a few folks expressed strong buy-in. If there are concerns, I can also schedule a meeting for whomever wants to participate to make sure everyone's okay with the choice of formatter.

As far as the second part, that's the plan. Right now, the rollout strategy is to add this to `core`, `backend`, and `python_backend`. After a week, I'll create PRs for the rest of the repositories, including updating the contributing guide in server.

Tabrizian: I don't have a strong preference on yapf over black. Just wanted to make sure we're consistent.

Tabrizian (commented):
Is it possible to move `pyproject.toml` and `.pre-commit-config.yaml` to a new folder named `ci`?

dyastremsky:
> Is it possible to move `pyproject.toml` and `.pre-commit-config.yaml` to a new folder named `ci`?

@Tabrizian I could look into whether it is, but I'd rather keep this consistent with other projects. The pre-commit documentation [recommends this setup](https://pre-commit.com/), and it's used across other every other repo I've looked at so far like NVIDIA [Merlin](https://github.com/NVIDIA-Merlin/Transformers4Rec/tree/main), [RAPIDS cuML](https://github.com/rapidsai/cuml), and [NeMo](https://github.com/NVIDIA/NeMo/tree/main). They don't put it in their ci folders, and I think this level of standardization will make it easier for developers to know where to locate the files. In addition, packages like [codespell](https://github.com/codespell-project/codespell#using-a-config-file) and [isort](https://pycqa.github.io/isort/docs/configuration/config_files.html) look for the pyproject.toml folder in the current directory (from where it's called, which would be core's top-level directory in this case). I think at least codespell lets this behavior be overridden with an extra flag. However, if a user wanted to run codespell locally (the package, not the hook), they'd need to supply an extra flag to provide the new toml location to use the configuration instead of the default, adding in some work and introducing the possibility of user error.

Overall, my reasoning for keeping the current set-up would be that it'd be consistent with other projects using pre-commit, so users, hooks, and packages find these files in the expected place.

Tabrizian (approved):

PR #229, opened 2023-06-27
Modify ensemble to better reflect composing model response pattern

GuanLuo:
Previously ensemble will detect "decoupled" field and send responses that may not reflect the composing model's response pattern, i.e.:
if the composing model sends response and response complete flag separately,
```
TRITONBACKEND_ResponseSend(response, 0, nullptr /* success */);
TRITONBACKEND_ResponseFactorySendFlags(factory, TRITONSERVER_RESPONSE_COMPLETE_FINAL);
```
ensemble may send one response with the flag set or follow the same pattern, depending on whether "decoupled" is set in model config.

With this change the behavior will be more determined. And it reduces special handling in ensemble w.r.t. "decoupled" field in model config, which is meant to be used in server frontend (HTTP / gRPC).

### DIFF
--- a/src/ensemble_scheduler.cc
+++ b/src/ensemble_scheduler.cc
@@ -1263,17 +1263,26 @@ EnsembleContext::ScheduleSteps(
       // the request ownership out of step here to avoid that
       std::unique_ptr<InferenceRequest> request = std::move(step->request_);
       auto step_status = context->is_->InferAsync(request);
-      if (!step_status.IsOk()) {
+      if (step_status.IsOk()) {
+        step.release();
+        continue;
+      } else {
         std::lock_guard<std::mutex> lock(context->mutex_);
         context->ensemble_status_ = step_status;
-        // The request is not sent to server properly, shouldn't expect its
-        // release function get called.
-        context->request_tracker_->DecrementCounter();
-        context->ensemble_status_ = context->FinishEnsemble();
-        break;
       }
     }
-    step.release();
+
+    // Reaching here means the step is not being scheduled, update corresponding
+    // counters and attempt to finish ensemble if it is the last step.
+    std::lock_guard<std::mutex> lock(context->mutex_);
+    // The request is not sent to server properly, shouldn't expect its
+    // release function get called.
+    context->request_tracker_->DecrementCounter();
+    --context->inflight_step_counter_;
+
+    if (context->inflight_step_counter_ == 0) {
+      context->ensemble_status_ = context->FinishEnsemble();
+    }
   }
 }
 
--- a/src/ensemble_scheduler.cc
+++ b/src/ensemble_scheduler.cc
@@ -141,8 +141,8 @@ struct Step {
       gpu_output_map_;
   std::set<std::pair<std::string, IterationCount>> updated_tensors_;
 
-  // Location to store information convey through response callback,
-  // should be properly consumed by EnsembleContext and clean up before
+  // Storing information conveyed through response callback.
+  // This should be properly consumed by EnsembleContext and cleaned up before
   // returning from the callback.
   uint32_t response_flags_;
   TRITONSERVER_InferenceResponse* response_;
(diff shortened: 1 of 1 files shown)


src/ensemble_scheduler.cc
```diff
@@ -1032,18 +1041,28 @@ EnsembleContext::FinishEnsemble(std::unique_ptr<InferenceResponse>&& response)
...
+      InferenceResponse::Send(std::move(response), flags);
+      response_sent_ = true;
+    } else if (flags != 0) {
+      // check if any response is sent, if no, decoupled must be specified
+      if ((!info_->is_decoupled_) && (!response_sent_)) {
+        ensemble_status_ = Status(
+            Status::Code::INVALID_ARG,
+            "in ensemble '" + info_->ensemble_name_ + "', " +
+                request_tracker_->Request()->LogRequest() +
+                "unexpected deadlock, at least one output is not set while no "
```
dyastremsky, line 1057: Is there a way to detect a deadlock in decoupled models?

src/ensemble_scheduler.cc
```diff
@@ -1032,18 +1041,28 @@ EnsembleContext::FinishEnsemble(std::unique_ptr<InferenceResponse>&& response)
   }
 
   if (ensemble_status_.IsOk()) {
-    if (info_->is_decoupled_) {
-      if (response != nullptr) {
-        InferenceResponse::Send(std::move(response), 0 /* flags */);
-      }
-      if (inflight_step_counter_ != 0) {
-        return ensemble_status_;
+    auto flags = (inflight_step_counter_ == 0)
```
oandreeva-nv, line 1044: I am wondering of this `flags` is actually needed. We only use it on L1048. Should we also use it on L1064?

src/ensemble_scheduler.cc
```diff
@@ -1100,14 +1121,8 @@ EnsembleContext::CheckAndSetEnsembleOutput(
...
   if (!ready) {
-    if (info_->is_decoupled_) {
-      return Status::Success;
-    }
-    return Status(
-        Status::Code::INVALID_ARG,
-        lrequest->LogRequest() +
-            "unexpected deadlock, at least one output is not set while no more "
-            "ensemble steps can be made");
+    // Do not create response if the actual response is not ready.
```
oandreeva-nv, line 1124: So this was partially related to original issue?  We were creating a response, when it was not ready?

src/ensemble_scheduler.cc
```diff
@@ -774,11 +781,13 @@ EnsembleContext::PrepareSteps(
         ensemble_status_ = GetNextSteps(updated_tensors, ready_steps);
       }
 
-      // Check and send ensemble response
-      if ((!ensemble_status_.IsOk()) || (inflight_step_counter_ == 0) ||
-          info_->is_decoupled_) {
+      // Reaching the end of processing completed step, check if the ensemble
+      // should send response / error.
+      if (!ensemble_status_.IsOk()) {
```
oandreeva-nv, line 786: I see, so the previous logic was checking `ensemble_status_`, `inflight_step_counter_` and `info_->is_decoupled_ ` here. The new logic introduced a more precise behavior based on  `inflight_step_counter_` and `info_->is_decoupled_  values in combination with `response_sent_`.

src/ensemble_scheduler.cc
```diff
@@ -340,6 +350,10 @@ class EnsembleContext {
   // Objects related to the ensemble infer request
   Status ensemble_status_;
   RequestTracker* request_tracker_;
+  // Use in conjunction with 'is_decoupled_' in EnsembleInfo to
+  // better distinguish ensemble ending behavior (see documentation in
+  // FinishEnsemble for detail).
```
oandreeva-nv, line 355: My only complain here is that I was expecting to see **documentation**. I think `annotation` would be better for current case

src/ensemble_scheduler.cc
GuanLuo: Yes, previously the conditional will "trim" the number of possible ensemble responses, which causes the ensemble response not being filled properly.

oandreeva-nv (dismissed):
LGTM

tanmayv25 (dismissed):

oandreeva-nv (approved):

oandreeva-nv (approved):

PR #231, opened 2023-07-04
Unify default naming on different instance group count

kthui:
server PR: https://github.com/triton-inference-server/server/pull/6018

~~There are three small but important changes:~~

~~Model instance name:~~
- ~~Assuming there is an instance group named `group`, and its `count` can vary.~~
- ~~Previously, instances from the group is named:~~
  - ~~If `count > 1`, `["group_0", "group_1", "group_2", ...]`.~~
  - ~~If `count == 1`, `["group"]`.~~
- ~~Now, the instances are named:~~
  - ~~Regardless of `count`, `["group", "group_1", "group_2", "group_3", ...]`.~~
  - ~~In other words, all `group_0` will be named `group`.~~

~~A model instance can only be reused if it will not be renamed upon reuse, based on the above new naming rule. For example:~~
- ~~Previously, instances named `["1", "2", "3"]` will be reused if they are renamed to `["1", "A", "B"]`, providing the instances are no different to each other beside than their names.~~
- ~~Now, only the instance named `"1"` can be reused, because other instances are renamed.~~

~~For the model instance signature hash function, only the instance name is used for hashing, because it is reasonable to believe instances within the same model should not share the same name, so it reduces the byte size passed to the hash function. If the user would to specify the same name across all instances (becomes linear search), the performance impact is expected to be minimal, as most use-cases do not create enough instances to impact performance.~~

### DIFF
--- a/src/backend_model_instance.cc
+++ b/src/backend_model_instance.cc
@@ -255,7 +255,7 @@ TritonModelInstance::SetInstances(
         const auto& kind = std::get<1>(is);
         const auto& id = std::get<2>(is);
 
-        const Signature signature(group, id);
+        const Signature signature(instance_name, id, group);
         // Check if an existing instance can be re-used.
         if (!ShareBackendThread(model->DeviceBlocking(), kind)) {
           auto itr = existing_instances.find(signature);
--- a/src/backend_model_instance.cc
+++ b/src/backend_model_instance.cc
@@ -219,8 +219,7 @@ TritonModelInstance::SetInstances(
           secondary_device.device_id());
     }
     for (int32_t c = 0; c < group.count(); ++c) {
-      std::string instance_name{
-          c > 0 ? group.name() + "_" + std::to_string(c) : group.name()};
+      std::string instance_name{group.name() + "_" + std::to_string(c)};
       const bool passive = group.passive();
       std::vector<std::tuple<
           std::string, TRITONSERVER_InstanceGroupKind, int32_t,
--- a/src/backend_model_instance.cc
+++ b/src/backend_model_instance.cc
@@ -220,8 +220,7 @@ TritonModelInstance::SetInstances(
     }
     for (int32_t c = 0; c < group.count(); ++c) {
       std::string instance_name{
-          group.count() > 1 ? group.name() + "_" + std::to_string(c)
-                            : group.name()};
+          c > 0 ? group.name() + "_" + std::to_string(c) : group.name()};
       const bool passive = group.passive();
       std::vector<std::tuple<
           std::string, TRITONSERVER_InstanceGroupKind, int32_t,
--- a/src/backend_model_instance.cc
+++ b/src/backend_model_instance.cc
@@ -253,7 +253,7 @@ TritonModelInstance::SetInstances(
         const auto& kind = std::get<1>(is);
         const auto& id = std::get<2>(is);
 
-        const Signature signature(instance_name, id, group);
+        const Signature signature(group, id);
         // Check if an existing instance can be re-used.
         if (!ShareBackendThread(model->DeviceBlocking(), kind)) {
           auto itr = existing_instances.find(signature);
--- a/src/backend_model_instance.h
+++ b/src/backend_model_instance.h
@@ -63,24 +63,24 @@ class TritonModelInstance {
   class Signature {
    public:
     Signature(
-        const inference::ModelInstanceGroup& group_config, int32_t device_id)
-        : group_config_(group_config), device_id_(device_id),
-          hash_(std::hash<std::string>{}(
-              std::to_string(device_id_) +
-              InstanceConfigSignature(group_config_)))
+        const std::string& name, int32_t device_id,
+        const inference::ModelInstanceGroup& group_config)
+        : name_(name), device_id_(device_id), group_config_(group_config),
+          hash_(std::hash<std::string>{}(name_))
     {
     }
     bool operator==(const Signature& rhs) const
     {
-      return device_id_ == rhs.device_id_ &&
+      return name_ == rhs.name_ && device_id_ == rhs.device_id_ &&
              EquivalentInInstanceConfig(group_config_, rhs.group_config_);
     }
     bool operator!=(const Signature& rhs) const { return !(*this == rhs); }
     std::size_t Hash() const { return hash_; }
 
    private:
-    const inference::ModelInstanceGroup group_config_;
+    const std::string name_;
     const int32_t device_id_;
+    const inference::ModelInstanceGroup group_config_;
     const std::size_t hash_;
   };
 
--- a/src/backend_model_instance.h
+++ b/src/backend_model_instance.h
@@ -63,24 +63,24 @@ class TritonModelInstance {
   class Signature {
    public:
     Signature(
-        const std::string& name, int32_t device_id,
-        const inference::ModelInstanceGroup& group_config)
-        : name_(name), device_id_(device_id), group_config_(group_config),
-          hash_(std::hash<std::string>{}(name_))
+        const inference::ModelInstanceGroup& group_config, int32_t device_id)
+        : group_config_(group_config), device_id_(device_id),
+          hash_(std::hash<std::string>{}(
+              std::to_string(device_id_) +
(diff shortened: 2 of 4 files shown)


src/backend_model_instance.cc
```diff
@@ -220,8 +220,7 @@ TritonModelInstance::SetInstances(
     }
     for (int32_t c = 0; c < group.count(); ++c) {
       std::string instance_name{
-          group.count() > 1 ? group.name() + "_" + std::to_string(c)
-                            : group.name()};
+          c > 0 ? group.name() + "_" + std::to_string(c) : group.name()};
```
rmccorm4, line 223: Nice detailed explanation. Two comments:

> In other words, all group_0 will be named group.

1. I was thinking the opposite, we should change "group" to "group_0" to have a consistent naming scheme regardless of count. 

2. What do you think about including device type / kind in the name? ex: "bert_cpu0_instance0", "bert_cpu0_instance1", "bert_gpu0_instance0", etc.

I think this would improve the intuition in the logging and debugging from user perspective. For example it may be more clear that a GPU instance is receiving work rather than a CPU instance in logs, compared to current naming convention.

kthui: > I was thinking the opposite, we should change "group" to "group_0" to have a consistent naming scheme regardless of count.

I am good with any scheme, as long as the naming is consistent. Update commit: [Use different instance name unification method](https://github.com/triton-inference-server/core/pull/231/commits/af88121df20fa8d578e1d7a6a791b8f72a203a78)

kthui: > What do you think about including device type / kind in the name? ex: "bert_cpu0_instance0", "bert_cpu0_instance1", "bert_gpu0_instance0", etc.

This is a good idea. I think users generally do not supply an instance name if it is optional, but when debugging is needed, they would want to see descriptive names.

But I think this change is out of scope for this PR, because
- [The code](https://github.com/triton-inference-server/core/blob/main/src/model_config_utils.cc#L779-L788) where the default name is assigned to instances is on a different file, independent from this change; and
- We can have more time design on the default instance name for different scenarios.

What do you think about the scope cutoff? On this PR, we only address how to assign different names to instances on the same instance group (i.e. `"group_default_name_0", "group_default_name_1", ...`). We can decide on the `group_default_name` on a different PR.

src/backend_model_instance.cc
```diff
@@ -219,9 +219,7 @@ TritonModelInstance::SetInstances(
           secondary_device.device_id());
     }
     for (int32_t c = 0; c < group.count(); ++c) {
-      std::string instance_name{
-          group.count() > 1 ? group.name() + "_" + std::to_string(c)
-                            : group.name()};
+      std::string instance_name{group.name() + "_" + std::to_string(c)};
```
GuanLuo, line 222: Do we have check that the groups must not share the same name? Otherwise we can have instances with the exact same name?

src/backend_model_instance.cc
GuanLuo: Looking at how the default group name is set and the signature check, so if user change the config from:
```
instance_group [
  { kind: KIND_CPU },
  { kind: KIND_GPU },
]
```
to
```
instance_group [
  { kind: KIND_GPU },
  { kind: KIND_CPU },
]
```
No instance will be reused?

src/backend_model_instance.cc
kthui: > Do we have check that the groups must not share the same name?

No, we actually allow one or more instances to share the same name. There are two scenarios:
1. One or more instances are the **same** and share the same name.
    - Exactly which instance is reused or removed during an instance update is undefined, because there is no way to tell the difference between those instances.
    - There is a [new test](https://github.com/triton-inference-server/server/pull/6018/files#diff-40ca2af021f4288f3fc6ae6071a244bca6350a0e657ab2822b61650cac266931R217-R230) that manipulate instances under this case. The test will pass as long as the number of instances added or removed is correct.
2. One or more instances are **different** but share the same name.
    - The instances are disambiguated by their instance group configuration.
    - There is a [new test](https://github.com/triton-inference-server/server/pull/6018/files#diff-40ca2af021f4288f3fc6ae6071a244bca6350a0e657ab2822b61650cac266931R231-R234) that the correct instance is removed and created, if an instance config is changed but still share the same name.

> Otherwise we can have instances with the exact same name?

Yes. They are distinguished by their instance configuration.

src/backend_model_instance.cc
kthui: Correct, because of the default name assigned to the instances will change if the location of the instances on the configuration changes.

kthui: This is the reason why I replaced the [test_gpu_cpu_instance_update](https://github.com/triton-inference-server/server/pull/6018/files#diff-40ca2af021f4288f3fc6ae6071a244bca6350a0e657ab2822b61650cac266931L190) test with the [test_instance_name_update](https://github.com/triton-inference-server/server/pull/6018/files#diff-40ca2af021f4288f3fc6ae6071a244bca6350a0e657ab2822b61650cac266931R190) and [test_instance_same_name](https://github.com/triton-inference-server/server/pull/6018/files#diff-40ca2af021f4288f3fc6ae6071a244bca6350a0e657ab2822b61650cac266931R211) tests.

rmccorm4: Thanks for making the change for default name to be consistent with `group_0` regardless of count.

Regarding this:
> But I think this change is out of scope for this PR, because

>
> [The code](https://github.com/triton-inference-server/core/blob/main/src/model_config_utils.cc#L779-L788) where the default name is assigned to instances is on a different file, independent from this change; and
We can have more time design on the default instance name for different scenarios.
What do you think about the scope cutoff? On this PR, we only address how to assign different names to instances on the same instance group (i.e. "group_default_name_0", "group_default_name_1", ...). We can decide on the group_default_name on a different PR.

I'm fine with cutting off scope to separate PR. Please make a ticket and link it.

GuanLuo: This behavior is counter-intuitive, what is the reason of doing so?

src/backend_model_instance.cc
rmccorm4: I think the ability for different instances to have the same name will be confusing in the long term, but is functional now since the logic is based on config and not names. However, with my other comment to add more descriptive  instance names like this:

> What do you think about including device type / kind in the name? ex: "bert_cpu0_instance0", "bert_cpu0_instance1", "bert_gpu0_instance0", etc.

Then I think the names will be unique by default, unless users specify an identical instance group name in two or more models.

GuanLuo: What is the reason of allowing instances with the same name?

kthui: It is always allowed. I am not sure on the reason why it is allowed on the first place, in the past.

src/backend_model_instance.cc
kthui: It is because of the logic which assigns default name to instances that do not specify a name.

I think there is a way to avoid the behavior, but it will be complicated: During an instance update, leave the default instance name empty, so the lookup can be done purely on the instance config. Instances with the name specified will be matched first, and then for those that do not specify a name. Those without a name will use the name on the instance that is matched, if any. Otherwise, a default name is assigned.

Maybe, we can simply print on the log notifying the user when an instance reused has a different name than the one requested.

src/backend_model_instance.cc
kthui: Other than what is currently implemented. I do not think there is a good reason to allow instances with the same name.

src/backend_model_instance.cc
kthui: Other than minimizing the scope of the code change. I agree the behavior is counter-intuitive. Maybe a redesign on how the default name works is needed.

nnshah1 (approved):
Providing consistent naming independent of count seems like a good step. Additional rules for matching instances can be handled separately. Defer to @GuanLuo and @rmccorm4 for final approvals.

rmccorm4 (approved):
I'm happy with this step for now if we need to go back and rethink some of the open questions from discussion.

GuanLuo (approved):
PR title needs to be changed?

PR #233, opened 2023-07-06
Fix InstanceGroup normalization when using backend attributes

tanmayv25:
The change resets the kind setting from the instance group being auto-completed, if the preferred group is not supported because of unavailable GPUs.

On CPU only machines:

### Before
UNAVAILABLE: Invalid argument: instance group resnet_v1_50_savedmodel of model resnet_v1_50_savedmodel has kind KIND_GPU but no GPUs are available

### After
Model loads properly.

### DIFF
--- a/src/model_config_utils.cc
+++ b/src/model_config_utils.cc
@@ -751,6 +751,8 @@ NormalizeInstanceGroup(
       if (pg.kind() == inference::ModelInstanceGroup::KIND_GPU) {
         // Don't use preferred group with KIND_GPU if there is no GPU.
         if (supported_gpus.empty()) {
+          group->set_kind(inference::ModelInstanceGroup::KIND_AUTO);
+          group->set_count(0);
           continue;
         }
         // If preferred group sets GPUs, limit deployment onto those that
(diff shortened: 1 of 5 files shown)


src/model_config_utils.cc
```diff
@@ -751,6 +751,8 @@ NormalizeInstanceGroup(
       if (pg.kind() == inference::ModelInstanceGroup::KIND_GPU) {
         // Don't use preferred group with KIND_GPU if there is no GPU.
         if (supported_gpus.empty()) {
+          group->set_kind(inference::ModelInstanceGroup::KIND_AUTO);
+          group->set_count(0);
```
GuanLuo, line 755: Should we rather just move line 748-749 down so we only set kind / count once when the preferred group is valid?

nv-kmcgill53:
Are there test updates as well in the server repository, I don't see a PR out yet? Also, does the resulting model config return the correct fields?

tanmayv25:
> Are there test updates as well in the server repository, I don't see a PR out yet? Also, does the resulting model config return the correct fields?

We don't have a test PR out yet. It will be part of the testing effort for TF savedmodel platform handler.

nv-kmcgill53 (approved):

src/model_config_utils.cc
tanmayv25: Addressed here: https://github.com/triton-inference-server/core/pull/233/commits/749e46b013a563bec9042dff6f28b5013874bad9

krishung5 (approved):

PR #235, opened 2023-07-21
Add BackendAttribute for parallel model instance loading

rmccorm4:
Current default value of the attribute is false, so no parallel loading will be done unless a backend implements and calls the API to enable it. This is just the scaffolding for such capabilities.

---

Identity Backend opt-in for sanity check in existing pipelines: https://github.com/triton-inference-server/identity_backend/pull/26

Seeing ballpark 2-3x speedup on my machine with Identity Backend loading 100 instances. I think the speedup will be more meaningful for model instances that take longer to initialize in more complicated backends. Will do more thorough performance analysis after enabling support in other backends.

### DIFF
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -1497,7 +1497,7 @@ TRITONBACKEND_BackendAttributeAddPreferredInstanceGroup(
 /// \param enabled Whether or not the backend supports loading model instances
 /// in parallel.
 TRITONSERVER_DECLSPEC TRITONSERVER_Error*
-TRITONBACKEND_BackendAttributeSetParallelInstanceLoading(
+TRITONBACKEND_BackendAttributeSetParallelModelInstanceLoading(
     TRITONBACKEND_BackendAttribute* backend_attributes, bool enabled);
 
 /// TRITONBACKEND Batching
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -94,7 +94,7 @@ struct TRITONBACKEND_Batcher;
 ///   }
 ///
 #define TRITONBACKEND_API_VERSION_MAJOR 1
-#define TRITONBACKEND_API_VERSION_MINOR 13
+#define TRITONBACKEND_API_VERSION_MINOR 14
 
 /// Get the TRITONBACKEND API version supported by Triton. This value
 /// can be compared against the TRITONBACKEND_API_VERSION_MAJOR and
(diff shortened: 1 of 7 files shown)


src/backend_model_instance.cc
```diff
@@ -204,6 +206,17 @@ TritonModelInstance::SetInstances(
...
+
+  // Deferred will be lazily evaluated when the result is requested. Since the
+  // creation_results are requested serially below, this is equivalent to making
+  // the calls serially.
+  auto launch_policy = std::launch::deferred;
+  // If the backend supports it, std::launch::async will allow the calls to be
+  // made concurrently.
+  if (model->Backend()->BackendAttributes().parallel_instance_loading_) {
+    launch_policy = std::launch::async;
+  }
```
rmccorm4, line 219: Note, I thought this would be a clean way of handling the if/else behavior of whether the parallel loading is supported or not. However, I understand that the `deferred` logic may not be intuitive to reason as the same as "sequential/serial" loading.

If you don't like this approach, I can simplify it to be more straightforward, maybe something like this.
```cpp
// Used to gather async future results, useless for non-parallel case
std::vector<std::future<Status>> creation_results;
...
// assume some lambda or helper func exists for InstanceCreate logic

for each instance {
    if (parallel_loading_supported) {
      creation_results.emplace_back(std::async(std::launch::async, InstanceCreateHelper(args...)));
    } else {
      RETURN_IF_ERROR(InstanceCreateHelper(args...));
    }
}

if (parallel_loading_supported) {
  for result in creation_results {
    // check results, set retval if any failures
  }
}

return Status::Success;
```

include/triton/core/tritonbackend.h
```diff
@@ -1480,6 +1480,26 @@ TRITONBACKEND_BackendAttributeAddPreferredInstanceGroup(
...
+/// default, this attribute is false and calls to the
+/// TRITONBACKEND_ModelInstanceInitialize function will be made serially. If
+/// this attribute is set to true, then Triton will make calls to
+/// TRITONBACKEND_ModelInstanceInitialize concurrently.
+///
+/// \param backend_attributes The backend attributes object.
+/// \param enabled Whether or not the backend supports loading model instances
+/// in parallel.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error*
+TRITONBACKEND_BackendAttributeSetParallelInstanceLoading(
```
Tabrizian, line 1500:
```suggestion
TRITONBACKEND_BackendAttributeSetParallelModelInstanceLoading(
```

src/backend_model.cc
```diff
@@ -406,6 +406,40 @@ TritonModel::PrepareInstances(
...
+      parallel = (s_env == "1") ? true : false;
+      LOG_INFO << "Using TRITON_PARALLEL_INSTANCE_LOADING environment variable "
+                  "override: "
+               << parallel;
+    }
+  }
+
+  // If the backend supports it, std::launch::async will allow concurrent calls
+  if (parallel) {
+    launch_policy = std::launch::async;
```
Tabrizian, line 434: This looks great!

Tabrizian (commented):
The API LGTM.

src/backend_model.cc
```diff
@@ -485,17 +519,26 @@ TritonModel::PrepareInstances(
...
-                       << "' with device id '" << is.device_id_ << "'";
-        RETURN_IF_ERROR(TritonModelInstance::CreateInstance(
-            this, instance_name, signature, is.kind_, is.device_id_,
-            profile_names, passive, is.policy_name_, *is.rate_limiter_config_,
-            secondary_devices, &new_instance));
-        added_instances->push_back(new_instance);
-        RegisterBackgroundInstance(std::move(new_instance), passive);
+        // Note that the local variables should be captured by value
+        creation_results.emplace_back(
+            std::async(launch_policy, [=, &instance_mu]() {
```
GuanLuo, line 524: Is it okay to copy all local variables? `creation_results` will be copied?

rmccorm4: According to docs/examples I read:

> A default capture (also called a capture-default) captures all variables that are mentioned in the lambda. Variables not mentioned in the lambda are not captured if a default capture is used.

A couples refs:
- https://learn.microsoft.com/en-us/cpp/cpp/lambda-expressions-in-cpp?view=msvc-170#:~:text=You%20can%20use%20a%20capture%2Ddefault%20mode%20to%20indicate%20how%20to%20capture%20any%20outside%20variables%20referenced%20in%20the%20lambda%20body
- https://www.learncpp.com/cpp-tutorial/lambda-captures/#:~:text=A%20default%20capture%20(also%20called,use%20a%20capture%20value%20of%20%3D%20.

I'm happy to make it more explicit if you'd prefer, I'm still cleaning up and finalizing on some finer details like this.

rmccorm4: @GuanLuo does 👍 you want more explicit? Or you're happy with it?

Tabrizian: How about using `&` instead of `=`?

PR #237, opened 2023-08-02
Disable pre-fetching when using queue policy

tanmayv25:
Disabling pre-fetching so queue policy can be locally applied to the model queue.

Test PR: https://github.com/triton-inference-server/server/pull/6133

Fixes https://github.com/triton-inference-server/server/issues/5783

### DIFF
--- a/src/rate_limiter.cc
+++ b/src/rate_limiter.cc
@@ -158,19 +158,78 @@ RateLimiter::UnregisterModel(const TritonModel* model)
   }
 }
 
-bool
-RateLimiter::PayloadSlotAvailable(const TritonModel* model)
+void
+RateLimiter::WaitForConsumer(
+    const TritonModel* model, const TritonModelInstance* model_instance)
 {
-  bool result;
   PayloadQueue* payload_queue = nullptr;
   {
     std::lock_guard<std::mutex> lk(payload_queues_mu_);
+    if (payload_queues_.find(model) == payload_queues_.end()) {
+      LOG_ERROR << "Should not print this! Waiting for the consumer for an "
+                   "unknown model";
+      return;
+    }
     payload_queue = payload_queues_[model].get();
   }
+
+  if (model_instance == nullptr) {
+    payload_queue->queue_->WaitForConsumer();
+  } else {
+    payload_queue->specific_queues_[model_instance]->WaitForConsumer();
+  }
+}
+
+
+int
+RateLimiter::WaitingConsumerCount(
+    const TritonModel* model, const TritonModelInstance* model_instance)
+{
+  PayloadQueue* payload_queue = nullptr;
   {
-    std::lock_guard<std::mutex> lk(payload_queue->mu_);
-    result = payload_queue->queue_->Size() <
-             2 * payload_queue->specific_queues_.size();
+    std::lock_guard<std::mutex> lk(payload_queues_mu_);
+    if (payload_queues_.find(model) == payload_queues_.end()) {
+      LOG_ERROR << "Should not print this! Waiting for the consumer for an "
+                   "unknown model";
+      return 0;
+    }
+    payload_queue = payload_queues_[model].get();
+  }
+
+  if (model_instance == nullptr) {
+    return payload_queue->queue_->WaitingConsumerCount();
+  } else {
+    return payload_queue->specific_queues_[model_instance]
+        ->WaitingConsumerCount();
+  }
+}
+
+bool
+RateLimiter::PayloadSlotAvailable(
+    const TritonModel* model, const TritonModelInstance* model_instance,
+    const bool support_prefetching, const bool force_non_blocking)
+{
+  bool result;
+  if (support_prefetching) {
+    PayloadQueue* payload_queue = nullptr;
+    {
+      std::lock_guard<std::mutex> lk(payload_queues_mu_);
+      payload_queue = payload_queues_[model].get();
+    }
+    {
+      std::lock_guard<std::mutex> lk(payload_queue->mu_);
+      auto multiplier = (model_instance == nullptr)
+                            ? payload_queue->specific_queues_.size()
+                            : 1;
+      result = payload_queue->queue_->Size() < (2 * multiplier);
+    }
+  } else {
+    result = true;
+    if (force_non_blocking) {
+      result = (WaitingConsumerCount(model, model_instance) > 0);
+    } else {
+      WaitForConsumer(model, model_instance);
+    }
   }
   return result;
 }
@@ -190,6 +249,14 @@ RateLimiter::EnqueuePayload(
     }
     payload_queue = payload_queues_[model].get();
   }
+
+  // Update the pending consumer counts to prevent additional
+  // requests from getting enqueued.
+  if (pinstance != nullptr) {
+    payload_queue->specific_queues_[pinstance]->DecrementConsumerCount();
+  }
+  payload_queue->queue_->DecrementConsumerCount();
+
   {
     std::lock_guard<std::mutex> lk(payload_queue->mu_);
     payload->SetState(Payload::State::REQUESTED);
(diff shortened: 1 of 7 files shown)


src/rate_limiter.cc
```diff
@@ -274,10 +350,31 @@ RateLimiter::DequeuePayload(
...
+        ->DecrementConsumerCount();
     instances.pop_front();
   } else {
     instances.erase(instances.begin() + instance_index);
   }
+
+  // Decrement the counts from the remaining specific
+  // instance handling as there will be no consumer for
+  // these queues.
+  // FIXME: For more accurate handling, these instances
```
dyastremsky, line 367: Would you be able to create a JIRA ticket for this and link it here? Ideally with more details. I'm not understanding from the comment what the concern about the current approach is, the costs (versus the other one), and why we're using an approximation here, so getting that information into the JIRA ticket would be good.

src/rate_limiter.cc
```diff
@@ -158,19 +158,78 @@ RateLimiter::UnregisterModel(const TritonModel* model)
...
+    {
+      std::lock_guard<std::mutex> lk(payload_queues_mu_);
+      payload_queue = payload_queues_[model].get();
+    }
+    {
+      std::lock_guard<std::mutex> lk(payload_queue->mu_);
+      auto multiplier = (model_instance == nullptr)
+                            ? payload_queue->specific_queues_.size()
+                            : 1;
+      result = payload_queue->queue_->Size() < (2 * multiplier);
```
dyastremsky, line 224: I see this already existed in the code, but could you add a comment helping understand what's going on here? Specifically, why does pre-fetching require 2 * multiplier space and what does the multiplier mean?

src/rate_limiter.cc
```diff
@@ -158,19 +158,78 @@ RateLimiter::UnregisterModel(const TritonModel* model)
...
-    result = payload_queue->queue_->Size() <
-             2 * payload_queue->specific_queues_.size();
+    std::lock_guard<std::mutex> lk(payload_queues_mu_);
+    if (payload_queues_.find(model) == payload_queues_.end()) {
+      LOG_ERROR << "Should not print this! Waiting for the consumer for an "
+                   "unknown model";
+      return 0;
+    }
+    payload_queue = payload_queues_[model].get();
+  }
```
nv-kmcgill53, line 197: nit: This is a completely subjective opinion, but you could make getting the payload queue it's own function.  I think it reads fine either way and will not block this PR on this.

src/rate_limiter.cc
```diff
@@ -158,19 +158,78 @@ RateLimiter::UnregisterModel(const TritonModel* model)
...
+  bool result;
+  if (support_prefetching) {
+    PayloadQueue* payload_queue = nullptr;
+    {
+      std::lock_guard<std::mutex> lk(payload_queues_mu_);
+      payload_queue = payload_queues_[model].get();
+    }
+    {
+      std::lock_guard<std::mutex> lk(payload_queue->mu_);
+      auto multiplier = (model_instance == nullptr)
```
nv-kmcgill53, line 221: This `auto` is probably ok since it should get typed to `size_t` rather than an `int`. My style is to be explicit in these situations to tell the reader of the code we do not want integer overflow in the preceding lines. 

I will preface this next comment with, "I don't know the implicit casting rules when using the `<` operator." This has the very small potential to be a pain in line 224 if multiplier is typed to `int` and `(2 * multiplier)` gets overflowed. Again, this shouldn't be a problem since we don't expect `multiplier` to be greater than 2^32-1.

src/rate_limiter.cc
```diff
@@ -190,6 +249,14 @@ RateLimiter::EnqueuePayload(
     }
```
nv-kmcgill53, line 249: Should probably update this error message as well.

src/rate_limiter.cc
```diff
@@ -230,15 +297,24 @@ RateLimiter::DequeuePayload(
 {
   payload->reset();
   PayloadQueue* payload_queue = nullptr;
+  auto model = instances[0]->Model();
   {
     std::lock_guard<std::mutex> lk(payload_queues_mu_);
-    if (payload_queues_.find(instances[0]->Model()) == payload_queues_.end()) {
+    if (payload_queues_.find(model) == payload_queues_.end()) {
       LOG_ERROR << "Should not print this! Dequeuing payload with an unknown "
                    "instance.";
```
nv-kmcgill53, line 305: `"Dequeuing a payload which has no queue for model " << model << "."`

src/rate_limiter.cc
```diff
@@ -274,10 +350,31 @@ RateLimiter::DequeuePayload(
   (*payload)->Callback();
   if ((*payload)->GetInstance() == nullptr) {
     (*payload)->SetInstance(instances.front());
+    // Enqueue did not specify the specific instance to
+    // run with the payload. Hence, need to explicitly
+    // decrement the consumer count for the instance
+    // which got allocated.
+    payload_queue->specific_queues_[instances.front()]
+        ->DecrementConsumerCount();
```
nv-kmcgill53, line 358: Lines 313 to 316 increment both the model scoped consumer count and the instance specific instance count. Shouldn't we have to decrement the model scoped consumer count here as well?

tanmayv25: The model scoped counter is decremented when we EnqueuePayload the request to the RateLimiter in L258. For accurate handling we should decrement the specific_queues_ count within EnqueuePayload call itself. However, we are not aware of the association between the instances being listened to by the consumer threads.

src/rate_limiter.h
```diff
@@ -93,7 +93,9 @@ class RateLimiter {
   /// Returns true if there is a payload slot available for the given model.
   /// \param model The pointer to TritonModel object to be removed.
   /// \return slot availability in boolean.
-  bool PayloadSlotAvailable(const TritonModel* model);
+  bool PayloadSlotAvailable(
+      const TritonModel* model, const TritonModelInstance* model_instance,
+      const bool support_prefetching, const bool force_non_blocking = false);
```
GuanLuo, line 98: document the variables

src/rate_limiter.cc
```diff
@@ -158,19 +158,78 @@ RateLimiter::UnregisterModel(const TritonModel* model)
...
+  }
+}
+
+bool
+RateLimiter::PayloadSlotAvailable(
+    const TritonModel* model, const TritonModelInstance* model_instance,
+    const bool support_prefetching, const bool force_non_blocking)
+{
+  bool result;
+  if (support_prefetching) {
```
GuanLuo, line 213: is `force_non_blocking` only use when `!support_prefectching`? Why this variable doesn't matter if `support_prefetching`?

tanmayv25: When pre-fetching, the call is always non-blocking.

GuanLuo: I guess my confusion is that why it needs to block for some cases if disabling pre-fetching? My impression is that pre-fetching or not only changes the "timing" of no available payload, why it will change the scheduler behavior: from always non-blocking to sometime blocking.

tanmayv25: This is a good question. There exists an assymetry in batcher thread based upon pre-fetch settings. 

The function PayloadSlotAvailable is called in two places. First in the Enqueue function to determine whether or not to wake up the batcher thread. There is no need to wake up the batcher thread if there isn't a slot available. Additionally, the call here should **not be blocking** as the thread is just enqueueing requests on policy queue. This should be symmetric for both pre-fetch enable as well as disable.


In batcher thread, assymetry is introduced by enabling pre-fetching. The batcher thread hold some requests in curr_payload_ object without enqueuing the payload to RateLimiter. Hence, even if the payload slot is not available on ratelimiter queue, we can keep building requests on curr_payload and push the payload as soon as a slot in the RateLimiter appears.
When pre-fetching is not enabled, the batcher thread shouldn't pull any requests from the policy queue until there is an idle runner calling DequeuePayload on the RateLimiter.

dyastremsky (dismissed):

src/rate_limiter.h
```diff
@@ -91,9 +91,18 @@ class RateLimiter {
...
 
   /// Returns true if there is a payload slot available for the given model.
-  /// \param model The pointer to TritonModel object to be removed.
+  /// Note the function can be a blocking call when support_prefetching is
+  /// false. In this case, the function will block until a slot is available to
+  /// start building the payload. force_non_blocking option can be set to True
+  /// to allow function to return back with availability. \param model The
+  /// pointer to TritonModel object to query for . \param model_instance The
+  /// pointer to TritonMode \param support_prefetching Whether or not
+  /// pre-fetching of payloads is enabled. \param force_non_blocking When set
```
GuanLuo, line 100: Newlines?

GuanLuo (dismissed):

GuanLuo (approved):

PR #238, opened 2023-08-03
[DO NOT MERGE] Python binding / wrapper look and feel

GuanLuo:
`FIXME` are the sections that further discussion is desired.

The wrapper is restricted that a lot of interaction with in-process API is pre-defined (i.e. how to handle released `TRITONSERVER_Request` and `TRITONSERVER_Response`). The intention is that user shouldn't need to handle object lifecycle explicitly. The lower level binding will also be provided if they want finer control (i.e. reuse underlying objects to avoid alloc/release overhead).

At the end of `_server.py` is a basic example usage of the wrapper. At the end of `_infer.py` is an example implementation of allocator.

### DIFF
--- a/python/_infer.py
+++ b/python/_infer.py
@@ -99,16 +99,20 @@ class RequestFlag(IntEnum):
 # Release callback is hidden, user don't need to be exposed to
 # Triton request detail.
 # Variables should not be changed until "consumed", an callback may be
-# provided to be invoked when the request is no longer in used
+# provided to be invoked when the request is no longer in used.
+# Similarly, a response callback may be provided when response has arrived
+# or completed
 class InferenceRequest:
 
     def __init__(self,
                  name: str,
                  version: int = -1,
-                 consumed_callback: Callable = None) -> None:
+                 consumed_callback: Callable = None,
+                 response_callabck: Callable = None) -> None:
         self.name = name
         self.version = version
         self.consumed_callback = consumed_callback
+        self.response_callback = response_callabck
 
         self.request_id: str = ""
         self.timeout_ms: int = 0
@@ -189,6 +193,7 @@ class InferenceResponse:
 #         if isinstance(output, NumpyTensor):
 #             res.append(output.as_numpy())
 #         else:
+#             # May also interact with Tensor attributes directly
 #             res.append(numpy.from_dlpack(output))
 import numpy
 class NumpyTensor(Tensor):
--- a/python/_infer.py
+++ b/python/_infer.py
@@ -1,0 +1,147 @@
+from typing import Any
+from ._common import *
+from enum import IntEnum
+from collections.abc import Iterable, Union, Tuple, Mapping, Callable
+import abc
+import queue
+
+
+# [FIXME] implement dlpack interface, buffer attribute?
+class Tensor:
+
+    def __init__(self) -> None:
+        # buffer attributes..
+        # [FIXME] need to point out where exactly the buffer is managed
+        self.buffer: int = 0
+        self.byte_size: int = 0
+        self.memory_type: Tuple(MemoryType, int) = (MemoryType.CPU, 0)
+        # [FIXME] CUDA specific.. how to / should introduce CUDA dependency?
+        self.cuda_ipc_handle: Any = None
+        self.host_policy_name: str = None
+
+    def __dlpack__(self, stream=None):
+        raise "Not Implemented"
+
+    def __dlpack_device__(self):
+        raise "Not Implemented"
+
+
+class ResponseAllocator(abc.ABC):
+    _allocator: Any = None
+
+    def __init__(self) -> None:
+        super().__init__()
+        # Create Triton allocator once and for all as
+        # a Triton allocator is collection of callbacks and
+        # can be reused across inference
+        if type(self)._allocator is None:
+            # TRITONSERVER_ResponseAllocatorNew
+            raise "Not Implemented"
+
+    # [FIXME] 'response_allocator_userp' can be enclosed in the derieved
+    # allocator instance.
+    @optional_callabck
+    def start(self):
+        # is TRITONSERVER_ResponseAllocatorStartFn_t
+        raise "Not Implemented"
+
+    # [FIXME] Should tensor be returned? Or a more primal form?
+    # [FIXME] return 'buffer_userp'?
+    @abc.abstractmethod
+    def alloc(self, tensor_name: str,
+              requested_tensor: Tensor) -> Tuple[Tensor, Any]:
+        # is TRITONSERVER_InferenceTraceTensorActivityFn_t
+        raise "Not Implemented"
+
+    @abc.abstractmethod
+    def release(self, tensor: Tensor, buffer_user_object: Any = None):
+        # is TRITONSERVER_ResponseAllocatorReleaseFn_t
+        raise "Not Implemented"
+
+    @optional_callabck
+    def last_alloc(self,
+                   tensor_name: str,
+                   user_object: Any = None,
+                   buffer_user_object: Any = None) -> Tensor:
+        # is TRITONSERVER_ResponseAllocatorBufferAttributesFn_t
+        raise "Not Implemented"
+
+    # [FIXME] pre-alloc? Query the Tensor to be returned if the same parameters
+    # are used for alloc() function. This function may be called to configure
+    # internal optimization in preparation for handling the allocated tensor.
+    @optional_callabck
+    def query(self, tensor: Tensor, tensor_name: str = None) -> Tensor:
+        # is TRITONSERVER_ResponseAllocatorQueryFn_t
+        raise "Not Implemented"
+
+
+class RequestFlag(IntEnum):
+    SEQUENCE_START: int = 1
+    SEQUENCE_END: int = 2
+
+
+# Release callback is hidden, user don't need to be exposed to
+# Triton request detail.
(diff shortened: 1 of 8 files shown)


python/_trace.py
```diff
@@ -0,0 +1,108 @@
...
+# inherit from TraceReportor to the abstract functions,
+# intiantiate the reportor() object with the desire configuration
+# Detail:
+# obj(TraceReportor) ->
+#     TraceTensorNew(&trace, obj._level, obj.trace_activity,
+#         obj.trace_tensor,
+#         [](){
+#             obj.trace_release;
+#             TraceDelete();
+#             // assumping root trace is always released last..
```
oandreeva-nv, line 104: Just a note. This assumption is not correct for OpenTelemetry tracing of bls + ensemble cases

GuanLuo: In those cases, when can the `userp` be safely released? I believe within the trace release callback, it needs to figure out that the current invocation is for the last trace and then can delete `userp`.

oandreeva-nv: The BLS tracing PR adds a tracker of spawned traces, and if no more release callbacks is expected, `userp` is deleted:

https://github.com/triton-inference-server/server/pull/6063/files#diff-c05667609085c993fd282ff18d0429945c44492b95c0d01fcbeec6946fda2db2R614-R626

Feel free to add any suggestions while it is in review

Tabrizian (commented):
One issue with the current design is tied closely to the Server APIs instead of starting with the user workflow and working our way backwards to the C-API.

Ideally I think we would want the user workflow to look like this:


```
from tritonserver import Model, TritonServer, Request


server = TritonServer()
model = Model('<onnx-model-object>|<path-to-an-onnx-model>|...')
server.load(model)

request = Request(...)
response_iterator = server.infer(model, request)
async_response_iterator = await server.async_infer(model, request)
```

Tabrizian:
> I don't see the necessary of providing synchronized infer API, user can read from response iterator after async_infer return and it will block until response is ready.

I think that's a valid argument for the C-API but for Python we need to integrate that with asyncIO. Also, I think it is not just the `infer` API. We probably need `async` version for all the time-consuming Triton APIs (e.g., load/unload). I'm just wondering whether there is a way to avoid ending up with two async infer APIs (one that uses callback based and another that uses asyncIO). We could provide only provide an asyncIO based async infer API but that would make the sync code a bit awkward. For example, they'd have to run `asyncio.run(server.async_infer)`.

Regarding having multiple models with the same name, I think we can document that all the models with same name would point to the same underlying object or we could also error out if the user tries to load another model with the same name.

python/_infer.py
```diff
@@ -0,0 +1,249 @@
...
+
+    def __init__(self) -> None:
+        # tensor metadata
+        self.data_type: DataType = None
+        self.shape: Iterable[int] = None
+        
+        # buffer attributes..
+        self.buffer_address: int = 0
+        self.byte_size: int = 0
+        self.memory_type: Tuple(MemoryType, int) = (MemoryType.CPU, 0)
```
rmccorm4, line 19: Explicit `DeviceId` over `int` for memory typing?
```suggestion
        self.memory_type: Tuple(MemoryType, DeviceID) = (MemoryType.CPU, 0)
```

python/_infer.py
```diff
@@ -0,0 +1,249 @@
+from typing import Any
+from ._common import *
+from enum import IntEnum
+from collections.abc import Iterable, Union, Tuple, Mapping, Callable
+import abc
+import queue
+
+
+class Tensor:
```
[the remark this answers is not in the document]
GuanLuo: > Is it possible to have only a single Tensor class?

Can you elaborate on this? I think there will only be one `Tensor` class, to store the information needed for Triton (buffer attribute, tensor data etc.). Although the `InferenceRequest` accepts other tensor types that implements dlpack interface, internally it is converted to `Tensor` and then process with the Triton APIs. On this note I agree that we should have `Tensor.from_dlpack`. However, the conversion doesn't involve data copy so the numpy / other dlpack-compatible tensor must be valid.

The part about allocator in `Tensor` is related to the output buffer allocation. The user controls how the tensor data is allocated, that is probably where having multiple Tensor classes seems to be happening: even the `Tensor` interface is unchanged, each implementation of the allocator will have to set additional attributes in `Tensor` for allocation purpose.

python/_infer.py
```diff
@@ -0,0 +1,249 @@
...
+        while not self.exhausted():
+            yield self.get()
+
+    def exhausted(self) -> bool:
+        return self._end and self.empty()
+
+
+# [FIXME] the Triton API doesn't decouple Triton response and its output, a
+# deletion of TRITONSERVER_InferenceResponse will call allocator release to
+# release the output buffers.
```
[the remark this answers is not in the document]
GuanLuo: Is it basically 2. mentioned in the comment below this line?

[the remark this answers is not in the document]
Tabrizian: Exactly. sorry, I missed that. Resolving.

python/_infer.py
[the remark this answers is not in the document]
GuanLuo: I see what you are saying now, the `NumpyAllocator` does extend `Tensor` class to add allocator specific detail instead of directly adding attributes to a `Tensor` instance. I think this is just a different way to achieve my earlier comment about buffer allocation, users should process the output buffer via `DLPack` / `Tensor` interface which is allocator agnostic, my example usage is not appropriate as it assumes the allocator detail..

python/_infer.py
```diff
@@ -0,0 +1,249 @@
...
+# the response's tensors before assuming derived class method is available, i.e.
+# def read_output_as_numpy(response: InferenceResponse) -> Iterable[numpy.ndarray]:
+#     res = []
+#     for output in response.outputs:
+#         if isinstance(output, NumpyTensor):
+#             res.append(output.as_numpy())
+#         else:
+#             # May also interact with Tensor attributes directly
+#             res.append(numpy.from_dlpack(output))
+import numpy
```
oandreeva-nv, line 198: Can we move this import to the top?

python/_logger.py
```diff
@@ -0,0 +1,108 @@
...
+    ERROR: int = 2
+    VERBOSE: int = 3
+
+
+class LogFormat(Enum):
+    DEFAULT: int = 0
+    ISO8601: int = 1
+
+
+# [WIP] figure out "singleton"
```
oandreeva-nv, line 17: Quick question is this still a WIP and this will be resolved before merging? is it possible to elaborate `figure out` if it is left for future work?

python/_trace.py
```diff
@@ -0,0 +1,108 @@
...
+    TIMESTAMPS: int = 4
+    TENSORS: int = 8
+
+    def __str__(self) -> str:
+        # TRITONSERVER_InferenceTraceLevelString
+        raise "Not Implemented"
+
+
+# Activity related to timeline tracing
+class TraceActivity(Enum):
```
oandreeva-nv, line 18: Just an FYI, @Tabrizian is working (IIRC) on custom tracing for python backend, which introduces new `TraceActivity`: https://github.com/triton-inference-server/core/pull/172. So wee need to coordinate this at some point

python/_server.py
```diff
@@ -0,0 +1,252 @@
...
+        self.strict_readiness: bool = None
+        self.exit_on_error: bool = None
+        self.exit_timeout: int = None
+
+        # [FIXME] this option is actually not in used in core
+        # TRITONSERVER_ServerOptionsSetBufferManagerThreadCount
+        self.buffer_manager_thread_count: int = None
+
+        # Model stuff..
+        self.model_repositories: Iterable[str] = None
```
krishung5, line 73: I was wondering if it would make sense to set the model repository path as a required field.

python/_logger.py
GuanLuo: This will be resolved before merging. This PR will include the actual implementation of the wrapper.

python/_server.py
GuanLuo: As in making the paths part of init parameter? i.e. `__init__(self, model_paths)`

krishung5: Yes. I just thought that instead of doing this
```
options = Options()
options.model_repositories = ["/path/to/models"]
```
we can do
```
options = Options(["/path/to/models"])
```
given that the model repo paths are required when starting the server.

python/_infer.py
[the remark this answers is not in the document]
Tabrizian: Exactly, I'm just saying that we should try to decouple the buffer allocation logic from the tensor representation.

python/_infer.py
```diff
@@ -0,0 +1,249 @@
...
+# Variables should not be changed until "consumed", an callback may be
+# provided to be invoked when the request is no longer in used.
+# Similarly, a response callback may be provided when response has arrived
+# or completed
+class InferenceRequest:
+
+    def __init__(self,
+                 name: str,
+                 version: int = -1,
+                 consumed_callback: Callable = None,
```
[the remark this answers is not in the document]
GuanLuo: It's not needed unless the user want to know when he may modify the content of the `InferenceRequest`, in most context the input tensors. But if there is no reuse of the request and its content, then it is not necessary:
```python
input_np = np.array(...)
request = InferenceRequest(...)
request.inputs = {"input", input_np}
# async infer..
input_np[0] = ... # modifying tensor content can cause issue if inference hasn't fully consume the input
```

PR #239, opened 2023-08-04
Add recursive lookup for better ensemble disconnectivity reporting

GuanLuo:


dyastremsky (dismissed):
Nice change! Very clean.

Approved your related tests here too: https://github.com/triton-inference-server/server/pull/6149

src/model_config_utils.cc
```diff
@@ -233,10 +239,31 @@ ValidateEnsembleSchedulingConfig(const inference::ModelConfig& config)
...
     if (!it->second.ready) {
-      return Status(
-          Status::Code::INVALID_ARG, "output '" + output.name() +
-                                         "' for ensemble '" + config.name() +
-                                         "' is not written");
+      std::string error_message = "output '" + output.name() +
+                                  "' for ensemble '" + config.name() +
+                                  "' is not written";
+
+      // recurrsively check 'prev_nodes' for the source of not-ready state
```
oandreeva-nv, line 246: Is there a possibility of being stuck in a cycle during this check?

GuanLuo: Yes and fixed.

dyastremsky (approved):

oandreeva-nv (approved):
LGTM, thanks! I am curious, if this is an issue, then would it make sense to provide a `sanity check` tool for ensembles ?

GuanLuo:
I think we could, it is relatively easy to extract the static checking part of the ensemble config.

dyastremsky:
> I think we could, it is relatively easy to extract the static checking part of the ensemble config.

This might be excessive, but it might even be valuable to provide some sort of visualization tool as part of this. Like one that shows a visual representation of the graph produced by Triton for the ensemble. This might be less trivial, but there are likely libraries that could make this easier.

oandreeva-nv:
I would be interested in implementing this

rmccorm4:
I believe the developer of [Netron](https://github.com/lutzroeder/netron) is quite open to contribution/collaboration, and we use it often for ONNX and some TensorRT stuff - it might be worth adding a config as an acceptable format.

PR #241, opened 2023-08-10
Use preserve_ordering config field for oldest strategy sequence scheduler

rmccorm4:
Common: https://github.com/triton-inference-server/common/pull/97
Core: https://github.com/triton-inference-server/core/pull/241
Server: https://github.com/triton-inference-server/server/pull/6185

### DIFF
--- a/src/sequence_batch_scheduler.cc
+++ b/src/sequence_batch_scheduler.cc
@@ -1805,9 +1805,6 @@ OldestSequenceBatch::OldestSequenceBatch(
 
   // TODO: Provide appropriate request_cache_enable flag when caching
   // is enabled for sequence models.
-  // TODO: Check if preserve ordering is manually specified or not,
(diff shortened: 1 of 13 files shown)


src/sequence_batch_scheduler.cc
```diff
@@ -1805,11 +1805,14 @@ OldestSequenceBatch::OldestSequenceBatch(
 
   // TODO: Provide appropriate request_cache_enable flag when caching
   // is enabled for sequence models.
+  // TODO: Check if preserve ordering is manually specified or not,
+  // will give default value of true if unspecified to keep current behavior.
+  // May need to wrap the field in a message to allow checking for emptiness.
```
rmccorm4, line 1810: Remove if changing default to false is OK, which it sounds like is the case.

tanmayv25 (approved):

PR #243, opened 2023-08-15
updating log message with input name

nnshah1:
based on feedback - adding the input name when logging mismatched data type.

Message is now:

inference input 'INPUT' data type is 'DATATYPE_1', but model 'MODEL' expects 'DATATYPE_2'

Open to other phrasing.

### DIFF
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -807,7 +807,9 @@ InferenceRequest::Normalize()
       if (!has_one_element) {
         return Status(
             Status::Code::INVALID_ARG, LogRequest() +
-                                           "For BYTE datatype raw input, the "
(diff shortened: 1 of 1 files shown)


Tabrizian (dismissed):

tanmayv25 (dismissed):

tanmayv25:
@nnshah1  there are couple of more locations were we are not including the input name in log message.

nnshah1:
> 

I added one more. There is more potential work to make messages more uniform (some give the model name, some do not) - but I think we'd do that separately..

Tabrizian (approved):

tanmayv25 (approved):

PR #245, opened 2023-08-17
Document 'userp' lifecycle on TRITONSERVER_InferenceRequestSetResponseCallback

GuanLuo:


rmccorm4 (approved):

include/triton/core/tritonserver.h
```diff
@@ -1329,6 +1329,10 @@ TRITONSERVER_InferenceRequestSetReleaseCallback(
 /// tensors included in responses that are produced for this
 /// request. The response callback is called to return response
 /// objects representing responses produced for this request.
+/// Typically 'response_allocator_userp' and 'response_userp' will no
```
nnshah1, line 1332: Is it possible to phrase it to remove 'Typically'?  Are there scenarios where they will be referenced? 
 Something like:

```suggestion
/// User supplied pointers, `response_allocator_userp` and `response_userp` for this request will not be referenced after a final response is sent (TRITONSERVER_RESPONSE_COMPLETE_FINAL). Users are free to release them at this point. 
```

PR #246, opened 2023-08-23
Add queue size (pending request count) metric per-model

rmccorm4:
# Overview

Server/Tests PR: https://github.com/triton-inference-server/server/pull/6233

For simplicity, the definition of "queued" requests in this PR is **any** requests that are sitting somewhere in Triton core and have not yet been picked up by a backend model instance thread for execution.

This PR adds a new prometheus metric "nv_inference_pending_request_count" per-model:
```
# HELP nv_inference_pending_request_count Instantaneous number of pending requests awaiting execution per-model.
# TYPE nv_inference_pending_request_count gauge
nv_inference_pending_request_count{model="default",version="1"} 17
nv_inference_pending_request_count{model="dynamic",version="1"} 3
nv_inference_pending_request_count{model="sequence_direct",version="1"} 42
nv_inference_pending_request_count{model="sequence_oldest",version="1"} 26
```

## How It Works

Currently the code is reporting `count++` on every `request->Scheduler->Enqueue(request)`, and `count--` when a request is about to be passed to a backend in `TritonModelInstance::Execute(requests)`, both done via state transition. 

## State Transitions

In the event that a request encounters an error anywhere along the way within core, some state transitions were added to help detect if the count should be updated when the request is getting destructed.

These state transitions may also be helpful in the near future for handling a potential `CANCELLED` state.

Currently the `RELEASED` state is just for completeness to have a state after execution has finished. It's not doing very much right now other than some possible error checking that we're not moving a request from RELEASED state to some other state, as that is probably a sign of invalid behavior or code smell for now.

## Performance Implications

We use `prometheus-cpp` for our metric updates, and there is a thorough benchmark of each operation [here](https://github.com/jupp0r/prometheus-cpp#benchmarks). For queue size, we are using a `gauge` which can both increase and decrease in value, compared to monotonic `counter`s. The relevant operations are timed here:
```
BM_Gauge_Increment                                  12 ns         12 ns   51511873
BM_Gauge_Decrement                                  12 ns         12 ns   56831098
```
As shown, the `Increment(value)` and `Decrement(value)` operations should be quite fast, adding a 24 nanosecond (12ns increment +12ns decrement) cost per request for the benefit of instant queue size updates for client-side load balancing/tracking.

I'll do some perf analyzer runs before/after as a sanity check.

## Alternatives Considered

One alternative considered instead of updating on every queue size change:

Tracking the queue size internally, but only updating the metric on a certain `--metrics-interval-ms` with a polling thread, similar to the GPU/CPU metrics today.

However, doing so would likely require exposing a lot of data structures from ModelRepositoryManager / ModelLifecycle / Server to the Metrics class to iterate over all models/schedulers and get their current reported queue sizes. This would probably (a) be much more involved code, (b) contend with the locks around models/modelInfos used for repository management / request creation, and (c) update the metric with less precision. So ultimately, I decided for the live updates since the cost (~24ns) appeared minimal, and independent of backend execution.

## Tests/Docs

Tests/docs: https://github.com/triton-inference-server/server/pull/6233

Local testing for all scheduler types is looking pretty good so far.

I intend to add some detailed docs to [metrics.md](https://github.com/triton-inference-server/server/blob/main/docs/user_guide/metrics.md) detailing the definition of "queue size" similar to above. Let me know if there's anything you'd want to see in the docs called out specifically.

---

# Archived Notes

The following is an outdated payload-based approach compared to current request-based tracking approach. I'll leave it in as it has some nice details for future reference, but is not very relevant to the current PR.

<details>
  <summary>Expand</summary>

### Queues Involved Per Scheduler

Depending on the type of scheduler used by the model, the definition of "queue" is varied. Additionally, all schedulers (not counting ensemble) currently go through RateLimiter Payloads, and queue up some payloads there in addition to their own notions of queues.

**Default Scheduler:**
- No scheduler queue, immediately pushes all requests to payload queues in Rate Limiter (per-model: "general/model queue")

**Dynamic Batch Scheduler:**
- Internal queue for forming batches
- Pushes batches into payload queues in Rate Limiter (per-model: "general/model queue")

**Sequence Batch Scheduler Direct:**
- Sequence Backlogs
- Rate Limiter Queues (per-model-instance: "specific queues")

**Sequence Batch Scheduler Oldest:**
- Sequence Backlogs
- Dynamic Batch Scheduler queue
- Rate Limiter Queues (per-model-instance; "specific queues")

**Ensemble Scheduler:**
- No scheduler queue, immediately schedules requests to composing model schedulers, following per-scheduler logic above
- Currently, the ensemble level "queue size" will just remain zero, and be reflected in the composing models. It may be possible to improve this to reflect the sum of all composing model queues, but I left that out for now.

### En-queueing Requests

To simplify the logic around tracking and syncing all subsequent data structures involved for holding "queued" requests, I decided to simply add 1 to the "queue size" when hitting the `Scheduler->Enqueue(request)` function. 

### De-queueing Requests

Similarly, since all schedulers ultimately end up hitting the Rate Limiter payloads before execution in a backend model instance, I decided to take advantage of the `Payload::OnRelease` callback that is called when a backend picks up a payload for execution. Since these are exposed to the schedulers via `payload->AddInternalReleaseCallback(callback)`, it allows for a nice separation of scheduler-specific logic without going through and modifying the existing RateLimiter/Payload logic.

Depending on the scheduler used and whether batching is supported or not, the number of requests in each payload (or in other words the number of requests that are de-queued at once) can vary up to the defined `max_batch_size`. 

**EDIT**: The payload appears to not actually be released until the model instance finishes executing the request (not counting [merge_payloads](https://github.com/triton-inference-server/core/blob/fc2ada26cfbabd17c4f65058d4ee1531debc3495/src/rate_limiter.cc#L351-L353)), whereas I originally thought it was released as soon as the backend picked it up/started executing. So the metric includes currently executing request count as well. If we transition to the idea of adding an "OnExecute" callback to payloads, this fixes the problem of not counting requests in-execution, but it introduces a lot of special handling required if requests exit early for an errors before getting to the execution point.

</details>

Closes https://github.com/triton-inference-server/server/issues/5927.

### DIFF
--- a/src/backend_model_instance.cc
+++ b/src/backend_model_instance.cc
@@ -753,6 +753,7 @@ TritonModelInstance::TritonBackendThread::BackendThread()
         model_instances_, &payload);
     NVTX_RANGE(nvtx_, "BackendThread " + name_);
     payload->Execute(&should_exit);
+    // LOG_STATUS_ERROR(payload->Wait(), "Failed to execute payload");
     model_instances_.push_back(payload->GetInstance());
     // Release the payload to the RateLimiter
     model_->Server()->GetRateLimiter()->PayloadRelease(payload);
--- a/src/backend_model_instance.cc
+++ b/src/backend_model_instance.cc
@@ -534,7 +534,7 @@ TritonModelInstance::GenerateWarmupData()
   return Status::Success;
 }
 
-void
+Status
 TritonModelInstance::Schedule(
     std::vector<std::unique_ptr<InferenceRequest>>&& requests,
     const std::function<void()>& OnCompletion)
@@ -545,16 +545,17 @@ TritonModelInstance::Schedule(
   triton_requests.clear();
   for (auto& r : requests) {
     // Load the input states for the inference request.
-    r->LoadInputStates();
+    RETURN_IF_ERROR(r->LoadInputStates());
     // Set request state to signtify that request is no longer pending.
-    r->SetState(InferenceRequest::State::EXECUTING);
+    RETURN_IF_ERROR(r->SetState(InferenceRequest::State::EXECUTING));
     triton_requests.push_back(
         reinterpret_cast<TRITONBACKEND_Request*>(r.release()));
   }
 
   Execute(triton_requests);
-
   OnCompletion();
+
+  return Status::Success;
 }
 
 Status
--- a/src/backend_model_instance.cc
+++ b/src/backend_model_instance.cc
@@ -569,8 +569,7 @@ TritonModelInstance::PrepareRequestsOrRespond(
 
 Status
 TritonModelInstance::Schedule(
-    std::vector<std::unique_ptr<InferenceRequest>>&& requests,
-    const std::function<void()>& OnCompletion)
+    std::vector<std::unique_ptr<InferenceRequest>>&& requests)
 {
   // Prepare requests for execution, respond to requests if any error occur.
   RETURN_IF_ERROR(PrepareRequestsOrRespond(requests));
@@ -585,8 +584,6 @@ TritonModelInstance::Schedule(
   }
 
   Execute(triton_requests);
-  OnCompletion();
-
   return Status::Success;
 }
 
--- a/src/backend_model_instance.cc
+++ b/src/backend_model_instance.cc
@@ -26,6 +26,8 @@
 
 #include "backend_model_instance.h"
 
+#include "status.h"
+
 #ifndef _WIN32
 #include <sys/resource.h>
 #include <sys/syscall.h>
@@ -535,19 +537,49 @@ TritonModelInstance::GenerateWarmupData()
 }
 
 Status
+TritonModelInstance::PrepareRequestsForExecution(
+    std::vector<std::unique_ptr<InferenceRequest>>& requests)
+{
+  for (auto& r : requests) {
+    // Load the input states for the inference request.
+    RETURN_IF_ERROR(r->LoadInputStates());
+    // Set request state to signify that request is no longer pending.
+    RETURN_IF_ERROR(r->SetState(InferenceRequest::State::EXECUTING));
+  }
+
+  return Status::Success;
+}
+
+Status
+TritonModelInstance::PrepareRequestsOrRespond(
+    std::vector<std::unique_ptr<InferenceRequest>>& requests)
+{
+  auto status = PrepareRequestsForExecution(requests);
+  // If any errors occurred, respond with error for each request.
+  if (!status.IsOk()) {
+    for (auto& r : requests) {
+      InferenceRequest::RespondIfError(r, status, true /* release_requests */);
+    }
+    // Log a single error for batch of requests for better visibility
+    LOG_STATUS_ERROR(status, "Requests failed pre-execution checks");
+  }
+
+  return status;
+}
+
+Status
 TritonModelInstance::Schedule(
     std::vector<std::unique_ptr<InferenceRequest>>&& requests,
     const std::function<void()>& OnCompletion)
 {
+  // Prepare requests for execution, respond to requests if any error occur.
+  RETURN_IF_ERROR(PrepareRequestsOrRespond(requests));
+
   // Use a thread local vector to avoid needing to malloc each
   // time an inference is run.
   thread_local std::vector<TRITONBACKEND_Request*> triton_requests(1024);
   triton_requests.clear();
   for (auto& r : requests) {
-    // Load the input states for the inference request.
-    RETURN_IF_ERROR(r->LoadInputStates());
-    // Set request state to signtify that request is no longer pending.
-    RETURN_IF_ERROR(r->SetState(InferenceRequest::State::EXECUTING));
     triton_requests.push_back(
         reinterpret_cast<TRITONBACKEND_Request*>(r.release()));
   }
@@ -753,7 +785,6 @@ TritonModelInstance::TritonBackendThread::BackendThread()
         model_instances_, &payload);
     NVTX_RANGE(nvtx_, "BackendThread " + name_);
     payload->Execute(&should_exit);
-    // LOG_STATUS_ERROR(payload->Wait(), "Failed to execute payload");
     model_instances_.push_back(payload->GetInstance());
     // Release the payload to the RateLimiter
     model_->Server()->GetRateLimiter()->PayloadRelease(payload);
--- a/src/backend_model_instance.cc
+++ b/src/backend_model_instance.cc
@@ -546,6 +546,8 @@ TritonModelInstance::Schedule(
   for (auto& r : requests) {
     // Load the input states for the inference request.
     r->LoadInputStates();
+    // Set request state to signtify that request is no longer pending.
+    r->SetState(InferenceRequest::State::EXECUTING);
     triton_requests.push_back(
         reinterpret_cast<TRITONBACKEND_Request*>(r.release()));
   }
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -140,8 +140,8 @@ InferenceRequest::SetState(InferenceRequest::State new_state)
   // Generate error when called rather than copying it into every case below.
   const auto generate_error = [&]() {
     std::stringstream ss;
-    ss << "Invalid request state transition from " << state_ << " to "
-       << new_state;
+    ss << LogRequest() << "Invalid request state transition from " << state_
+       << " to " << new_state;
     return Status(Status::Code::INVALID_ARG, ss.str());
   };
 
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -106,7 +106,8 @@ InferenceRequest::InferenceRequest(
     : needs_normalization_(true), model_raw_(model),
       requested_model_version_(requested_model_version), flags_(0),
       correlation_id_(0), batch_size_(0), timeout_us_(0), collect_stats_(true),
-      state_(InferenceRequest::State::INITIALIZED)
+      state_(InferenceRequest::State::INITIALIZED),
+      decrement_pending_count_(false)
 {
   SetPriority(0);
 }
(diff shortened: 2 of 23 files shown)


rmccorm4:
Edit: The payload appears to not actually be released until the model instance **finishes** executing the request, whereas I originally thought it was released as soon as the backend picked it up/started executing. Looking into this more.

tanmayv25:
I think you'd need to decrement you counters here: https://github.com/triton-inference-server/core/blob/main/src/payload.cc#L201
Within payload execute for INFER_RUN: 
`Payload::Execute(bool* should_exit)`

A separate callback would be fine as it would avoid exporting the reporter to payload class.

> Essentially at the time of the EnqueuePayload() the number of requests don't always match the final number of requests in the payload at release time.

How are we dealing with Intiialization and WarmUp payloads? I wonder if we have additional decrements in the counts.

Lastly, can we rename "nv_inference_queue_size" metric to represent something more comprehensive to what it is actually tracking? For example, "nv_inference_pending_requests_core" or something similar.

src/sequence_batch_scheduler.cc
```diff
@@ -128,8 +145,8 @@ SequenceBatchScheduler::Create(
   const size_t model_batch_size = std::max(1, config.max_batch_size());
   sched->seq_slot_cnt_ = model_batch_size;
   if (config.sequence_batching().has_oldest()) {
-    sched->seq_slot_cnt_ =
-        config.sequence_batching().oldest().max_candidate_sequences();
+    sched->seq_slot_cnt_ = std::max(
+        1, config.sequence_batching().oldest().max_candidate_sequences());
```
rmccorm4, line 149: NOTE: This is just a Quality-Of-Life update because I noticed the default max_candidate_sequences would be zero if left unspecified with minimal config, and no error is reported to alert user:
```
sequence_batching {
  oldest { }
}
```
And the model would never run any requests.

Tabrizian: I think we need to set it to `model_batch_size` if it is zero.

Tabrizian (commented):
Other than one comment, LGTM! Very clean implementation!

src/infer_request.cc
```diff
@@ -104,11 +105,87 @@ InferenceRequest::InferenceRequest(
...
+  // No-op if this is the current state.
+  if (state_ == new_state) {
+    return Status::Success;
+  }
+
+  // Define state transitions
+  std::stringstream ss;
+  ss << LogRequest() << "Invalid request state transition from " << state_
+     << " to " << new_state;
+  const auto error_status = Status(Status::Code::INVALID_ARG, ss.str());
```
GuanLuo, line 137: Generate only when error needs to be returned.

src/infer_request.cc
```diff
@@ -104,11 +105,87 @@ InferenceRequest::InferenceRequest(
...
+    case InferenceRequest::State::STARTED: {
+      if (new_state != InferenceRequest::State::EXECUTING) {
+        return error_status;
+      }
+      state_ = new_state;
+      DecrementPendingRequestCount();
+      break;
+    }
+    case InferenceRequest::State::EXECUTING: {
+      if (new_state != InferenceRequest::State::RELEASED) {
```
GuanLuo, line 157: If valid transition is linear, you may measure if the transition is correct by distance.

src/model.h
```diff
@@ -225,6 +225,12 @@ class Model {
 
   uint64_t MaxPriorityLevel() const { return max_priority_level_; }
 
+  // Returns the scheduler's metric reporter
+  std::shared_ptr<MetricModelReporter> MetricReporter()
+  {
+    return scheduler_->MetricReporter();
```
GuanLuo, line 231: Should metric reporter be owned by the model?

src/metric_model_reporter.h
```diff
@@ -81,6 +81,12 @@ class MetricModelReporter {
   const MetricReporterConfig& Config();
   // Lookup counter metric by name, and increment it by value if it exists.
   void IncrementCounter(const std::string& name, double value);
+  // Lookup gauge metric by name. Return gauge if found, nullptr otherwise.
+  prometheus::Gauge* GetGauge(const std::string& name);
```
GuanLuo, line 85: private?

src/backend_model_instance.cc
```diff
@@ -546,6 +546,8 @@ TritonModelInstance::Schedule(
   for (auto& r : requests) {
     // Load the input states for the inference request.
     r->LoadInputStates();
+    // Set request state to signtify that request is no longer pending.
+    r->SetState(InferenceRequest::State::EXECUTING);
```
GuanLuo, line 550: Error checking?

src/metric_model_reporter.cc
```diff
@@ -178,12 +186,13 @@ void
...
     const std::map<std::string, std::string>& labels)
 {
-  // Always setup these counters, regardless of config
+  // Always setup these inference request metrics, regardless of config
   counter_families_["inf_success"] = &Metrics::FamilyInferenceSuccess();
   counter_families_["inf_failure"] = &Metrics::FamilyInferenceFailure();
   counter_families_["inf_count"] = &Metrics::FamilyInferenceCount();
   counter_families_["inf_exec_count"] =
       &Metrics::FamilyInferenceExecutionCount();
+  gauge_families_[kPendingRequestMetric] = &Metrics::FamilyInferenceQueueSize();
```
GuanLuo, line 195: Separate initialize function?

src/infer_request.cc
rmccorm4: Was thinking something similar. In the back of my mind I've been wondering if transitioning to `RELEASED` state from any other state should be relaxed. Ex: If some C API user creates a request and then Releases it without using it, they'd get an error (though just a log error right now). I'm wondering if it makes sense to allow Release transition from anywhere.

src/metric_model_reporter.cc
rmccorm4: Good catch, I must've been too tired when writing this :sweat_smile:

src/model.h
rmccorm4: It think it makes a bit more sense to promote ownership from the scheduler to the model, and it will remove reporter from the scheduler interace. I'm not that opinionated on promoting it at this time, but I'll try it out.

rmccorm4: Promoted reporter to model, and realized I could more easily remove the the `response_cache_enable` that was getting passed around everywhere: https://github.com/triton-inference-server/core/pull/246/commits/b9527b5ba46a17fcd0d86fc51e653b452954e24e

rmccorm4:
Added fixes for the null_request case (coming from sequence batcher mostly) and also fixed error handling to actually respond to requests with errors prior to execution rather than leaving the requests hanging.

src/backend_model_instance.cc
```diff
@@ -534,27 +536,58 @@ TritonModelInstance::GenerateWarmupData()
...
-    r->LoadInputStates();
-    // Set request state to signtify that request is no longer pending.
-    r->SetState(InferenceRequest::State::EXECUTING);
     triton_requests.push_back(
         reinterpret_cast<TRITONBACKEND_Request*>(r.release()));
   }
 
   Execute(triton_requests);
-
   OnCompletion();
```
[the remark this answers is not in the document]
rmccorm4: If we keep the error checking / early return above, what do you think of defining and using a [ScopedDefer](https://github.com/triton-inference-server/python_backend/blob/6f369ef10312ad3db0aef73df5b1138b8467cf14/src/scoped_defer.cc#L30-L50)() like in Python Backend for this?

```cpp
Status
TritonModelInstance::Schedule(
    std::vector<std::unique_ptr<InferenceRequest>>&& requests,
    const std::function<void()>& OnCompletion)
{
  // Ensure completion callback is always called regardless of error
  auto deferred_callback = []{ OnCompletion(); }
  ScopedDefer _(deferred_callback);

  // Prepare requests for execution, respond to requests if any error occur.
  RETURN_IF_ERROR(PrepareRequestsOrRespond(requests));
 
  ...
  Execute(triton_requests);
  return Status::Success
}

```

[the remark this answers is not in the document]
rmccorm4: Actually, the callback is associated with the payload. The expectations for when it's called I don't think is well documented, but assuming it's intended to be always be called after Execute() as it currently is -- wouldn't it make more sense to call the callback closer to Payload code, such as at end of [payload->Execute()](https://github.com/triton-inference-server/core/blob/8807e87027cb8a38643984c32c71a6e43d4a9c23/src/payload.cc#L214)?

src/backend_model_instance.cc
```diff
@@ -534,27 +536,58 @@ TritonModelInstance::GenerateWarmupData()
...
+  return Status::Success;
+}
+
+Status
+TritonModelInstance::PrepareRequestsOrRespond(
+    std::vector<std::unique_ptr<InferenceRequest>>& requests)
+{
+  auto status = PrepareRequestsForExecution(requests);
+  // If any errors occurred, respond with error for each request.
+  if (!status.IsOk()) {
```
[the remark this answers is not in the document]
rmccorm4: Resolved offline, returning an error here is only for an exceptional case and it not expected. However, in the event some critical error happens at this stage such as a truly invalid state transition or future potential errors, this is a good place to handle it.

src/backend_model_instance.cc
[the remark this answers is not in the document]
rmccorm4: After discussing with Guan and Tanmay, I added some clarifying comments on the existing `payload->Callback()` usage. It may be an over-optimization at the moment compared to Release callbacks, but Tanmay and I decided not to change it for now. Tanmay may look into cleaning these callbacks up a bit in the future.

As for calling the callback after Execute regardless of error, I moved the callback into `payload->Execute()` to keep the payload-specific callback with the payload code. No other code currently calls `instance->Schedule()` with any callbacks, so we can simplify the instance logic there.

GuanLuo (approved):

PR #247, opened 2023-08-24
Backend Shape and Datatype Access API

fpetrini15:
API to obtain the shape and datatype of a particular response output.

### DIFF
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -1569,6 +1569,20 @@ TRITONBACKEND_ISPEC TRITONSERVER_Error* TRITONBACKEND_ModelBatchInitialize(
 TRITONBACKEND_ISPEC TRITONSERVER_Error* TRITONBACKEND_ModelBatchFinalize(
     void* userp);
 
+/// Accessor function to obtain the shape and data type of a particular output.
+///
+/// \param response The response.
+/// \param name The name of the output tensor to extract shape and data
+/// type information.
+/// \param datatype Returns the tensor datatype.
+/// \param shape Returns the tensor shape.
+/// \param dim_count Returns the number of dimensions in the tensor shape.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONBACKEND_ISPEC TRITONSERVER_Error* TRITONBACKEND_OutputShapeAndDType(
+    TRITONBACKEND_Response* response, const char* name,
+    TRITONSERVER_DataType* datatype, const int64_t** shape,
+    uint32_t* dim_count);
+
 #ifdef __cplusplus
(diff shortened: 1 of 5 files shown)


include/triton/core/tritonbackend.h
```diff
@@ -1569,6 +1569,60 @@ TRITONBACKEND_ISPEC TRITONSERVER_Error* TRITONBACKEND_ModelBatchInitialize(
...
+/// \param memory_type_id Returns the memory type id of the data.
+/// \param userp The user-specified value associated with the buffer
+/// in TRITONSERVER_ResponseAllocatorAllocFn_t.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONBACKEND_ISPEC TRITONSERVER_Error*
+TRITONBACKEND_InferenceResponseOutputByName(
+    TRITONBACKEND_Response* response, const char* name,
+    TRITONSERVER_DataType* datatype, const int64_t** shape, uint64_t* dim_count,
+    const void** base, size_t* byte_size, TRITONSERVER_MemoryType* memory_type,
+    int64_t* memory_type_id, void** userp);
```
Tabrizian, line 1596: I think the userp should be probably hidden from the backend APIs. @GuanLuo what do you think?

Tabrizian (commented):
Looks good! One minor comment for discussion.

include/triton/core/tritonbackend.h
```diff
@@ -1569,6 +1569,60 @@ TRITONBACKEND_ISPEC TRITONSERVER_Error* TRITONBACKEND_ModelBatchInitialize(
 TRITONBACKEND_ISPEC TRITONSERVER_Error* TRITONBACKEND_ModelBatchFinalize(
     void* userp);
 
+/// Get all information about an output tensor by name.  The tensor data is
```
Tabrizian, line 1572: You also need to increment the API version at the top of the file.

fpetrini15: Good catch! I must've lost this change transferring from my test branch.

include/triton/core/tritonbackend.h
fpetrini15: Spoke with @GuanLuo, he said that there is already a function to obtain the buffer attributes so we should not only remove `userp`, but the whole list of buffer attributes including `byte_size`, `memory_type`, and `memory_type_id`. Will refactor.

GuanLuo: Yes, the `BufferAttributes` is supposed to be the struct to group tensor metadata like those.

Tabrizian (dismissed):

rmccorm4 (approved):

PR #248, opened 2023-08-28
[DO NOT MERGE] Python binding of Triton server C API

GuanLuo:
This PR provides the low level binding for Python user to interact with Triton library within the same process, however, this binding is not intended for Python user to use directly but to enable the development of the [Python wrapper](https://github.com/triton-inference-server/core/pull/238). Python user should use the wrapper mentioned because it encapsualtes many concepts that are not intuitive to Python user but inevitably leaked into the binding.

Most part of the binding is straight forward mapping of the C API, however, the `PyResponseAllocator`, `PyTrace`, `PyInferenceRequest` and `PyInferenceResponse` requires extra attention because they are objects that involves callbacks and ownership transfer. Also, below is the snippet from `.cc` file to help reviewers navigate things that are different from the C equivalent of the binding.

```C++
// This binding is merely used to map Triton C API into Python equivalent,
// and therefore, the naming will be the same as the one used in corresponding
// sections. However, there are a few exceptions to better transit to Python:
// Structs:
//  * Triton structs are encapsulated in a thin wrapper to isolate raw pointer
//    operations which is not supported in pure Python. A thin 'PyWrapper' base
//    class is defined with common utilities
//  * Trival getters and setters are grouped to be a Python class property.
//    However, this creates asymmetry that some APIs are called like function
//    while some like member variables. So I am open to expose getter / setter
//    if it may be more intuitive.
//  * The wrapper is only served as communication between Python and C, it will
//    be unwrapped when control reaches C API and the C struct will be wrapped
//    when control reaches Python side. Python binding user should respect the
//    "ownership" and lifetime of the wrapper in the same way as described in
//    the C API. Python binding user must not assume the same C struct will
//    always be referred through the same wrapper object.
// Enums:
//  * In C API, the enum values are prefixed by the enum name. The Python
//    equivalent is an enum class and thus the prefix is removed to avoid
//    duplication, i.e. Python user may specify a value by
//    'TRITONSERVER_ResponseCompleteFlag.FINAL'.
// Functions / Callbacks:
//  * Output parameters are converted to return value. APIs that return an error
//    will be thrown as an exception. The same applies to callbacks.
//  ** Note that in the C API, the inference response may carry an error object
//     that represent an inference failure. The equivalent Python API will raise
//     the corresponding exception if the response contains error object.
//  * The function parameters and return values are exposed in Python style,
//    for example, object pointer becomes py::object, C array and length
//    condenses into Python array.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -415,3 +415,25 @@ export(
 )
 
 export(PACKAGE TritonCore)
+
+FetchContent_Declare(
+  pybind11
+  GIT_REPOSITORY "https://github.com/pybind/pybind11"
+  # COMMIT ID for v2.10.0
+  GIT_TAG "aa304c9c7d725ffb9d10af08a3b34cb372307020"
+  GIT_SHALLOW ON
+)
+FetchContent_MakeAvailable(pybind11)
+set(
+  PYTHON_BINDING_SRCS
+  python/binding/tritonserver_pybind.cc
+)
+
+pybind11_add_module(triton_bindings SHARED ${PYTHON_BINDING_SRCS})
+target_link_libraries(
+  triton_bindings
+  PRIVATE
+  triton-core-serverapi           # from repo-core
+  triton-core-serverstub          # from repo-core
+)
+target_compile_features(triton_bindings PRIVATE cxx_std_17)
--- a/python/binding/tritonserver_pybind.cc
+++ b/python/binding/tritonserver_pybind.cc
@@ -298,6 +298,9 @@ class PyResponseAllocator
   DESTRUCTOR_WITH_LOG(
       PyResponseAllocator, TRITONSERVER_ResponseAllocatorDelete);
 
+  // Callback resource that holds Python user provided buffer and
+  // Triton C callback wrappers. This struct will be used for both
+  // 'allocator_userp' and 'buffer_userp'
   struct CallbackResource {
     CallbackResource(py::object a, py::object uo)
         : allocator(a), user_object(uo)
@@ -382,7 +385,7 @@ class PyResponseAllocator
     catch (py::error_already_set& ex) {
       err = CreateTRITONSERVER_ErrorFrom(ex);
     }
-    // Done with CallbackResource
+    // Done with CallbackResource associated with this buffer
     delete cr;
     return err;
   }
@@ -558,7 +561,12 @@ class PyTrace : public PyWrapper<struct TRITONSERVER_InferenceTrace> {
     std::set<uintptr_t> seen_traces;
   };
 
-  // Use internally when interacting with C APIs that takes ownership
+  // Use internally when interacting with C APIs that takes ownership,
+  // this function will also release the ownership of the callback resource
+  // because once the ownership is transferred, the callback resource
+  // will be accessed in the callback pipeline and should not be tied to the
+  // PyWrapper's lifecycle. The callback resource will be released in the
+  // Triton C callback wrapper.
   struct TRITONSERVER_InferenceTrace* Release()
   {
     owned_ = false;
@@ -705,6 +713,7 @@ class PyInferenceResponse
         : complete_fn(c), allocator_resource(a), user_object(u)
     {
     }
+    CompleteFn complete_fn;
     // During 'TRITONSERVER_InferenceRequestSetResponseCallback', a
     // PyResponseAllocator::CallbackResource is allocated and passed as
     // 'response_allocator_userp', which is used during any output buffer
@@ -716,7 +725,6 @@ class PyInferenceResponse
     // responses to be generated and so does output allocation, therefore
     // 'allocator_resource' may be released as part of releasing
     // 'PyInferenceResponse::CallbackResource'
-    CompleteFn complete_fn;
     PyResponseAllocator::CallbackResource* allocator_resource;
     py::object user_object;
   };
@@ -854,8 +862,11 @@ class PyInferenceRequest
 
 
   // Use internally when interacting with C APIs that takes ownership,
-  // this function releases the ownership of the C object and the callback
-  // resources tracked.
+  // this function will also release the ownership of the callback resource
+  // because once the ownership is transferred, the callback resource
+  // will be accessed in the callback pipeline and should not be tied to the
+  // PyWrapper's lifecycle. The callback resource will be released in the
+  // Triton C callback wrapper.
   struct TRITONSERVER_InferenceRequest* Release()
   {
     owned_ = false;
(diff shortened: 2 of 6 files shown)


python/binding/tritonserver_pybind.cc
```diff
@@ -0,0 +1,1962 @@
...
+    ThrowIfError(TRITONSERVER_InferenceResponseOutputClassificationLabel(
+        triton_object_, index, class_index, &val));
+    return val;
+  }
+};
+
+// forward declaration
+class PyServer;
+
+class PyInferenceRequest
```
Tabrizian, line 845: Is it possible to breakdown this file to one class per file (e.g. `PyInferenceRequest`, `PyServerOptions`, ...)?

python/binding/tritonserver_pybind.cc
```diff
@@ -0,0 +1,1962 @@
...
+  // requires a function wrapper to generalize the destructor.
+
+  // Use internally to get the pointer of the underlying Triton object
+  TritonStruct* Ptr() { return triton_object_; }
+
+  DISALLOW_COPY_AND_ASSIGN(PyWrapper);
+
+ protected:
+  TritonStruct* triton_object_{nullptr};
+  bool owned_{false};
```
Tabrizian, line 191: I was wondering whether it is possible to do the same thing with `shared_ptr` and custom deleters. For example, 

```python
class PyParameter {
  ...
  std::shared_ptr<TRITONSERVER_Parameter, TRITONSERVER_ParameterDelete> pointer_;
};
```

We would have to allow COPY_AND_ASSIGN for this to work.

python/binding/tritonserver_pybind.cc
```diff
@@ -0,0 +1,1962 @@
...
+      .def("metrics", &PyServer::Metrics)
+      .def("infer_async", &PyServer::InferAsync);
+
+  // TRITONSERVER_MetricFamily
+  py::class_<PyMetricFamily>(m, "TRITONSERVER_MetricFamily")
+      .def(py::init<
+           TRITONSERVER_MetricKind, const std::string&, const std::string&>());
+
+  // TRITONSERVER_Metric
+  py::class_<PyMetric>(m, "TRITONSERVER_Metric")
```
Tabrizian, line 1953: Perhaps remove the "TRITONSERVER_" prefix from the python bindings names?

python/binding/tritonserver_pybind.cc
```diff
@@ -0,0 +1,1962 @@
...
+    : public PyWrapper<struct TRITONSERVER_ResponseAllocator> {
+ public:
+  DESTRUCTOR_WITH_LOG(
+      PyResponseAllocator, TRITONSERVER_ResponseAllocatorDelete);
+
+  // Callback resource that holds Python user provided buffer and
+  // Triton C callback wrappers. This struct will be used for both
+  // 'allocator_userp' and 'buffer_userp'
+  struct CallbackResource {
+    CallbackResource(py::object a, py::object uo)
```
Tabrizian, line 305: Use const&

python/binding/tritonserver_pybind.cc
GuanLuo: I would like to keep the name align with the C API, but I can change it if the majority prefer a concise naming

python/binding/tritonserver_pybind.cc
GuanLuo: It may work for objects that doesn't involve release of ownership, but for request object, user may perform the following and cause `TRITONSERVER_InferenceRequest` being deleted:
```python
req = triton_bindings.TRITONSERVER_InferenceRequest(...) # and initialize
req_2 = req
server.infer_async(req)
del req
del req_2
```
In current implementation of `infer_async`, it informs `req` that it no longer owns the Triton object, but such information is not passed to `req_2` and it will delete the Triton object while it is needed for inference.

python/binding/tritonserver_pybind.cc
```diff
@@ -0,0 +1,1962 @@
...
+
+  TRITONSERVER_MemoryType MemoryType()
+  {
+    TRITONSERVER_MemoryType memory_type = TRITONSERVER_MEMORY_CPU;
+    ThrowIfError(
+        TRITONSERVER_BufferAttributesMemoryType(triton_object_, &memory_type));
+    return memory_type;
+  }
+
+  size_t CudaIpcHandle()
```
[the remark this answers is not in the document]
GuanLuo: yes. IIRC `cudaIpcMemHandle_t` is a pointer address, and for all pointer it is represent as `int` in Python side.

CMakeLists.txt
```diff
@@ -415,3 +415,25 @@ export(
...
+  GIT_TAG "aa304c9c7d725ffb9d10af08a3b34cb372307020"
+  GIT_SHALLOW ON
+)
+FetchContent_MakeAvailable(pybind11)
+set(
+  PYTHON_BINDING_SRCS
+  python/binding/tritonserver_pybind.cc
+)
+
+pybind11_add_module(triton_bindings SHARED ${PYTHON_BINDING_SRCS})
```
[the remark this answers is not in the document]
nnshah1: I think `triton_bindings` is ok for initial release to get feedback / test. We can rename when the wrappers are in place.

Also ok to keep in same repository until then as well- but open to other thoughts.

python/binding/tritonserver_pybind.cc
```diff
@@ -0,0 +1,1991 @@
...
+      throw AlreadyExists(std::move(msg));
+    default:
+      throw Unknown(std::move(msg));
+  }
+}
+
+template <typename TritonStruct>
+class PyWrapper {
+ public:
+  explicit PyWrapper(TritonStruct* to, bool o) : triton_object_(to), owned_(o)
```
nnshah1, line 176: recommend: consistent naming for 'owned' - some places below we use 'owned'

same for triton_object

python/binding/tritonserver_pybind.cc
nnshah1: I agree keeping the naming aligned is useful and helps map them. 

One option could be to have the module name be triton_bindings.TRITONSERVER. <suffix>

python/binding/tritonserver_pybind.cc
```diff
@@ -0,0 +1,1991 @@
...
+    PyServer& server, const std::string& model_name,
+    const int64_t model_version)
+{
+  ThrowIfError(TRITONSERVER_InferenceRequestNew(
+      &triton_object_, server.Ptr(), model_name.c_str(), model_version));
+  owned_ = true;
+}
+
+// [FIXME] module name?
+PYBIND11_MODULE(triton_bindings, m)
```
nnshah1, line 1595: Are there any potentially long running c++ methods where we want to release the GIL or would all calls be  non blocking in any case?

python/binding/tritonserver_pybind.cc
GuanLuo: Actually I think in Python it is just reference counting the same object so `req_2` and `req` should be in sync. Although, just because I am less familiar with binding code in general,I still prefer to keep `DISALLOW_COPY_AND_ASSIGN` to make sure the ownership is controlled within c++ side.

PR #249, opened 2023-08-30
Add inference request cancellation APIs

Tabrizian:
server: https://github.com/triton-inference-server/server/pull/6252
common: https://github.com/triton-inference-server/common/pull/99


src/infer_request.h
```diff
@@ -766,6 +774,9 @@ class InferenceRequest {
 
   // Sequence I/O states used for implicit state.
   std::shared_ptr<SequenceStates> sequence_states_;
+
+  // Cancellation status
+  std::atomic<bool> is_cancelled_;
```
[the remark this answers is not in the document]
Tabrizian: good point, fixed.

nnshah1:
can we add the error code for cancellation into https://github.com/triton-inference-server/core/blob/b20f7b5eb12642a8ed6ddf58827dc2a0eb48d885/include/triton/core/tritonserver.h#L311 - so that will be the expected error code in responses for cancelled requests?

Tabrizian:
@nnshah1 sorry, I missed that. Added cancellation status.

src/test/request_cancellation_test.cc
```diff
@@ -0,0 +1,150 @@
...
+  FAIL_TEST_IF_ERR(
+      TRITONSERVER_InferenceRequestIsCancelled(request, &is_cancelled));
+  ASSERT_TRUE(is_cancelled);
+
+  is_cancelled = false;
+  FAIL_TEST_IF_ERR(TRITONBACKEND_ResponseFactoryIsCancelled(
+      response_factory, &is_cancelled));
+  ASSERT_TRUE(is_cancelled);
+
+  FAIL_TEST_IF_ERR(TRITONBACKEND_ResponseFactoryDelete(response_factory));
```
nnshah1, line 137: One test that might be nice here is to release the request from the backend - but the state of the response factory should still be cancelled / non cancelled -

pattern in some decoupled backends is to copy the request to an internal queue and release them while holding a reference to the response factory

Tabrizian: good point added.

src/infer_request.h
```diff
@@ -680,6 +679,20 @@ class InferenceRequest {
...
   }
 
+  Status Cancel()
+  {
+    if (state_ != InferenceRequest::State::INITIALIZED) {
+      response_factory_->Cancel();
+    } else {
+      return Status(
+          Status::Code::INTERNAL,
+          "Request cannot be cancelled before it has started executing.");
```
nnshah1, line 689: We may want to allow cancellation before Run() is called to make frontend life simpler -

But things to consider: is there any race condition between initialized and cancel? Cancel could be called from any thread as it is public. (I don't think there would be - but not sure) 

Also would tweak the message - its not necessarily executing - but before passed to InferAsync .. not sure best way to rephrase that though .


```suggestion
          "Request cannot be cancelled before it has been passed to Infer_Async");
```

Tabrizian: > We may want to allow cancellation before Run() is called to make frontend life simpler -

I don't think it would make the frontend life simpler. Frontend needs to make sure to only cancel the request after TRITONSERVER_InferAsync. This would be the same even if we allow cancelling before Run is called.

> But things to consider: is there any race condition between initialized and cancel? Cancel could be called from any thread as it is public. (I don't think there would be - but not sure)

I converted the request state to atomic. As long as cancel is called after initialize, I think we should be fine.

Will update the message.

nnshah1: One thing - we still need to reset cancelled status I believe on Run/Prepare - I may have missed it

Tabrizian: Addressed the comment.

nnshah1 (approved):

PR #250, opened 2023-08-31
Revert "Load PyTorch backend as a persistent backend (#224)"

Tabrizian:
This reverts commit 081152997cdada346fb0bde2a41adb99cf6a3c20.

### DIFF
--- a/src/backend_manager.cc
+++ b/src/backend_manager.cc
@@ -381,43 +381,4 @@ TritonBackendManager::BackendState(
   return Status::Success;
 }
 
-Status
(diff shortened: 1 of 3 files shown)


dyastremsky:
Approving! For future context, I believe you said this is no longer needed due to there was a fix to UCX in the HPCX library that resolved this issue.

dyastremsky (approved):

Tabrizian:
Unfortunately, the fix won't available in the min container until a couple of releases so I've applied the suggested workaround for now.

PR #251, opened 2023-08-31
Fix state transitions for re-running requests

rmccorm4:
Fixes L0_simple_lib--base, and allows requests to be re-used for multiple inferences. 

Adds a sanity check that pending count isn't incremented multiple times without getting decremented in some edge case where a request is PENDING, fails to get to the PENDING state, and ends up in PENDING state again (re-used).

General logic:
1. When a request is created, it is in INITIALIZED state
2. When a request is run, it is set to PENDING state
3. If a request makes it to the backend, it should be in EXECUTING state
    - At this point, the backend takes ownership of the request, and is responsible to call RequestRelease() on it, moving request into RELEASED state
4. If any error occurs before getting to backend, it should hit InferenceRequest::RespondIfError and be set to RELEASED state.
5. If somehow a request didn't make it to backend (EXECUTING state), but also didn't get RELEASED, then it should still be in PENDING state. If the request is to be re-used for another inference, it should call PrepareForInference() which should set state back to INITIALIZED at the start of the cycle. 

If a request were to be transitioned from RELEASED to PENDING/EXECUTING directly, this should indicate a logical error as the request should only execute via InferAsync entrypoint, which calls PrepareForInference ([re]sets to INITIALIZED state) then InferenceRequest::Run (sets to PENDING state).

### DIFF
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -150,13 +150,13 @@ InferenceRequest::SetState(InferenceRequest::State new_state)
   // Define state transitions
   switch (state_) {
     case InferenceRequest::State::INITIALIZED: {
-      if (new_state != InferenceRequest::State::STARTED) {
+      if (new_state != InferenceRequest::State::PENDING) {
         return generate_error();
       }
       IncrementPendingRequestCount();
       break;
     }
-    case InferenceRequest::State::STARTED: {
+    case InferenceRequest::State::PENDING: {
       if (new_state != InferenceRequest::State::EXECUTING) {
         return generate_error();
       }
@@ -170,7 +170,7 @@ InferenceRequest::SetState(InferenceRequest::State new_state)
       break;
     }
     case InferenceRequest::State::RELEASED: {
-      if (new_state != InferenceRequest::State::STARTED) {
+      if (new_state != InferenceRequest::State::PENDING) {
         // Only transition currently supported after release is to start again
         // when re-using request objects for multiple inferences.
         return generate_error();
@@ -384,7 +384,7 @@ InferenceRequest::OutputBufferProperties(
 Status
 InferenceRequest::Run(std::unique_ptr<InferenceRequest>& request)
 {
-  RETURN_IF_ERROR(request->SetState(InferenceRequest::State::STARTED));
+  RETURN_IF_ERROR(request->SetState(InferenceRequest::State::PENDING));
   return request->model_raw_->Enqueue(request);
 }
 
@@ -1588,8 +1588,8 @@ operator<<(std::ostream& out, const InferenceRequest::State& state)
       out << "INITIALIZED";
       break;
     }
-    case InferenceRequest::State::STARTED: {
-      out << "STARTED";
+    case InferenceRequest::State::PENDING: {
+      out << "PENDING";
       break;
     }
     case InferenceRequest::State::EXECUTING: {
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -106,20 +106,10 @@ InferenceRequest::InferenceRequest(
     : needs_normalization_(true), model_raw_(model),
       requested_model_version_(requested_model_version), flags_(0),
       correlation_id_(0), batch_size_(0), timeout_us_(0), collect_stats_(true),
-      state_(InferenceRequest::State::INITIALIZED), null_request_(false),
-      decrement_pending_count_(false)
+      state_(InferenceRequest::State::INITIALIZED), null_request_(false)
 {
   SetPriority(0);
 }
-
-InferenceRequest::~InferenceRequest()
-{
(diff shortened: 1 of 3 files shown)


kthui (approved):

src/infer_request.cc
```diff
@@ -151,40 +153,46 @@ InferenceRequest::SetState(InferenceRequest::State new_state)
       if (new_state != InferenceRequest::State::STARTED) {
```
nnshah1, line 153: would QUEUED or PENDING be clearer - STARTED seems to imply executing ...

rmccorm4: I'm good with either, if I had to choose I'd pick PENDING

src/infer_request.cc
```diff
@@ -151,40 +153,46 @@ InferenceRequest::SetState(InferenceRequest::State new_state)
...
       if (new_state != InferenceRequest::State::RELEASED) {
         return generate_error();
       }
-      state_ = new_state;
       break;
     }
     case InferenceRequest::State::RELEASED: {
-      // No state transition currently supported after release.
-      return generate_error();
+      if (new_state != InferenceRequest::State::STARTED) {
```
nnshah1, line 173: I think to reuse the request we'll need to move back to the INITIALIZED state (pending is only incremented in transition from INITIALIZED -> STARTING).

What if we reset state to INITIALIZED as part of PrepareForInference()

@Tabrizian To allow for reusing requests I think we should also reset cancelled state to false in either AsyncInfer() or PrepareForInference() -

rmccorm4: > What if we reset state to INITIALIZED as part of PrepareForInference()

I like this idea. Would you want to see some default state between construction `new InferenceRequest()` and PrepareForInference() ? Or use initialized for both?

Tabrizian: > @Tabrizian To allow for reusing requests I think we should also reset cancelled state to false in either AsyncInfer() or PrepareForInference() -

Agreed, I'll make this change.

src/infer_request.cc
```diff
@@ -142,49 +144,55 @@ InferenceRequest::SetState(InferenceRequest::State new_state)
...
-  if (reporter) {
-    reporter->IncrementGauge(kPendingRequestMetric, 1);
+  // Only increment once and do not increment again until decremented.
+  const bool increment_pending_count = !decrement_pending_count_;
+  if (increment_pending_count) {
+    auto reporter = model_raw_->MetricReporter();
+    if (reporter) {
+      reporter->IncrementGauge(kPendingRequestMetric, 1);
+    }
     decrement_pending_count_ = true;
```
nnshah1, line 196: Instead of "decrement_pending_count_" could we call this "is_pending_" ?

That seems to be clearer. When we move into pending state we set the flag.

The flag resets when initialized.

src/infer_request.cc
```diff
@@ -142,49 +144,55 @@ InferenceRequest::SetState(InferenceRequest::State new_state)
...
       break;
     }
     case InferenceRequest::State::RELEASED: {
-      // No state transition currently supported after release.
-      return generate_error();
+      if (new_state != InferenceRequest::State::INITIALIZED) {
+        // Only transition currently supported after release is to start over
+        // again, such as re-using request objects for multiple inferences.
+        return generate_error();
+      }
```
nnshah1, line 177: As the time between release and destruction is not specified - I think we would want to decrement pending count on release and not in the destructor.

So overall, the logic would be:

1) Initialize -> set is_pending to false
2) Initialize -> Pending -> set is_pending to true and Increment
3) Pending -> Execute -> set is_pending to false and decrement
4) *-> Released -> if (is_pending) set is_pending to false and decrement 

it is also possible I believe that it can all be handled in the state method without the need for the pending flag - if we check on which state is before release (pending -> release decrement, otherwise do not)

1. Initialized -> Pending: Increment
2. Pending -> Execute: Decrement
3. Pending -> Released: Decrement
4. (other) -> Released: Do nothing

I'd like to see if we can use the state or the pending flag - and not both if possible.

rmccorm4: > I'd like to see if we can use the state or the pending flag - and not both if possible.

I originally had decrement in destructor as a RAII-style sanity check that we'll clean up no matter what code path may get hit. However, I think your sentiment makes sense that we can capture the intended code paths in the transitions, such as Pending->Released for early exits. I'll try this approach within states rather than having an additional flag.

rmccorm4: @nnshah1 moved logic into state transitions only, appears to be working and passing CI as well. Let me know what you think.

nnshah1: LGTM -

nnshah1 (approved):

kthui (approved):

PR #252, opened 2023-09-06
Update cache_entry.cc to fix data type mismatches and out of bound er…

rvroge:
…rors


rvroge:
These fixes were required to compile for the following platform:
--platform=linux/arm/v7

PR #253, opened 2023-09-07
Fix request re-use when cancelling a request

Tabrizian:
When cancelling a request that is being re-used, we need to make sure that a new response factory is created for each inference. Otherwise, in the old code, if the user didn't call `SetResponseReleaseCallback`, cancelling a new request could cancel the old response factory as well which is not desired.

### DIFF
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -1023,11 +1023,7 @@ TRITONBACKEND_RequestIsCancelled(
 {
   InferenceRequest* tr = reinterpret_cast<InferenceRequest*>(request);
 
-  auto status = tr->IsCancelled(is_cancelled);
-  if (!status.IsOk()) {
-    return TRITONSERVER_ErrorNew(
-        StatusCodeToTritonCode(status.StatusCode()), status.Message().c_str());
-  }
+  RETURN_TRITONSERVER_ERROR_IF_ERROR(tr->IsCancelled(is_cancelled));
   return nullptr;
 }
 
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -1022,7 +1022,12 @@ TRITONBACKEND_RequestIsCancelled(
     TRITONBACKEND_Request* request, bool* is_cancelled)
 {
   InferenceRequest* tr = reinterpret_cast<InferenceRequest*>(request);
-  *is_cancelled = tr->IsCancelled();
+
+  auto status = tr->IsCancelled(is_cancelled);
+  if (!status.IsOk()) {
+    return TRITONSERVER_ErrorNew(
+        StatusCodeToTritonCode(status.StatusCode()), status.Message().c_str());
+  }
   return nullptr;
 }
 
--- a/src/backend_model_instance.cc
+++ b/src/backend_model_instance.cc
@@ -631,6 +631,10 @@ TritonModelInstance::WarmUp()
         request->CaptureQueueStartNs();
         triton_requests.push_back(
             reinterpret_cast<TRITONBACKEND_Request*>(request.get()));
+
+        // For warmup requests we need to manually set ResponseFactory
+        // since they don't run `PrepareForInference`.
+        request->SetResponseFactory();
       }
 
       Execute(triton_requests);
--- a/src/backend_model_instance.cc
+++ b/src/backend_model_instance.cc
@@ -623,6 +623,12 @@ TritonModelInstance::WarmUp()
         request->SetResponseCallback(
             &warmup_allocator, nullptr, WarmupResponseComplete,
             &response_complete[i]);
+
+        // For warmup requests we need to manually set ResponseFactory
+        // since they modify the callback after PrepareForInference has
+        // been called.
+        request->SetResponseFactory();
+
         // Capture timestamp before run to avoid incorrect accumulation from
         // sequential warmup runs
 #ifdef TRITON_ENABLE_STATS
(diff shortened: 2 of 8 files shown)


src/infer_request.h
```diff
@@ -679,10 +680,32 @@ class InferenceRequest {
     secondary_stats_aggregator_ = secondary_stats_aggregator;
   }
 
-  void Cancel() { response_factory_->Cancel(); }
-  void ResetCancel() { response_factory_->ResetCancel(); }
+  Status Cancel()
+  {
+    if (response_factory_) {
```
nnshah2, line 685: Question: if I understand this was also an issue in the previous iteration if Cancel() were called before SetResponseCallback() - is that correct?

Question: if a user calls inferasync without calling setresponsecallback - is that allowed?

src/infer_request.h
```diff
@@ -503,9 +503,10 @@ class InferenceRequest {
       TRITONSERVER_InferenceResponseCompleteFn_t response_fn,
       void* response_userp)
   {
-    response_factory_.reset(new InferenceResponseFactory(
-        model_shared_, id_, allocator, alloc_userp, response_fn, response_userp,
-        response_delegator_));
+    response_allocator_ = allocator;
```
rmccorm4, line 506: >   // Initialize the response factory that is to be used with any
  // responses produced for this request.
  
Comment above this function should be updated if making this change

src/infer_request.h
```diff
@@ -679,10 +680,32 @@ class InferenceRequest {
...
+  {
+    if (response_factory_) {
+      response_factory_->Cancel();
+    } else {
+      return Status(
+          Status::Code::INTERNAL,
+          "It is not possible to cancel an inference request before calling "
+          "TRITONSERVER_InferAsync.");
+    }
+    response_factory_->Cancel();
```
rmccorm4, line 693: Should this be called twice?

src/backend_model.cc
```diff
@@ -1022,7 +1022,12 @@ TRITONBACKEND_RequestIsCancelled(
     TRITONBACKEND_Request* request, bool* is_cancelled)
 {
   InferenceRequest* tr = reinterpret_cast<InferenceRequest*>(request);
-  *is_cancelled = tr->IsCancelled();
+
+  auto status = tr->IsCancelled(is_cancelled);
+  if (!status.IsOk()) {
+    return TRITONSERVER_ErrorNew(
```
rmccorm4, line 1028: Can we use [macro](https://github.com/triton-inference-server/core/blob/a23786c25b36f07d2cfd510821682c117b50a99e/src/status.h#L88C9-L94) here?
```cpp
RETURN_TRITONSERVER_ERROR_IF_ERROR(tr->IsCancelled(is_cancelled));
```

src/infer_request.h
```diff
@@ -679,10 +680,32 @@ class InferenceRequest {
...
+    return Status::Success;
+  }
 
-  bool IsCancelled() { return response_factory_->IsCancelled(); }
+  Status IsCancelled(bool* is_cancelled)
+  {
+    if (response_factory_) {
```
nnshah1, line 699: Question: in previous iteration - we would have had an issue if user called cancelled before setresponsecallback - is that correct?

Question: can user call inferasync without calling setresponse callback? does that error?

nnshah1 (approved):
had a few questions but only for clarification.

src/infer_request.h
Tabrizian: >in previous iteration - we would have had an issue if user called cancelled before setresponsecallback - is that correct?

that's correct.

>Question: can user call inferasync without calling setresponse callback? does that error?

I think if it is not provided it would segfault when the backend tries to send responses: https://github.com/triton-inference-server/core/blob/a23786c25b36f07d2cfd510821682c117b50a99e/src/infer_response.cc#L219C11-L223

rmccorm4: > I think if it is not provided it would segfault when the backend tries to send responses: https://github.com/triton-inference-server/core/blob/a23786c25b36f07d2cfd510821682c117b50a99e/src/infer_response.cc#L219C11-L223

Maybe low hanging fruit for the future, but we could catch that error and return it for some improved UX instead of segfaulting for C-API and bindings users.

GuanLuo: `SetResponseCallback` is part of the request initialization, the server assumes the request is properly initialized when `InferAsync` is called.

I think we can draw the separation between before and after `InferAsync`, only after the call, infer request is passed into Triton realm. Before that, the user / frontend has all the control of infer request and calling `Cancel / IsCancelled` doesn't have real meaning.

GuanLuo (commented):
LGMT, just need to address the other comments

rmccorm4 (approved):
Clean :rocket:

src/backend_model_instance.cc
```diff
@@ -631,6 +631,11 @@ TritonModelInstance::WarmUp()
         request->CaptureQueueStartNs();
         triton_requests.push_back(
             reinterpret_cast<TRITONBACKEND_Request*>(request.get()));
+
+        // For warmup requests we need to manually set ResponseFactory
+        // since they modify the callback after PrepareForInference has
+        // been called.
+        request->SetResponseFactory();
```
nnshah1, line 638: consider moving above the requests.push_back() for consistency with other request setup calls.

Tabrizian: moved.

PR #254, opened 2023-09-07
[WIP] Adding ability to specify subdirectory to download for `LocalizePath`

oandreeva-nv:
To avoid unnecessary copy during localization, I added `fetch_subdir` parameter to `LocalizePath`: If specified, will only download provided sub directory, otherwise all subdirectories will be downloaded. Does not affect files individual files, located under `path`.

[WIP] tag due to the following. When we agree on the proposed solution, I will extend it to GCS and AS. 

What was seen before:
```
root@oandreeva-dt:/opt/tritonserver# tree /tmp
/tmp
|-- folderDkrpXO
|   |-- 1
|   |   `-- model.pt
|   |-- 2
|   |   `-- model.pt
|   |-- 3
|   |   `-- model.pt
|   |-- config.pbtxt
|   `-- output0_labels.txt
|-- folderVuS3x0
|   |-- 1
|   |   `-- model.graphdef
|   |-- 2
|   |   `-- model.graphdef
|   |-- 3
|   |   `-- model.graphdef
|   |-- config.pbtxt
|   `-- output0_labels.txt
|-- folderfQ1yf3
|   |-- 1
|   |   `-- model.pt
|   |-- 2
|   |   `-- model.pt
|   |-- 3
|   |   `-- model.pt
|   |-- config.pbtxt
|   `-- output0_labels.txt
`-- folderwxZ6hr
    |-- 1
    |   `-- model.graphdef
    |-- 2
    |   `-- model.graphdef
    |-- 3
    |   `-- model.graphdef
    |-- config.pbtxt
    `-- output0_labels.txt
```
With this fix:

```
    root@oandreeva-dt:/opt/tritonserver# tree /tmp
/tmp
|-- folder9sfzV0
|   |-- 1
|   |   `-- model.pt
|   |-- config.pbtxt
|   `-- output0_labels.txt
|-- folderHBt0pT
|   |-- 3
|   |   `-- model.pt
|   |-- config.pbtxt
|   `-- output0_labels.txt
|-- folderZbfFyR
|   |-- 3
|   |   `-- model.graphdef
|   |-- config.pbtxt
|   `-- output0_labels.txt
|-- folderryK9pq
|   |-- 1
|   |   `-- model.graphdef
|   |-- config.pbtxt
|   `-- output0_labels.txt
```

### DIFF
--- a/src/filesystem/api.cc
+++ b/src/filesystem/api.cc
@@ -558,11 +558,13 @@ ReadTextProto(const std::string& path, google::protobuf::Message* msg)
 }
 
 Status
-LocalizePath(const std::string& path, std::shared_ptr<LocalizedPath>* localized)
+LocalizePath(
+    const std::string& path, const std::string& fetch_subdir,
+    std::shared_ptr<LocalizedPath>* localized)
 {
   std::shared_ptr<FileSystem> fs;
   RETURN_IF_ERROR(fsm_.GetFileSystem(path, fs));
-  return fs->LocalizePath(path, localized);
+  return fs->LocalizePath(path, fetch_subdir, localized);
 }
 
 Status
--- a/src/filesystem/api.h
+++ b/src/filesystem/api.h
@@ -145,11 +145,15 @@ Status ReadTextFile(const std::string& path, std::string* contents);
 
 /// Create an object representing a local copy of a path.
 /// \param path The path of the directory or file.
+/// \param fetch_subdir If specified, will only download provided
+/// sub directory, otherwise all subdirectories will be downloaded.
+/// Does not affect files individual files, located under `path`.
 /// \param localized Returns the LocalizedPath object
 /// representing the local copy of the path.
 /// \return Error status
 Status LocalizePath(
-    const std::string& path, std::shared_ptr<LocalizedPath>* localized);
+    const std::string& path, const std::string& fetch_subdir,
+    std::shared_ptr<LocalizedPath>* localized);
 
 /// Write a string to a file.
 /// \param path The path of the file.
--- a/src/filesystem/implementations/as.h
+++ b/src/filesystem/implementations/as.h
@@ -86,7 +86,7 @@ class ASFileSystem : public FileSystem {
       const std::string& path, std::set<std::string>* files) override;
   Status ReadTextFile(const std::string& path, std::string* contents) override;
   Status LocalizePath(
-      const std::string& path,
+      const std::string& path, const std::string& fetch_subdir,
       std::shared_ptr<LocalizedPath>* localized) override;
   Status WriteTextFile(
       const std::string& path, const std::string& contents) override;
@@ -424,7 +424,8 @@ ASFileSystem::DownloadFolder(
 
 Status
 ASFileSystem::LocalizePath(
-    const std::string& path, std::shared_ptr<LocalizedPath>* localized)
+    const std::string& path, const std::string& fetch_subdir,
+    std::shared_ptr<LocalizedPath>* localized)
 {
   bool exists;
   RETURN_IF_ERROR(FileExists(path, &exists));
--- a/src/filesystem/implementations/gcs.h
+++ b/src/filesystem/implementations/gcs.h
@@ -75,7 +75,7 @@ class GCSFileSystem : public FileSystem {
       const std::string& path, std::set<std::string>* files) override;
   Status ReadTextFile(const std::string& path, std::string* contents) override;
   Status LocalizePath(
-      const std::string& path,
+      const std::string& path, const std::string& fetch_subdir,
       std::shared_ptr<LocalizedPath>* localized) override;
   Status WriteTextFile(
       const std::string& path, const std::string& contents) override;
@@ -363,7 +363,8 @@ GCSFileSystem::ReadTextFile(const std::string& path, std::string* contents)
 
 Status
 GCSFileSystem::LocalizePath(
-    const std::string& path, std::shared_ptr<LocalizedPath>* localized)
+    const std::string& path, const std::string& fetch_subdir,
+    std::shared_ptr<LocalizedPath>* localized)
 {
   bool exists;
   RETURN_IF_ERROR(FileExists(path, &exists));
--- a/src/filesystem/implementations/local.h
+++ b/src/filesystem/implementations/local.h
@@ -49,7 +49,7 @@ class LocalFileSystem : public FileSystem {
       const std::string& path, std::set<std::string>* files) override;
   Status ReadTextFile(const std::string& path, std::string* contents) override;
   Status LocalizePath(
-      const std::string& path,
+      const std::string& path, const std::string& fetch_subdir,
       std::shared_ptr<LocalizedPath>* localized) override;
   Status WriteTextFile(
       const std::string& path, const std::string& contents) override;
@@ -204,7 +204,8 @@ LocalFileSystem::ReadTextFile(const std::string& path, std::string* contents)
 
 Status
 LocalFileSystem::LocalizePath(
-    const std::string& path, std::shared_ptr<LocalizedPath>* localized)
+    const std::string& path, const std::string& fetch_subdir,
+    std::shared_ptr<LocalizedPath>* localized)
 {
   // For local file system we don't actually need to download the
   // directory or file. We use it in place.
--- a/src/backend_model.cc
+++ b/src/backend_model.cc
@@ -75,10 +75,11 @@ TritonModel::Create(
   }
 
   // Localize the content of the model repository corresponding to
-  // 'model_path'. This model holds a handle to the localized content
-  // so that it persists as long as the model is loaded.
+  // 'model_path' and model's version. This model holds a handle to
+  // the localized content so that it persists as long as the model is loaded.
   std::shared_ptr<LocalizedPath> localized_model_dir;
-  RETURN_IF_ERROR(LocalizePath(model_path, &localized_model_dir));
+  RETURN_IF_ERROR(
+      LocalizePath(model_path, std::to_string(version), &localized_model_dir));
 
   // Localize paths in backend model config
   // [FIXME] Remove once a more permanent solution is implemented (DLIS-4211)
--- a/src/filesystem/implementations/common.h
+++ b/src/filesystem/implementations/common.h
@@ -83,7 +83,8 @@ class FileSystem {
   virtual Status ReadTextFile(
       const std::string& path, std::string* contents) = 0;
   virtual Status LocalizePath(
-      const std::string& path, std::shared_ptr<LocalizedPath>* localized) = 0;
+      const std::string& path, const std::string& fetch_subdir,
+      std::shared_ptr<LocalizedPath>* localized) = 0;
   virtual Status WriteTextFile(
       const std::string& path, const std::string& contents) = 0;
   virtual Status WriteBinaryFile(
--- a/src/filesystem/implementations/s3.h
+++ b/src/filesystem/implementations/s3.h
@@ -151,7 +151,7 @@ class S3FileSystem : public FileSystem {
       const std::string& path, std::set<std::string>* files) override;
   Status ReadTextFile(const std::string& path, std::string* contents) override;
   Status LocalizePath(
-      const std::string& path,
+      const std::string& path, const std::string& fetch_subdir,
       std::shared_ptr<LocalizedPath>* localized) override;
   Status WriteTextFile(
       const std::string& path, const std::string& contents) override;
@@ -628,7 +628,8 @@ S3FileSystem::ReadTextFile(const std::string& path, std::string* contents)
 
 Status
 S3FileSystem::LocalizePath(
-    const std::string& path, std::shared_ptr<LocalizedPath>* localized)
+    const std::string& path, const std::string& fetch_subdir,
+    std::shared_ptr<LocalizedPath>* localized)
 {
   // Check if the directory or file exists
   bool exists;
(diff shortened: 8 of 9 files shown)


src/filesystem/api.h
```diff
@@ -145,11 +145,15 @@ Status ReadTextFile(const std::string& path, std::string* contents);
...
+/// \param fetch_subdir If specified, will only download provided
+/// sub directory, otherwise all subdirectories will be downloaded.
+/// Does not affect files individual files, located under `path`.
 /// \param localized Returns the LocalizedPath object
 /// representing the local copy of the path.
 /// \return Error status
 Status LocalizePath(
-    const std::string& path, std::shared_ptr<LocalizedPath>* localized);
+    const std::string& path, const std::string& fetch_subdir,
+    std::shared_ptr<LocalizedPath>* localized);
```
kthui, line 156: I think the interface is insufficiently generic for functions at the level of `LocalizePath()`, because `LocalizePath()` should be available to any callers who want to localize any path on the cloud, but `fetch_subdir` imposes a limitation on the caller to either fetch all sub-directories within the path or only the sub-directory specified. The caller may not "choose" which sub-directories to fetch or not fetch. let alone choosing sub-directories within a sub-directory.

I think there are many different generic design choices. For example:
1. Use `bool recursive` instead of `string fetch_subdir`. The caller can choose not fetching any sub-directories with `recursive = false` or all sub-directories with `recursive = true`. Then, the problem can be accomplished by first fetching the given model path with `recursive = false` (i.e. `/repo/model`) and then the model path with version with `recursive = true` (i.e. `/repo/model/1`). However, this will require the `LocalizedPath` object to be able to join the model path and model version path localization objects into one localization object.
2. Allow the caller to specify a list of sub-paths to localize (or not localize). The "complicated" structure of sub-paths allows for fetching different combinations of sub-directories in a single call, which eliminates the need for joining model path and model version path localization objects, but the "complicated" structure is less intuitive for the caller to use and may not be easier to implement than the first design choice.

However, the interface accomplishes exactly what we want to solve at this time. I will defer to @GuanLuo and @nnshah1 to decide.

oandreeva-nv: > However, this will require the LocalizedPath object to be able to join the model path and model version path localization objects into one localization object.
Do we? If I understand correctly, versioned path is set here: https://github.com/triton-inference-server/core/blob/a23786c25b36f07d2cfd510821682c117b50a99e/src/backend_model.cc#L113

so instead of 
```
const auto version_path =
      JoinPath({localized_model_path, std::to_string(version)});
```
we would do something like:
```
 std::shared_ptr<LocalizedPath> version_path;
  RETURN_IF_ERROR(LocalizePath(path/to/version, &version_path));
```
or am I missing something?

kthui: I am thinking something like:
```
std::string model_path = "/repo/model_dir";
std::string model_version_path = "/repo/model_dir/1";

std::shared_ptr<LocalizedPath> model_localized;
std::shared_ptr<LocalizedPath> model_version_localized;
LocalizePath(model_path, recursive=false, &model_localized);
LocalizePath(model_version_path, recursive=true, &model_version_localized);

model_localized->Include(model_version_localized);
```

The idea behind `LocalizedPath::Include()` is to convert from
```
model_tmp
|
`--- config.pbtxt
```
and
```
model_version_tmp
|
`--- model.py
```
to
```
model_tmp
|
`--- config.pbtxt
|
`--- 1
     |
     `--- model.py
```
on the local disk.

oandreeva-nv: I actually am missing something. Even though I like option 1 suggestion, localization in this case will happen to another directory. Meaning, at current stage every call to `LocalizePath` creates a new temporary directory. Thus if we separately will try to load model and its version, they will be copied in 2 different directories. So unless we want to change that, option 1 is not the best option.

kthui: I think it would be ok to only implement the base case above, and throw an error if the caller is trying to join/include anything more complicated.

oandreeva-nv: Based on offline discussion, with @kthui, we can add one more parameter `location` to control where to localize provided path, i.e. something like
```
LocalizePath(model_version_path, recursive-true/false, location=<string>, &model_version_localized);
```
Then we can control where version directory will be localized , and if `location` is empty, we create a new, temporary directory.

This will also help with the follow up ticket.

GuanLuo: > Meaning, at current stage every call to LocalizePath creates a new temporary directory. Thus if we separately will try to load model and its version, they will be copied in 2 different directories. So unless we want to change that, option 1 is not the best option.

I am thinking on the same direction as `LocalizedPath::Include()` that if we want to have a generic way to localizing directories (with customization on fetching subset of contents / recurrsiveness), we want to make `LocalizedPath` a "filler" class: as on calling `fs::LocalizePath`, it returns an fully localized object (current) or empty object (with local root path assigned). Then if user want to customize the localization, they can call:
```
LocalizedPath::Include("sub/path/content", recursive);
```
Although I think it then the process is convoluted that there is a iterative calls of `GetDirectorySubdirs/Files` and `LocalizedPath::Include`

GuanLuo: Note that this is really an internal API (unless we want to expose file system API in the future) so we are free to modify the interface as frequent as we want and don't need to worry about being generic. If localizing model directory is somewhat specific, we may just introduce a `LocalizeModelDirectory` method to fit the specific need.

Because when you look at what to be localized for a model version, it is actually "everything in the directory except for the unwanted version subdirectories", that is tricky to express it with some generic language.

oandreeva-nv:
@kthui, @GuanLuo  and @nnshah1. It is ready for re-review. 

I didn't implement `LocalizedPath::Include`, since it can be done as a separate follow-up PR with [rename](https://man7.org/linux/man-pages/man2/rename.2.html) for Linux and [MoveFile](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-movefile?redirectedfrom=MSDN) for Windows with concerns regarding 
> convoluted that there is a iterative calls of GetDirectorySubdirs/Files and LocalizedPath::Include

properly considered. As of now, I think specifying `mount_dir` of a model's localized path when calling `LocalizePath(
      version_path, true /*recursive*/, localized_model_dir->Path(),
      &localized_version_dir)`  should be sufficient. To make sure that localized content for version persists, I add `localized_version_dir` to `localized_model_dir->other_localized_path`:
      `localized_model_dir->other_localized_path.push_back(localized_version_dir);`
      
Regarding environmental variable. Current name is `TRITON_AWS_MOUNT_DIRECTORY`. I think it would make sense to create filesystem type-specific env names for gcs and as as well. Open for discussion. 
I thought about adding the ability to specify this location in `credentials` file, but I think since `local mount directory` is not really a credential, it is not the best location + code's logic reads contents into `S3Credential` struct, it would be strange to have `mount_dir` member in this file, thus it would need some other wrapping structure + more code change. We can do it later upon user request.

Finally, current requirements for mount directory:
1) Should exist before 
2) Should not contain same elements (folders and files) as remote. Otherwise, upon stopping server, all same contents will be deleted from the local directory. Let me know if this is a reasonable ask. 

Current look: bucket  is mounted to `/tmp/aws`:
```
/tmp
`-- aws
    |-- graphdef_float32_float32_float32
    |   |-- 1
    |   |   `-- model.graphdef
    |   |-- 3
    |   |   `-- model.graphdef
    |   |-- config.pbtxt
    |   `-- output0_labels.txt
    `-- libtorch_float32_float32_float32
        |-- 1
        |   `-- model.pt
        |-- 3
        |   `-- model.pt
        |-- config.pbtxt
        `-- output0_labels.txt
```

src/filesystem/api.cc
```diff
@@ -558,11 +558,23 @@ ReadTextProto(const std::string& path, google::protobuf::Message* msg)
...
 Status
-LocalizePath(const std::string& path, std::shared_ptr<LocalizedPath>* localized)
+LocalizePath(
+    const std::string& path, const bool recursive,
+    std::shared_ptr<LocalizedPath>* localized)
 {
   std::shared_ptr<FileSystem> fs;
   RETURN_IF_ERROR(fsm_.GetFileSystem(path, fs));
-  return fs->LocalizePath(path, localized);
+  return fs->LocalizePath(path, recursive, "", localized);
```
oandreeva-nv, line 567: this (L565-L567) can potentially be shortened to 
```
retrurn LocalizePath(path, recursive, "", localized)
```

src/filesystem/implementations/gcs.h
```diff
@@ -363,7 +364,8 @@ GCSFileSystem::ReadTextFile(const std::string& path, std::string* contents)
 
 Status
 GCSFileSystem::LocalizePath(
-    const std::string& path, std::shared_ptr<LocalizedPath>* localized)
+    const std::string& path, const bool recursive, const std::string& mount_dir,
```
GuanLuo, line 367: Are you going to implement the new API in other cloud file system in this PR as well? If not, should return error if `!mount_dir.empty()`

src/filesystem/implementations/local.h
```diff
@@ -254,8 +256,12 @@ LocalFileSystem::MakeDirectory(const std::string& dir, const bool recursive)
   if (mkdir(dir.c_str(), S_IRWXU) == -1)
 #endif
   {
-    // Only allow the error due to parent directory does not exist
-    // if 'recursive' is requested
+    // Return success if directory already exists
+    if (errno == EEXIST) {
```
GuanLuo, line 260: Should it be a function argument on whether `file exist` is allowed?

GuanLuo (dismissed):
Minor comments

src/filesystem/implementations/gcs.h
oandreeva-nv: Yes, I wanted to first get an agreement on the implementation for s3. I'll populate it to GCS and AS in the next commit for easier review

src/filesystem/implementations/local.h
oandreeva-nv: Good idea, I'll add it

kthui (dismissed):

src/filesystem/api.h
```diff
@@ -145,11 +145,27 @@ Status ReadTextFile(const std::string& path, std::string* contents);
...
 Status LocalizePath(
-    const std::string& path, std::shared_ptr<LocalizedPath>* localized);
+    const std::string& path, const bool recursive,
+    std::shared_ptr<LocalizedPath>* localized);
+
+/// Create an object representing a local copy of a path.
+/// \param path The path of the directory or file.
+/// \param recursive If true, will fetch all sub-directories in
+/// the provided path.
+/// \param mount_dir If specified, will use provided local directory
```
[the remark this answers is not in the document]
oandreeva-nv: I'll clarify, thank you for suggestion

src/filesystem/implementations/as.h
```diff
@@ -406,17 +409,21 @@ ASFileSystem::DownloadFolder(
...
-          const_cast<char*>(local_path.c_str()), S_IRUSR | S_IWUSR | S_IXUSR);
-      if (status == -1) {
-        return Status(
-            Status::Code::INTERNAL,
-            "Failed to create local folder: " + local_path +
-                ", errno:" + strerror(errno));
+    if (recursive) {
+      for (const auto& directory_item : blob_prefixes) {
+        const auto& local_path = JoinPath({dest, BaseName(directory_item)});
+        int status = mkdir(
```
GuanLuo, line 415: Shouldn't the `FileSystem` wrapped API be used here as well?

oandreeva-nv: right, for Windows. good catch

oandreeva-nv: changed

GuanLuo (approved):

kthui (approved):

rmccorm4:
@oandreeva-nv just checking my understanding, I see your comment saying:

> Finally, current requirements for mount directory:
> 1. Should exist before

But in the code I'm seeing calls to `MakeDirectory()`, which I'm inferring would create the directory if it doesn't exist.

So to clarify, is the mount directory required to exist before server startup? Or will Triton create it on demand, similar to tmp folder workflow?

src/filesystem/implementations/as.h
```diff
@@ -441,21 +443,31 @@ ASFileSystem::LocalizePath(
...
+  std::string tmp_folder;
+  if (mount_dir.empty() && env_mount_dir == nullptr) {
+    RETURN_IF_ERROR(triton::core::MakeTemporaryDirectory(
+        FileSystemType::LOCAL, &tmp_folder));
+  } else {
+    tmp_folder = mount_dir.empty() ? std::string(env_mount_dir) : mount_dir;
+    tmp_folder =
+        JoinPath({tmp_folder, path.substr(path.find_last_of('/') + 1)});
+    RETURN_IF_ERROR(triton::core::MakeDirectory(
+        tmp_folder, true /*recursive*/, true /*allow_dir_exist*/));
```
rmccorm4, line 461: @oandreeva-nv just checking my understanding, I see your comment saying:

> Finally, current requirements for mount directory:
> 1. Should exist before

But in the code I'm seeing calls to `MakeDirectory()`, which I'm inferring would create the directory if it doesn't exist.

So to clarify, is the mount directory required to exist before server startup? Or will Triton create it on demand, similar to tmp folder workflow?

src/filesystem/implementations/as.h
```diff
@@ -441,21 +443,31 @@ ASFileSystem::LocalizePath(
...
-        "Failed to create local temp folder: " + folder_template +
-            ", errno:" + strerror(errno));
+  // Create a local directory for azure model store.
+  // If `mount_dir` or ENV variable are not set,
+  // creates a temporary directory under `/tmp` with the format: "folderXXXXXX".
+  // Otherwise, will create a folder under specified directory with the name
+  // indicated in path (i.e. everything after the last encounter of `/`).
+  const char* env_mount_dir = std::getenv("TRITON_AZURE_MOUNT_DIRECTORY");
+  std::string tmp_folder;
+  if (mount_dir.empty() && env_mount_dir == nullptr) {
```
rmccorm4, line 453: If there's already a follow-up ticket, I'd like to see use of some helper function like [this one](https://github.com/triton-inference-server/core/blob/80fef04220975fe332b35faa91d852cacea84e2f/src/filesystem/implementations/s3.h#L106-L108) that safely checks if the env var is set or is nullptr, and just returns a `std::string`. This helper could probably just take an env var name as an argument to condense the usage.

Currently this if/else is OK, but if the conditions were to be tweaked in the future, there is some risk for `std::string(env_mount_dir)` when it is a `nullptr`.

src/filesystem/implementations/as.h
oandreeva-nv: I will clarify this in a follow-up commit. This PR would need more de-bugging and testing

PR #257, opened 2023-09-08
Dynamic batch scheduler request cancellation

kthui:
The requests queued on the dynamic batch scheduler can be cancelled, if the cancellation is received before the request is added into a payload.

### DIFF
--- a/src/dynamic_batch_scheduler.cc
+++ b/src/dynamic_batch_scheduler.cc
@@ -57,15 +57,12 @@ IsStaleState(Payload::State payload_state)
 
 void
 FinishSkippedRequests(
-    std::shared_ptr<
-        std::vector<std::deque<std::unique_ptr<InferenceRequest>>>>&& requests,
+    std::vector<std::deque<std::unique_ptr<InferenceRequest>>>&& requests,
     const Status& response_status)
 {
-  if (requests != nullptr) {
-    for (auto& queue : *requests) {
-      for (auto& request : queue) {
-        InferenceRequest::RespondIfError(request, response_status, true);
-      }
+  for (auto& queue : requests) {
+    for (auto& request : queue) {
+      InferenceRequest::RespondIfError(request, response_status, true);
     }
   }
 }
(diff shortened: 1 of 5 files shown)


src/dynamic_batch_scheduler.cc
```diff
@@ -315,7 +330,7 @@ DynamicBatchScheduler::BatcherThread(const int nice)
     NVTX_RANGE(nvtx_, "DynamicBatcher " + model_name_);
 
     std::shared_ptr<std::vector<std::deque<std::unique_ptr<InferenceRequest>>>>
-        rejected_requests;
+        rejected_requests, cancelled_requests;
```
nnshah1, line 333: do these need to be shared pointers? I only see them used locally and not returned - maybe missing something though

kthui: I agree they (`rejected_requests`, `cancelled_requests`) do not need to be shared pointers, but just in case, @GuanLuo @tanmayv25 do you recall why `rejected_requests` is a shared pointer in the first place?

kthui: The L0_batcher test passed after removing the shared pointer wrapping.

Tabrizian (commented):
LGTM, but I defer to final approval to Guan/Tanmay since they are more knowledgable in this area of the code.

src/infer_request.cc
```diff
@@ -819,6 +819,7 @@ InferenceRequest::PrepareForInference()
   // inference execution.
   inputs_.clear();
   override_inputs_.clear();
+  ResetCancel();
```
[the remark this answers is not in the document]
kthui: Updated

src/infer_request.h
```diff
@@ -712,6 +713,16 @@ class InferenceRequest {
...
+  {
+    bool is_cancelled = false;
+    Status status = IsCancelled(&is_cancelled);
+    if (!status.IsOk()) {
+      LOG_ERROR << status.Message();
+    }
+    return is_cancelled;
+  }
+
 #endif  // TRITON_ENABLE_STATS
```
nnshah1, line 726: I did not notice this before - but looks like the Cancel and isCancel methods are within the TRITON_ENABLE_STATS ifdef - and should be moved out.

kthui: Thanks for catching this. It is moved out.

src/dynamic_batch_scheduler.cc
tanmayv25: I am not sure... but I think we should be fine.

tanmayv25 (approved):

PR #258, opened 2023-09-08
Add warning for transition from RELEASED to INITIALIZED

Tabrizian:
https://github.com/triton-inference-server/server/pull/6283

### DIFF
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -157,11 +157,14 @@ InferenceRequest::SetState(InferenceRequest::State new_state)
       if (new_state == InferenceRequest::State::RELEASED) {
         // no-op
       } else if (new_state == InferenceRequest::State::INITIALIZED) {
-        LOG_WARNING
-            << LogRequest() << " Invalid state transition detected from "
-            << state_ << " to " << new_state
-            << ". When re-using requests, make sure that the request release "
-               "callback is called before calling TRITONSERVER_InferAsync.";
+        if (!invalid_state_transition_warning_printed) {
+          invalid_state_transition_warning_printed = true;
+          LOG_WARNING
+              << LogRequest() << " Invalid state transition detected from "
+              << state_ << " to " << new_state
+              << ". When re-using requests, make sure that the request release "
+                 "callback is called before calling TRITONSERVER_InferAsync.";
+        }
       } else {
         // Unexpected state transition
         return generate_error();
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -154,7 +154,16 @@ InferenceRequest::SetState(InferenceRequest::State new_state)
       break;
     }
     case InferenceRequest::State::EXECUTING: {
-      if (new_state != InferenceRequest::State::RELEASED) {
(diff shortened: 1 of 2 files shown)


nnshah1 (approved):
Ok with warning or error as others see fit.

src/infer_request.cc
```diff
@@ -154,7 +154,16 @@ InferenceRequest::SetState(InferenceRequest::State new_state)
...
     case InferenceRequest::State::EXECUTING: {
-      if (new_state != InferenceRequest::State::RELEASED) {
+      if (new_state == InferenceRequest::State::RELEASED) {
+        // no-op
+      } else if (new_state == InferenceRequest::State::INITIALIZED) {
+        LOG_WARNING
+            << LogRequest() << " Invalid state transition detected from "
+            << state_ << " to " << new_state
+            << ". When re-using requests, make sure that the request release "
+               "callback is called before calling TRITONSERVER_InferAsync.";
```
rmccorm4, line 164: I'm fine with an interim warning here for now if we're worried about breaking existing cases.

Does it make sense to re-use generate_error here for the first part? Something like:
```cpp
LOG_WARNING << generate_error().Message() << ". When re-using requests ...";
```

Tabrizian: I think we can relnote this.

> Does it make sense to re-use generate_error here for the first part? Something like:

We would have to delete the error in that case.

rmccorm4: I think it would be cleaned up automatically because its a `status` not a `TRITONSERVER_Error*`, but I'm not particular to either way. I'm fine with how it is.

src/infer_request.cc
```diff
@@ -154,7 +154,19 @@ InferenceRequest::SetState(InferenceRequest::State new_state)
       break;
     }
     case InferenceRequest::State::EXECUTING: {
-      if (new_state != InferenceRequest::State::RELEASED) {
+      if (new_state == InferenceRequest::State::RELEASED) {
+        // no-op
+      } else if (new_state == InferenceRequest::State::INITIALIZED) {
```
rmccorm4, line 159: I think in one of your side branches I saw the `state_` changed to an atomic variable. I think this issue with callback vs. re-use timing exposed that there can be a race condition on state updates from release callback. Do you think we should keep it atomic for now as long as we "allow" release/re-use to happen in any order?

```
# Expected flow
EXECUTING -> (RELEASED) -> INITIALIZED -> PENDING -> EXECUTING

# Possible flows based on previous (release) callback timing, these can also error out, right?

## ERROR: invalid released->pending
EXECUTING -> INITIALIZED -> (RELEASED) -> PENDING -> EXECUTING

## ERROR: invalid released->executing
EXECUTING -> INITIALIZED -> PENDING -> (RELEASED) -> EXECUTING  
```

GuanLuo: So we are not enforcing request can only be reused after it is released (or ownership is back to the user)? As we only log warning message and continue the state transition

Tabrizian:
Closing this PR based on the offline discussion.

PR #260, opened 2023-09-09
Sequence batch scheduler request cancellation

kthui:
There are two cancellation scenarios:
1. If a sequence is queued on the backlog and one or more requests of the sequence is cancelled, then the sequence is cancelled along with all requests queued on the sequence..
2. If a sequence is inferencing (one or more requests are inferencing or completed) and a request of the sequence that has not started inference is cancelled, then the sequence is cancelled along with all requests queued on the sequence.


tanmayv25 (approved):

rmccorm4 (approved):
Looks very clean! 🚀

PR #261, opened 2023-09-09
Forward ensemble request parameters to composing models

rmccorm4:
Improves LLM pipeline UX. Applies to other workflows as well, but use of per-request parameters is particularly common for LLMs.

Previously, when certain model parameters are passed as request parameters to an ensemble (ex: preprocess, LLM, postprocess), the parameters wouldn't be accessible to the composing models. Now they will be forwarded to each step request in the ensemble.

Server: https://github.com/triton-inference-server/server/pull/6284
Core: https://github.com/triton-inference-server/core/pull/261

### DIFF
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -241,6 +241,13 @@ InferenceRequest::AddParameter(const char* name, const bool value)
   return Status::Success;
 }
 
+Status
+InferenceRequest::AddParameter(const InferenceParameter& parameter)
+{
+  parameters_.emplace_back(parameter);
+  return Status::Success;
+}
+
 #ifdef TRITON_ENABLE_TRACING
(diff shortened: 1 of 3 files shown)


src/infer_request.cc
```diff
@@ -241,6 +241,13 @@ InferenceRequest::AddParameter(const char* name, const bool value)
   return Status::Success;
 }
 
+Status
+InferenceRequest::AddParameter(const InferenceParameter& parameter)
```
nnshah1, line 245: One other option would be a 'SetParameters()' method that takes in a deque and uses the assignment oeprator of the dequeue.

nnshah1 (approved):

GuanLuo (approved):

PR #263, opened 2023-09-11
Ensemble scheduler request cancellation

kthui:
In between each ensemble steps, the parent request is checked for cancellation. If the parent request is cancelled, the cancellation is propagated to the next child request step. The scheduler of the individual request can then pick up the cancellation request, cancel the request and propagate the cancellation status back to parent.

### DIFF
--- a/src/ensemble_scheduler.cc
+++ b/src/ensemble_scheduler.cc
@@ -1264,6 +1264,11 @@ EnsembleContext::ScheduleSteps(
       }
     }
     if (should_schedule) {
+      // If the ensemble request is cancelled, propagate the cancellation to the
+      // next request step.
+      if (context->request_tracker_->Request()->IsCancelled()) {
+        step->request_->Cancel();
+      }
       // On a successful call to InferAsync(), the step will be released by
       // the response callback. When the response callback is invoked, the
       // step must not own (and release) the request as the request should be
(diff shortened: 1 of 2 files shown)


src/ensemble_scheduler.cc
```diff
@@ -1264,6 +1264,11 @@ EnsembleContext::ScheduleSteps(
       }
     }
     if (should_schedule) {
+      // If the ensemble request is cancelled, propagate the cancellation to the
+      // next request step.
+      if (context->request_tracker_->Request()->IsCancelled()) {
+        step->request_->Cancel();
```
Tabrizian, line 1270: I think rather than propagating the cancellation status we should stop scheduling new requests altogether. I believe for errors ensemble wouldn't call the next steps and I think we should do something similar here.

kthui: This is a good point, but there is a corner case. If the next step is a request belonging to an in-flight sequence, the entire in-flight sequence should be cancelled along with any queued requests on that sequence that may not have been explicitly cancelled. The ensemble scheduler is not knowledgeable enough to perform the cancellation, so it need to be deferred to the sequence scheduler on the next step. The simplest way of passing the information of cancellation to the sequence scheduler is to simply mark the next request as cancelled and pass it downstream.

Tabrizian: thanks for the detailed response! Resolving.

Tabrizian (approved):

tanmayv25 (approved):

PR #266, opened 2023-09-18
Specify non-random location for filesystem's temporary cache

oandreeva-nv:
Main ask for [DLIS-5340](https://jirasw.nvidia.com/browse/DLIS-5340)

Previous PR: https://github.com/triton-inference-server/core/pull/254 needs some debugging.

This PR will only read env variable directory (if specified) and create temporary folder there.

### DIFF
--- a/src/filesystem/api.cc
+++ b/src/filesystem/api.cc
@@ -619,12 +619,12 @@ MakeTemporaryDirectory(const FileSystemType type, std::string* temp_dir)
 {
   std::shared_ptr<FileSystem> fs;
   RETURN_IF_ERROR(fsm_.GetFileSystem(type, fs));
-  return fs->MakeTemporaryDirectory("/tmp", temp_dir);
+  return fs->MakeTemporaryDirectory(kDefaultMountDirectory, temp_dir);
 }
 
 Status
 MakeTemporaryDirectory(
-    const FileSystemType type, std::string& dir_path, std::string* temp_dir)
+    const FileSystemType type, std::string dir_path, std::string* temp_dir)
 {
   std::shared_ptr<FileSystem> fs;
   RETURN_IF_ERROR(fsm_.GetFileSystem(type, fs));
--- a/src/filesystem/api.cc
+++ b/src/filesystem/api.cc
@@ -619,7 +619,16 @@ MakeTemporaryDirectory(const FileSystemType type, std::string* temp_dir)
 {
   std::shared_ptr<FileSystem> fs;
   RETURN_IF_ERROR(fsm_.GetFileSystem(type, fs));
-  return fs->MakeTemporaryDirectory(temp_dir);
+  return fs->MakeTemporaryDirectory("/tmp", temp_dir);
+}
+
+Status
+MakeTemporaryDirectory(
+    const FileSystemType type, std::string& dir_path, std::string* temp_dir)
+{
+  std::shared_ptr<FileSystem> fs;
+  RETURN_IF_ERROR(fsm_.GetFileSystem(type, fs));
+  return fs->MakeTemporaryDirectory(dir_path, temp_dir);
 }
 
 Status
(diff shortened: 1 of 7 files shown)


src/filesystem/implementations/gcs.h
```diff
@@ -380,9 +381,12 @@ GCSFileSystem::LocalizePath(
         "GCS file localization not yet implemented " + path);
   }
 
+  // Create a local directory for GCS model store.
+  std::string env_mount_dir =
+      GetEnvironmentVariableOrDefault("TRITON_GCS_MOUNT_DIRECTORY", "/tmp");
```
rmccorm4, line 386: Can `"/tmp"` be some `common.h` constant like `constexpr char kDefaultMountDirectory[] = "/tmp"` or whatever name makes most sense?

src/filesystem/api.h
```diff
@@ -198,6 +198,15 @@ Status MakeDirectory(const std::string& dir, const bool recursive);
...
 Status MakeTemporaryDirectory(const FileSystemType type, std::string* temp_dir);
 
+/// Create a temporary directory of the specified filesystem type
+/// in a provided location.
+/// \param type The type of the filesystem.
+/// \param  dir_path The specified path.
+/// \param temp_dir Returns the path to the temporary directory.
+/// \return Error status
+Status MakeTemporaryDirectory(
+    const FileSystemType type, std::string& dir_path, std::string* temp_dir);
```
rmccorm4, line 208: I think `dir_path` is defined as a reference here `std::string& dir_path`, but a copy `std::string` everywhere else. Should it be copy here? It should be consistent.

oandreeva-nv: Yes, correct. Something strange happening with my local repo

src/filesystem/implementations/local.h
```diff
@@ -314,7 +316,10 @@ LocalFileSystem::MakeTemporaryDirectory(std::string* temp_dir)
         "Failed to create local temp folder: " + *temp_dir);
   }
 #else
-  std::string folder_template = "/tmp/folderXXXXXX";
+  if (dir_path.empty()) {
+    dir_path = "/tmp";
```
rmccorm4, line 320: use constant

rmccorm4 (approved):

src/filesystem/api.h
```diff
@@ -198,6 +198,15 @@ Status MakeDirectory(const std::string& dir, const bool recursive);
...
 Status MakeTemporaryDirectory(const FileSystemType type, std::string* temp_dir);
 
+/// Create a temporary directory of the specified filesystem type
+/// in a provided location.
+/// \param type The type of the filesystem.
+/// \param  dir_path The specified path.
+/// \param temp_dir Returns the path to the temporary directory.
+/// \return Error status
+Status MakeTemporaryDirectory(
+    const FileSystemType type, std::string dir_path, std::string* temp_dir);
```
kthui, line 208: A small thing, I think the `dir_path` can be declared as `const std::string& dir_path` here, and leave the `LocalFileSystem::MakeTemporaryDirectory(std::string dir_path, ...)` as it is. Then, we can save a copy of `dir_path` when passing the variable from the caller to `MakeTemporaryDirectory()` here, and only copy it once when passing `dir_path` from here to `LocalFileSystem::MakeTemporaryDirectory()`.

kthui, line 208:
```suggestion
    const FileSystemType type, const std::string& dir_path, std::string* temp_dir);
```

src/filesystem/api.cc
```diff
@@ -619,7 +619,16 @@ MakeTemporaryDirectory(const FileSystemType type, std::string* temp_dir)
 {
   std::shared_ptr<FileSystem> fs;
   RETURN_IF_ERROR(fsm_.GetFileSystem(type, fs));
-  return fs->MakeTemporaryDirectory(temp_dir);
+  return fs->MakeTemporaryDirectory(kDefaultMountDirectory, temp_dir);
+}
+
+Status
+MakeTemporaryDirectory(
+    const FileSystemType type, std::string dir_path, std::string* temp_dir)
```
kthui, line 627:
```suggestion
    const FileSystemType type, const std::string& dir_path, std::string* temp_dir)
```

kthui (approved):

src/filesystem/api.h
oandreeva-nv: I'll make sure to check and introduce this in a follow-up pr with the bug fix.   Thanks for suggestion!

PR #267, opened 2023-09-18
Minor fix to supress warning

fpetrini15:
Minor fix to suppress warning that control reaches the end of non-void function.


rmccorm4 (approved):
:pray:

PR #269, opened 2023-09-28
Added logic to core for handling python backend based backends

oandreeva-nv:
Initial look and feel:

~Added new field in `class TritonModel ` to store location to `model.py`, provided by the backend. Then, when `TritonModel::Create` invoked, we check if `.so` is present, if not we load python backend (if not loaded already) and store the location of `model.py` in the backend repo.~

~`TRITONBACKEND_BackendModelLocation` will be used by Python backend to request the possible backend based `model.py`.~ 

~Main con of this approach at the moment is that these changes (including backend API) are only relevant to python.~

All namings are open for discussion, ~mainly api name `TRITONBACKEND_BackendModelLocation`.~

[DONE] Will probably followup with backend_model.cc L122-154 refactoring.

~`TritonBackendManager::StoreCustomBackend` was introduced so that I could print custom backend info on start up:
Screenshot from 2023-09-28 14-42-27~

~I don't populate config now, because custom backends are based on python + I need to test how configs are propagated~

[Edit 2] 
Each python backend based backend is a new backend with `libpath` pointing to shared library and `libdir` to `model.py` location. 
No new APIs with this PR
All namings are still open for discussion
Did refactoring

Current config table:
image

### DIFF
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -375,6 +375,31 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_OutputBufferAttributes(
 TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_RequestId(
     TRITONBACKEND_Request* request, const char** id);
 
+/// Query whether the request is cancelled or not.
+///
+/// If possible the backend should terminate any processing and
+/// send an error response with cancelled status.
+///
+/// \param request The inference request.
+/// \param is_cancelled Returns true if the request is cancelled otherwise it
+/// would return false.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_RequestIsCancelled(
+    TRITONBACKEND_Request* request, bool* is_cancelled);
+
+/// Query whether the response factory is cancelled or not.
+///
+/// If possible the backend should terminate any processing and
+/// send an error response with cancelled status.
+///
+/// \param factory The response factory
+/// \param is_cancelled Returns true if the request is cancelled otherwise it
+/// would return false.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONBACKEND_DECLSPEC TRITONSERVER_Error*
+TRITONBACKEND_ResponseFactoryIsCancelled(
+    TRITONBACKEND_ResponseFactory* factory, bool* is_cancelled);
+
 /// Get the correlation ID of the request if it is an unsigned integer.
 /// Zero indicates that the request does not have a correlation ID.
 /// Returns failure if correlation ID for given request is not an unsigned
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -94,7 +94,7 @@ struct TRITONBACKEND_Batcher;
 ///   }
 ///
 #define TRITONBACKEND_API_VERSION_MAJOR 1
-#define TRITONBACKEND_API_VERSION_MINOR 16
+#define TRITONBACKEND_API_VERSION_MINOR 17
 
 /// Get the TRITONBACKEND API version supported by Triton. This value
 /// can be compared against the TRITONBACKEND_API_VERSION_MAJOR and
@@ -982,6 +982,15 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_ModelVersion(
 TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_ModelRepository(
     TRITONBACKEND_Model* model, TRITONBACKEND_ArtifactType* artifact_type,
     const char** location);
+
+/// Get the location of the model (moedl.py), provided by a python based
+/// backend.
+///
+/// \param model The model.
+/// \param location Returns the location.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_BackendModelLocation(
+    TRITONBACKEND_Model* model, const char** location);
 
 /// Get the model configuration. The caller takes ownership of the
 /// message object and must call TRITONSERVER_MessageDelete to release
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -94,7 +94,7 @@ struct TRITONBACKEND_Batcher;
 ///   }
 ///
 #define TRITONBACKEND_API_VERSION_MAJOR 1
-#define TRITONBACKEND_API_VERSION_MINOR 17
+#define TRITONBACKEND_API_VERSION_MINOR 16
 
 /// Get the TRITONBACKEND API version supported by Triton. This value
 /// can be compared against the TRITONBACKEND_API_VERSION_MAJOR and
@@ -982,15 +982,6 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_ModelVersion(
 TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_ModelRepository(
     TRITONBACKEND_Model* model, TRITONBACKEND_ArtifactType* artifact_type,
     const char** location);
-
-/// Get the location of the model (moedl.py), provided by a python based
-/// backend.
-///
-/// \param model The model.
-/// \param location Returns the location.
-/// \return a TRITONSERVER_Error indicating success or failure.
-TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_BackendModelLocation(
-    TRITONBACKEND_Model* model, const char** location);
 
 /// Get the model configuration. The caller takes ownership of the
 /// message object and must call TRITONSERVER_MessageDelete to release
--- a/src/backend_manager.cc
+++ b/src/backend_manager.cc
@@ -338,7 +338,7 @@ Status
 TritonBackendManager::CreateBackend(
     const std::string& name, const std::string& dir, const std::string& libpath,
     const triton::common::BackendCmdlineConfig& backend_cmdline_config,
-    bool& is_python_backend_based, std::shared_ptr<TritonBackend>* backend)
+    bool is_python_based_backend, std::shared_ptr<TritonBackend>* backend)
 {
   std::lock_guard<std::mutex> lock(mu_);
 
@@ -357,8 +357,8 @@ TritonBackendManager::CreateBackend(
   RETURN_IF_ERROR(TritonBackend::Create(
       name, dir, libpath, backend_cmdline_config, backend));
 
-  (*backend)->SetPythonBackendBasedFlag(is_python_backend_based);
-  if (is_python_backend_based) {
+  (*backend)->SetPythonBasedBackendFlag(is_python_based_backend);
+  if (is_python_based_backend) {
     backend_map_.insert({std::string(dir + "/model.py"), *backend});
   } else {
     backend_map_.insert({libpath, *backend});
--- a/src/backend_manager.cc
+++ b/src/backend_manager.cc
@@ -386,7 +386,7 @@ TritonBackendManager::BackendState(
     backend->BackendConfig().Serialize(&backend_config, &backend_config_size);
     backend_state_map->insert(
         {backend->Name(), std::vector<std::string>{libpath, backend_config}});
-    if (backend->IsPythonBackendBased()) {
+    if (backend->IsPythonBackendBased() && python_backend_config.empty()) {
       python_backend_config.emplace_back(backend->LibPath());
       python_backend_config.emplace_back(std::string(backend_config));
     }
--- a/src/backend_manager.cc
+++ b/src/backend_manager.cc
@@ -345,6 +345,10 @@ TritonBackendManager::CreateBackend(
   const auto& itr = backend_map_.find(libpath);
   if (itr != backend_map_.end()) {
     *backend = itr->second;
+    // Python backend based backends use the same shared library as python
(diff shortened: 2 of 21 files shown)


include/triton/core/tritonbackend.h
```diff
@@ -983,6 +983,15 @@ TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_ModelRepository(
...
 
+/// Get the location of the model (moedl.py), provided by a python based
+/// backend.
+///
+/// \param model The model.
+/// \param location Returns the location.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONBACKEND_DECLSPEC TRITONSERVER_Error* TRITONBACKEND_BackendModelLocation(
+    TRITONBACKEND_Model* model, const char** location);
+
```
rmccorm4, line 994: Isn't [TRITONBACKEND_BackendArtifacts](https://github.com/triton-inference-server/core/blob/55358c65e222c559c72527b783abd5971dcd293b/include/triton/core/tritonbackend.h#L897-L915) essentially the same API? Then you can just search for `model.py` within the backend directory returned by this API.

oandreeva-nv: I don't think so, [TRITONBACKEND_BackendArtifacts](https://github.com/triton-inference-server/core/blob/55358c65e222c559c72527b783abd5971dcd293b/include/triton/core/tritonbackend.h#L897-L915) interact with a backend, which will be `python backend` and all related `.so` libs locations.

Under the hood of new python-based backends is `python backend`, so `backend` passed to  `TRITONBACKEND_BackendArtifacts(
    TRITONBACKEND_Backend* backend, TRITONBACKEND_ArtifactType* artifact_type,
    const char** location)` is associated with `***_python.so` and not with custom backend. Thus, this api won't work.. Let me know, if I am wrong

rmccorm4: Hm, I think based on current implementation details the `backend->Directory()` is being set to where python backend lives, ex: `/opt/tritonserver/backends/python`, so you're right it wouldn't work. 

In my head, I thought the backend would still be considered it's own "backend" and living in `/opt/tritonserver/backends/vllm` in the `TritonBackend` object for example, but under the hood just loaded `libtriton_python.so`. I think this would remove the need for storing "custom backends" and the need for adding a relatively duplicate new C API. But this would require some refactoring. @Tabrizian @tanmayv25 any thoughts here?

oandreeva-nv: You are correct, it will be living there, however we can't create a proper backend without `.so` file and `CreateBackend` can't be called on custom backends. I believe the idea was: we look for `.so` -> look for `model.py`. If we found `model.py` it still needs proper `libtriton_python.so` to load all handles from, so we create python backend and load `python`'s shared library first, then store path to  `/opt/tritonserver/backends/vllm/model.py` as a part of `TritonModel`.

Tabrizian: Is there a way to "hardcode" this special use of Python backend while somehow make `TRITONBACKEND_Artifacts`return the path to e.g. vllm backend?

I agree that having a separate API looks a bit awkward.

oandreeva-nv: `TRITONBACKEND_BackendArtifacts` would need somehow to return 2 locations then: to `libtriton_python.so` and vllm backend + this path to vllm backend should be know associated with the backend and not the model. One option can be to create a copy of python's backend for vllm backend ( to have an access to `.so` and `.py` stored in the same backend object). Still, I think even though we call these custom backends `backends`, they really just store model's executor, which should be associated with a Model object. Under the hood, it is still a `python backend`.

oandreeva-nv: We can introduce `TRITONBACKEND_ModelArtifacts` and make it more general.

oandreeva-nv: Although even with the copy, `TRITONBACKEND_BackendArtifacts` should return location to `.so` , and it will in most cases be in a different location than `model.py`

rmccorm4: Did we decide on how we will handle dependencies like vllm python package? For example, if we intended to go the route of execution environments or custom stubs, etc. - then I think we would **want** TRITONBACKEND_BackendArtifacts to return the path to the `/opt/tritonserver/backends/vllm` and have a custom stub in that directory. 

Would it make sense to copy the libtriton_python.so and stub into the `/vllm` backend directory, rather than routing it to the python backend dir, and keep them completely separate? @Tabrizian

oandreeva-nv: I think this would be very `vllm`-oriented. Should we then extend this idea (copying `.so` to the same custom backend folder) for every custom backend?

rmccorm4: I don't think it's that vllm oriented. Pytorch handler requires `pip install torch / torchvision` , Hugginface handler requires `pip install huggingface`. 

If we intended to handle these dependencies via stubs or environments, then I think we need to think about that. 

If we intend a somewhat simpler path for now and just install the dependencies in the container environment `pip install <backend>/requirements.txt` either baked into container, or on demand, then that's different.

oandreeva-nv: I see, thanks for clarification

oandreeva-nv: My original thought was going in the route of `pip install <backend>/requirements.txt` for first round of vllm container.

src/backend_model.cc
```diff
@@ -114,42 +114,58 @@ TritonModel::Create(
...
       JoinPath({backend_dir, specialized_backend_name});
-  const std::vector<std::string> search_paths = {
+  std::vector<std::string> search_paths = {
       version_path, localized_model_path, global_path};
 
   std::string backend_libdir;
   std::string backend_libpath;
-  for (const auto& path : search_paths) {
-    const auto full_path = JoinPath({path, backend_libname});
+  RETURN_IF_ERROR(LocateBackendLibrary(
```
nnshah1, line 122: Question: could LocateBackendLibaray try to find model.py as well as .so / .dll and return a bool flag for a 'python_backend" - that way the search paths for .so and model.py would be the same?

src/backend_model.cc
```diff
@@ -114,42 +114,58 @@ TritonModel::Create(
...
+  RETURN_IF_ERROR(
+      ResolveBackendConfigs(backend_cmdline_config_map, backend_name, config));
 
   RETURN_IF_ERROR(SetBackendConfigDefaults(config));
 
   std::shared_ptr<TritonBackend> backend;
   RETURN_IF_ERROR(server->BackendManager()->CreateBackend(
-      model_config.backend(), backend_libdir, backend_libpath, config,
-      &backend));
+      backend_name, backend_libdir, backend_libpath, config, &backend));
```
nnshah1, line 168: Question: could we modify `CreateBackend`, and the `TritonBackend` class with additional flag to indicate a `python_backend`?

Wondering if instead of adding path to the model or having a new custom map - we could somehow augment the Backend Creation  / Backend class to store the additional information ....

Would that allow for the artifact directory to return the python folder (if the backend class had a flag) and avoid the need for a new api?

src/backend_model.h
```diff
@@ -218,6 +235,12 @@ class TritonModel : public Model {
   // persist as along as this object is retained by this model.
   std::shared_ptr<LocalizedPath> localized_model_dir_;
 
+  // Python-based backends provide model.py in their directory.
+  // This `model.py` is later reused by models,
+  // using this python-based backend.
+  // We provide the location of common model.py, when model is initialized.
+  std::string py_backend_model_path_ = "";
```
nnshah1, line 242: Could this be added to the `TritonBackend` class instead?

nnshah1:
Very high level thought: could the `TritonBackend` class be updated to handler the distinction between `python_backend` and `shared_library_backend`. Not sure its possible so more of a question

src/backend_model.cc
rmccorm4: This is essentially what I was getting at in my other comment. I think the first iteration of this can be done without introducing any new APIs with a little refactoring.

oandreeva-nv: I would like to point out that, when backend is being created, it checks if the same backend was already loaded in `backend_map`, if it was, then model is associated with this already created backend. Thus, if we want to make sure that Triton handles `python` model + `vllm` model, for both of these instances we should have a separate copy of a backend then. One standard, one with specified location to `model.py`.  If this is the route, then probably on the `python backend` side I can add a special handling when `TRITONBACKEND_BackendName` returns non-`python` backend name. I can explore that route. A quick disadwantage I see, is multeple copies of essentially the same backend

oandreeva-nv: Basically, some special logic here to see if a backend is a true `python` backend or not: https://github.com/triton-inference-server/python_backend/blob/b136bf3fad77d5d467a6db5bd739be16eb9d9400/src/python_be.cc#L1722

oandreeva-nv: cc @tanmayv25 this is basically goes back to our discussion on where to keep path to `model.py`.

src/backend_model.h
oandreeva-nv: this [comment](https://github.com/triton-inference-server/core/pull/269#discussion_r1341875824)

src/backend_model.cc
oandreeva-nv: Just an update. There will be no special logic, since backend-model is 1 to 1 relationship. Thus I would need to copy python backend's `dlopen / dlsym handles` into custom backend, which I referred to earlier as `multiple copies of the same backend)

src/backend_manager.cc
```diff
@@ -338,19 +338,31 @@ Status
 TritonBackendManager::CreateBackend(
     const std::string& name, const std::string& dir, const std::string& libpath,
     const triton::common::BackendCmdlineConfig& backend_cmdline_config,
-    std::shared_ptr<TritonBackend>* backend)
+    bool& is_python_backend_based, std::shared_ptr<TritonBackend>* backend)
```
rmccorm4, line 341: Would remove reference to avoid confusion or possible bugs
```suggestion
    bool is_python_backend_based, std::shared_ptr<TritonBackend>* backend)
```

src/backend_manager.h
```diff
@@ -91,6 +92,11 @@ class TritonBackend {
 
   void* State() { return state_; }
   void SetState(void* state) { state_ = state; }
+  bool& IsPythonBackendBased() { return is_python_backend_based_; }
+  void SetPythonBackendBasedFlag(bool& is_python_backend_based)
```
rmccorm4, line 96:
```suggestion
  bool IsPythonBackendBased() { return is_python_backend_based_; }
  void SetPythonBackendBasedFlag(bool is_python_backend_based)
```

src/backend_model.cc
```diff
@@ -114,42 +114,51 @@ TritonModel::Create(
...
-    if (exists) {
-      backend_libdir = path;
-      backend_libpath = full_path;
-      break;
-    }
-  }
+  std::string python_runtime_modeldir;
 
+  RETURN_IF_ERROR(ResolveBackendPaths(
+      backend_name, backend_dir, model_config.name(), search_paths,
```
oandreeva-nv, line 125: note to self: replace backend_name with specialized_backend_name

src/backend_manager.cc
tanmayv25, line 341: is_python_based_backend

src/backend_manager.cc
```diff
@@ -338,19 +338,31 @@ Status
...
 {
   std::lock_guard<std::mutex> lock(mu_);
 
   const auto& itr = backend_map_.find(libpath);
   if (itr != backend_map_.end()) {
     *backend = itr->second;
-    return Status::Success;
+    // Python backend based backends use the same shared library as python
```
tanmayv25, line 348: Python based backends

src/backend_manager.cc
```diff
@@ -338,19 +338,31 @@ Status
...
+    if ((*backend)->Name() == name) {
+      return Status::Success;
+    }
   }
 
   RETURN_IF_ERROR(TritonBackend::Create(
       name, dir, libpath, backend_cmdline_config, backend));
-  backend_map_.insert({libpath, *backend});
+
+  (*backend)->SetPythonBackendBasedFlag(is_python_backend_based);
```
tanmayv25, line 360: SetPythonBasedBackendFlag

src/backend_manager.cc
```diff
@@ -338,19 +338,31 @@ Status
...
+  if (is_python_backend_based) {
```
tanmayv25, line 361: is_python_based_backend

src/backend_manager.h
```diff
@@ -135,6 +141,8 @@ class TritonBackend {
   // Full path to the backend shared library.
   const std::string libpath_;
 
+  bool is_python_backend_based_;
```
tanmayv25, line 144: is_python_based_backend_

rmccorm4 (commented):
Looks a lot cleaner to me overall compared to iteration 1. 

As mentioned offline, my only open question is if using `Backend==vllm` and `SpecializedBackend==python` would simplify any checks or logic. Didn't look too deeply into it, just something to consider. 

Any other improvements can probably be iterated on if tests and general flow look good, since we're not adding any new C APIs to maintain.

src/backend_model.cc
```diff
@@ -308,6 +317,78 @@ TritonModel::GetExecutionPolicy(const inference::ModelConfig& model_config)
...
+                                       "' requires python backend to operate.");
+  }
+  search_paths.emplace_back(python_backend_dir);
+  std::string runtime_model_path =
+      JoinPath({global_backend_dir, backend_name, "model.py"});
+  bool exists;
+  RETURN_IF_ERROR(FileExists(runtime_model_path, &exists));
+  if (!exists) {
+    return Status(
+        Status::Code::INVALID_ARG,
```
oandreeva-nv, line 375: cc @tanmayv25 , here is the check I mentioned on python_backend PR

Tabrizian (approved):

PR #270, opened 2023-09-29
Fix state dimension selection when initial state is specified

Tabrizian:
When the state dimensions had variable dimensions, the initial state was not configured using the shape of the initial state and would just use 1 for variable dimensions which was not correct. e.g.

initial state: [1, 32, 2, 1]
model config: [1, 32, -1, 1]

The shape of the state was [1, 32, 1, 1] instead of [1, 32, 2, 1].

### DIFF
--- a/src/sequence_batch_scheduler/sequence_batch_scheduler.cc
+++ b/src/sequence_batch_scheduler/sequence_batch_scheduler.cc
@@ -367,6 +367,14 @@ SequenceBatchScheduler::GenerateInitialStateData(
             std::to_string(initial_state.dims().size()) +
             " != " + std::to_string(state.dims().size()));
   }
+  const auto& initial_state_pair = initial_state_.emplace(
+      std::piecewise_construct, std::forward_as_tuple(state.input_name()),
+      std::forward_as_tuple(initial_state.name()));
+  auto& initial_state_data = initial_state_pair.first->second;
+
+  if (max_batch_size_ > 0) {
+    initial_state_data.shape_.emplace_back(1);
+  }
 
   // Check the dimensions to make sure it doesn't have variable-sized dims and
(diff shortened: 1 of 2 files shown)


dyastremsky:
Thanks for making this fix! Do you have any testing associated with this ticket? It sounds like the behavior was not correct before. It would be good to have testing to confirm the behavior is correct after this change.

dyastremsky (dismissed):

rmccorm4 (dismissed):

src/sequence_batch_scheduler/sequence_batch_scheduler.cc
```diff
@@ -367,6 +367,14 @@ SequenceBatchScheduler::GenerateInitialStateData(
             std::to_string(initial_state.dims().size()) +
             " != " + std::to_string(state.dims().size()));
   }
+  const auto& initial_state_pair = initial_state_.emplace(
+      std::piecewise_construct, std::forward_as_tuple(state.input_name()),
+      std::forward_as_tuple(initial_state.name()));
+  auto& initial_state_data = initial_state_pair.first->second;
+
+  if (max_batch_size_ > 0) {
+    initial_state_data.shape_.emplace_back(1);
```
rmccorm4, line 376: Just sanity checking, for a request with batch size > 1, what is the expected behavior here?

Tabrizian: Sequence batcher doesn't allow requests with batch size larger than 1.

rmccorm4: That's what I figured, just double checking 🙏

rmccorm4 (approved):

PR #272, opened 2023-10-03
Add version info to prometheus metrics

ClifHouck:
Addresses https://github.com/triton-inference-server/server/issues/6320

Example metric output: 
`$ curl -s --request GET "http://localhost:8002/metrics" | grep api
triton_version_information{api_major_version="1",api_minor_version="16",server_major_version="1",server_minor_version="24"} 0`

### DIFF
--- a/src/metrics.cc
+++ b/src/metrics.cc
@@ -195,11 +195,6 @@ Metrics::Metrics()
           prometheus::BuildGauge()
             .Name("triton_version_information")
             .Help("Triton API & Server Versions")
-            /*.Labels({
-                //{std::string("api_major_version"), std::to_string(TRITONBACKEND_API_VERSION_MAJOR)},
-                {std::string("api_minor_version"),    std::to_string(TRITONBACKEND_API_VERSION_MINOR)},
-                {std::string("server_major_version"), std::to_string(TRITONSERVER_API_VERSION_MAJOR)},
-                {std::string("server_minor_version"), std::to_string(TRITONSERVER_API_VERSION_MINOR)}})*/
             .Register(*registry_)),
 
 #ifdef TRITON_ENABLE_METRICS_GPU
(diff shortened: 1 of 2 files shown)


dyastremsky:
Thanks for contributing this! Have you [submitted a CLA](https://github.com/triton-inference-server/server/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla) yet?

ClifHouck:
> Thanks for contributing this! Have you [submitted a CLA](https://github.com/triton-inference-server/server/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla) yet?

I think the CLA finally got sorted. I should be good to go.

ClifHouck:
@dyastremsky Is there anything I can do to further this PR?

dyastremsky:
Thank you for submitting your CLA, Clif! Apologies for the delay.

Ryan and Olga, do these look good to you? The original issue mentioned exposing the version info via Prometheus as one way of addressing the need to see Triton version info. Do we want these to be accessible via Prometheus?

src/metrics.cc
```diff
@@ -190,6 +190,13 @@ Metrics::Metrics()
                     "microseconds.")
               .Register(*registry_)),
 
+      // Versions
+      version_info_family_(
+          prometheus::BuildGauge()
+            .Name("triton_version_information")
```
rmccorm4, line 196: Hi @ClifHouck, thanks for the contrbution.

I reached out on the original issue to clarify the ask - I'm not sure if server/backend API versions are necessary if the actual triton release version is reflected and can be mapped.

However, it does raise a question that this info may be useful to add to an existing endpoint. Do you have any reasoning on exposing these through prometheus rather than another existing REST endpoint? Pros of adding it to a standard (non-metrics) endpoint is that it would always be available, regardless of whether the server was built or configured at runtime to use prometheus metrics or not.

PR #276, opened 2023-10-16
Dependency graph to track unloading nodes until unload completes

kthui:
Related PR: https://github.com/triton-inference-server/server/pull/6436

Previously, the model repository manager stops tracking a model once it is sure that the model will be unloaded. This behavior is problematic with the model lifecycle because it assumes a model cannot be loaded or unloaded concurrently, this guarantee is broken if models are not tracked while they are unloading. For example,
1. A model is unloading by the model lifecycle, while the model repository manager stops tracking it.
2. Without knowing the model is unloading by the model lifecycle, the model repository manager will be happy to load the same model again, while it is unloading by the model lifecycle.
3. The model lifecycle sees the same model has started loading while unloading has not yet completed, it sends an error on the unload process.

After the fix, the model repository manager will track any unloading model until it is unloaded. Thus, the overlapping between unloading and loading the same model on the model lifecycle cannot happen.

### DIFF
--- a/src/model_repository_manager.cc
+++ b/src/model_repository_manager.cc
@@ -2109,10 +2109,12 @@ ModelRepositoryManager::DependencyGraph::GetNode(
   const auto it = removed_nodes_.find(model_id);
   if (it != removed_nodes_.end()) {
     return it->second.get();
-  } else {
-    return FindNode(model_id, false /* allow_fuzzy_matching */);
   }
-  return nullptr;
+  auto* node = FindNode(model_id, false /* allow_fuzzy_matching */);
+  if (node == nullptr) {
+    throw std::invalid_argument("model_id '" + model_id.str() + "' not found");
+  }
+  return node;
 }
 
 std::unique_ptr<ModelIdentifier>
@@ -2156,15 +2158,24 @@ ModelRepositoryManager::DependencyGraph::Writeback(
     auto* node = GetNode(model_id);
     auto* updated_node = updated_dependency_graph.GetNode(model_id);
     // Writeback
-    node->status_ = updated_node->status_;
-    node->checked_ = updated_node->checked_;
-    node->loaded_versions_ = updated_node->loaded_versions_;
-    node->is_locked_ = updated_node->is_locked_;
-    // Notify retry(s)
-    node->retry_notify_cv_->notify_all();
+    node->Writeback(*updated_node);
     // Erase removed nodes at last to complete its lifecycle
     removed_nodes_.erase(model_id);
   }
+}
+
+
+void
+ModelRepositoryManager::DependencyNode::Writeback(
+    const DependencyNode& updated_dependency_node)
+{
+  // Writeback
+  status_ = updated_dependency_node.status_;
+  checked_ = updated_dependency_node.checked_;
+  loaded_versions_ = updated_dependency_node.loaded_versions_;
+  is_locked_ = updated_dependency_node.is_locked_;
+  // Notify retry(s)
+  retry_notify_cv_->notify_all();
 }
 
 
--- a/src/model_repository_manager.cc
+++ b/src/model_repository_manager.cc
@@ -767,9 +767,8 @@ ModelRepositoryManager::LoadUnloadModels(
 
   // Check for collision in affected models and try to lock those models.
   std::shared_ptr<std::condition_variable> retry_notify_cv;
-  auto conflict_model = new_dependency_graph.LockNodes(
-      affected_models, dependency_graph_ /* prev_dependency_graph */,
-      &retry_notify_cv);
+  auto conflict_model =
+      new_dependency_graph.LockNodes(affected_models, &retry_notify_cv);
   if (conflict_model) {
     // A collision is found. Since info is only written to local copy, so it is
     // safe to rollback by simply returning the conflict information.
@@ -1972,13 +1971,13 @@ ModelRepositoryManager::DependencyGraph::RemoveNode(
     downstreams.emplace(downstream->model_id_);
   }
 
-  // Drop the node from all reference,
-  // remove from graph at last to complete its lifecycle
+  // Drop the node from all references and only track it on removed nodes
   (*global_map_ptr_)[model_id.name_].erase(model_id);
   for (const auto& model_name : node->missing_upstreams_) {
     auto mit = missing_nodes_.find(model_name);
     mit->second.erase(model_id);
   }
+  removed_nodes_.emplace(std::make_pair(model_id, std::move(node)));
   nodes_.erase(it);
 
   return {std::move(upstreams), std::move(downstreams)};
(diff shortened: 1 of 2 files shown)


src/model_repository_manager.cc
```diff
@@ -2156,20 +2153,17 @@ ModelRepositoryManager::DependencyGraph::Writeback(
...
-      // Notify retry(s)
-      node->retry_notify_cv_->notify_all();
-    }
+    auto* node = GetNode(model_id);
+    auto* updated_node = updated_dependency_graph.GetNode(model_id);
+    // Writeback
+    node->status_ = updated_node->status_;
+    node->checked_ = updated_node->checked_;
+    node->loaded_versions_ = updated_node->loaded_versions_;
+    node->is_locked_ = updated_node->is_locked_;
```
rmccorm4, line 2162: Would a simple copy constructor override make sense here?

src/model_repository_manager.cc
```diff
@@ -2156,20 +2153,17 @@ ModelRepositoryManager::DependencyGraph::Writeback(
...
-      // Writeback
-      node->status_ = updated_node->status_;
-      node->checked_ = updated_node->checked_;
-      node->loaded_versions_ = updated_node->loaded_versions_;
-      node->is_locked_ = updated_node->is_locked_;
-      // Notify retry(s)
-      node->retry_notify_cv_->notify_all();
-    }
+    auto* node = GetNode(model_id);
+    auto* updated_node = updated_dependency_graph.GetNode(model_id);
```
rmccorm4, line 2157: Do we need to check `if (updated_node)` here for nullptr?

kthui: Not anymore.

Previously, because models can be removed and the same model may be added during remove, so the graphs could looks like this for example:
```
updated_dependency_graph:    [unmodified_model]
this->dependency_graph_:     [unmodified_model]    [previously_deleted_but_re_added_model]
```
The models contained on the `updated_dependency_graph` is always a subset of `this->dependency_graph_`.

Now, any deleted models are not removed from the dependency graph until they are completely unloaded, so any previously deleted model cannot be re-added until it is fully unloaded (in other words, no more load/unload overlapping). For example,
```
updated_dependency_graph:    [unmodified_model]    [deleted_model]
this->dependency_graph_:     [unmodified_model]    [deleted_model]
```
Thus, the `[previously_deleted_but_re_added_model]` cannot exist, so checking is no longer necessary.

src/model_repository_manager.cc
```diff
@@ -2095,38 +2099,36 @@ ModelRepositoryManager::DependencyGraph::Swap(DependencyGraph& rhs)
...
+ModelRepositoryManager::DependencyGraph::GetNode(
+    const ModelIdentifier& model_id) const
+{
+  const auto it = removed_nodes_.find(model_id);
+  if (it != removed_nodes_.end()) {
+    return it->second.get();
+  } else {
+    return FindNode(model_id, false /* allow_fuzzy_matching */);
+  }
+  return nullptr;
```
rmccorm4, line 2115:
```suggestion
  const auto it = removed_nodes_.find(model_id);
  if (it != removed_nodes_.end()) {
    return it->second.get();
  }
  return FindNode(model_id, false /* allow_fuzzy_matching */);
```

src/model_repository_manager.cc
kthui: Maybe a `Writeback()` for nodes, so the `// Writeback` and `// Notify retry(s)` parts can be grouped?

src/model_repository_manager.cc
rmccorm4: Having both of the pointers used here with the function that _can_ return nullptr, even if maybe not in this exact scenario makes me a bit uneasy. 

Do you think we could add some kind of check just in case for future stability if other parts of the code changes that behavior unknowingly?
```cpp
if (!node || !updated_node) {
  LOG_ERROR << "Received unexpected nullptr for dependency graph node: " << model_id << " " << node << " " << updated_node;
} else {
  // Writeback
  ...
}
```

src/model_repository_manager.cc
rmccorm4: Sure, whatever you think makes most sense. I think writeback for nodes aligns with current writeback for graph and modelinfo

src/model_repository_manager.cc
kthui: The reason why I did not add more checking is this is an unrecoverable error, meaning
```
if (nullptr is returned) {
  log error;
  skip working on nullptr objects;
}
...
```
will corrupt the model repository manager internal states, if allowed to move forward when `nullptr` is returned.

I think we could throw an exception if that happens and terminates the core, instead of a `nullptr` dereference. Given the ordering inside `GetNode()` is specific for the Lock/Unload/Writeback, maybe we can throw it there instead, so it will always return a valid pointer.
```
GetNode(model_id) {
  if (cannot find model from model_id) {
    throw std::invalid_argument("model_id is not on the graph!");
  }
}
```

src/model_repository_manager.cc
kthui: Nice catch!

kthui: Updated with the throw exception change.

src/model_repository_manager.cc
kthui: Updated with a `Writeback()` for nodes.

src/model_repository_manager.cc
rmccorm4: Exception sounds good to me if it shouldn't happen. Better to gracefully throw than hit an unexpected segfault IMO for our own sanity in the future.

src/model_repository_manager.cc
```diff
@@ -2095,38 +2099,38 @@ ModelRepositoryManager::DependencyGraph::Swap(DependencyGraph& rhs)
...
+ModelRepositoryManager::DependencyGraph::GetNode(
+    const ModelIdentifier& model_id) const
+{
+  const auto it = removed_nodes_.find(model_id);
+  if (it != removed_nodes_.end()) {
+    return it->second.get();
+  }
+  auto* node = FindNode(model_id, false /* allow_fuzzy_matching */);
+  if (node == nullptr) {
+    throw std::invalid_argument("model_id '" + model_id.str() + "' not found");
```
rmccorm4, line 2115: Out of curiosity, is there a good spot anywhere up the call chain to catch the exception and log an error? Or should this just crash the program?

rmccorm4:
Other than exception comment LGTM - but I think Guan should take a look in case.

src/model_repository_manager.cc
kthui: I just want to crash the program, because to recover from this failure, I think the model repository manager needs to be restarted, meaning unloading all the models and re-loading them again, which is not much difference from crashing the program and then restarting.

The good thing is, this should never happen in production under any circumstances, but only to make debugging easier if it happens (instead of debugging a `nullptr` dereference).

rmccorm4 (approved):

PR #277, opened 2023-10-17
Add reschedule flag for supporting generative sequence

GuanLuo:


include/triton/core/tritonserver.h
```diff
@@ -936,6 +937,13 @@ typedef enum tritonserver_responsecompleteflag_enum {
 ///     itself nor any input tensor data associated with the
 ///     request. The callback should free or otherwise manage the
 ///     'request' object and all associated tensor data.
+///   - TRITONSERVER_REQUEST_RELEASE_RESCHEDULE: This flag is currently being
+///     consumed internally and the callback is not expected to receive nor
+///     process this kind of release. The backend will call
+///     TRITONBACKEND_RequestRelease with this flag when it wish the request
```
Tabrizian, line 943:
```suggestion
///     TRITONBACKEND_RequestRelease with this flag when it wishes to reschedule the request
```

include/triton/core/tritonserver.h
```diff
@@ -936,6 +937,13 @@ typedef enum tritonserver_responsecompleteflag_enum {
...
+///     to be rescheduled back to the model. An example is that the model is
```
Tabrizian, line 944:
```suggestion
///   back to the model. An example is that the model is
```

src/sequence_batch_scheduler/sequence_utils.h
```diff
@@ -0,0 +1,83 @@
...
+    if (irequest->Flags() & TRITONSERVER_REQUEST_FLAG_SEQUENCE_START) {
+      irequest->AddInternalReleaseCallback(std::move(callback));
+    }
+  }
+
+  void RescheduleRequest(
+      std::unique_ptr<InferenceRequest>& request, const uint32_t flags);
+
+ private:
+  std::atomic<uint64_t> sequence_id_{1};
```
Tabrizian, line 79: Would this internal sequence id be a problem if the server is running for a long time? Or it would simply reset to 0 after overflow which is fine.

GuanLuo: Overflow to 0 will cause the particular request to fail (sequence batcher expects non-zero sequence ID), but overflowing on 64 bit type is unlikely based on calculation.

Tabrizian (approved):

Tabrizian (approved):

krishung5 (approved):

PR #282, opened 2023-11-01
Add CUDA block manager for implicit state

Tabrizian:


src/sequence_state.cc
```diff
@@ -96,13 +107,35 @@ SequenceStates::Initialize(
...
+                        "enable 'use_single_buffer' for state input name '") +
+                state_config.input_name() + ("'."));
+      }
+      if (kind != TRITONSERVER_INSTANCEGROUPKIND_GPU) {
+        return Status(
+            Status::Code::INVALID_ARG,
+            std::string("'use_growable_memory' can only work with KIND_GPU "
+                        "model instances right now. Please either change the "
+                        "instance type to KIND_GPU or set "
+                        "'use_growable_memory' to false for state input '") +
```
rmccorm4, line 135: Should this be enforced to gpu instances only? What about KIND_MODEL? Or even KIND_CPU that requests cpu memory but gets gpu memory back anyways? Just curious if instance kind should enforce it here.

rmccorm4: Is TRT-LLM KIND_GPU or KIND_MODEL?

src/sequence_state.cc
```diff
@@ -300,7 +370,8 @@ SequenceStates::CopyAsNull(const std::shared_ptr<SequenceStates>& from)
           std::forward_as_tuple(from_input_state_tensor->Name()),
           std::forward_as_tuple(new SequenceState(
               from_input_state_tensor->Name(), from_input_state_tensor->DType(),
-              from_input_state_tensor->Shape())));
+              from_input_state_tensor->Shape(), false /* reused buffer */,
+              false /* use growable memory */)));
```
rmccorm4, line 374: nit: match var names more closely for Ctrl-F/grep hits when looking through the code
```suggestion
              from_input_state_tensor->Shape(), false /* reuse_buffer */,
              false /* use_growable_memory */)));
```

src/test/memory_test.cc
```diff
@@ -393,6 +395,72 @@ TEST_F(AllocatedMemoryTest, Release)
...
+    cudaFree(0);
+    cuInit(0);
+    auto status = tc::CudaBlockManager::Create(6.0);
+    EXPECT_TRUE(status.IsOk()) << status.Message();
+  }
+
+  void TearDown() override {}
+};
+
+TEST_F(GrowableMemoryTest, AllocGPU)
```
rmccorm4, line 412: Can we have some kind of simple test that requests CPU memory and verifies that it gets back GPU memory (or fails)? Would be a good placeholder for testing if we eventually add CPU, and will fail if we try to make those changes in the future without updating the testing.

src/sequence_state.cc
```diff
@@ -118,23 +151,43 @@ SequenceStates::Initialize(
...
-      memcpy(
-          dst_buffer, initial_state_buffer,
-          initial_state_it->second.data_->TotalByteSize());
+      if (use_growable_memory) {
+        // TODO: use a constant or a flag for virtual address space size.
+        std::unique_ptr<GrowableMemory> growable_memory;
+        RETURN_IF_ERROR(GrowableMemory::Create(
+            initial_state_it->second.data_->TotalByteSize(),
+            TRITONSERVER_MEMORY_GPU, device_id, growable_memory,
+            1024 * 1024 * 1024));
```
rmccorm4, line 163: Can you add the constant and remove the TODO? `constexpr ... kDefaultVirtualGpuMemorySize = 1024*1024*1024;` or something like that?

src/sequence_state.cc
```diff
@@ -85,7 +90,13 @@ SequenceState::SetStringDataToZero()
...
       reinterpret_cast<const std::shared_ptr<MutableMemory>&>(Data());
   char* buffer = memory->MutableBuffer(&memory_type, &memory_type_id);
-  memset(buffer, 0, Data()->TotalByteSize());
+  if (memory_type == TRITONSERVER_MEMORY_GPU) {
+    memset(buffer, 0, Data()->TotalByteSize());
+  } else {
+    RETURN_IF_CUDA_ERR(
+        cudaMemset(buffer, 0, Data()->TotalByteSize()),
+        std::string("failed to set the data to zero."));
+  }
```
rmccorm4, line 99: are these conditions flipped? if GPU (cudaMemset) else (memset) ?

src/backend_model.cc
```diff
@@ -1367,36 +1367,40 @@ TRITONBACKEND_StateBuffer(
...
       std::shared_ptr<AllocatedMemory> memory =
           std::make_shared<AllocatedMemory>(
               buffer_byte_size, *memory_type, *memory_type_id);
       *buffer = memory->MutableBuffer(memory_type, memory_type_id);
       to->RemoveAllData();
       status = to->SetData(memory);
+      if (to->ReuseBuffer()) {
+        to->OtherState()->RemoveAllData();
+        to->OtherState()->SetData(memory);
+      }
```
rmccorm4, line 1402: Why isn't this section needed in UseGrowableMemory condition too? Doesn't growable memory re-use buffer by definition?

src/cuda_block_manager.cc
```diff
@@ -0,0 +1,149 @@
...
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+
+#include "cuda_block_manager.h"
+
+#include "cuda_utils.h"
+
+
+namespace triton { namespace core {
+std::unique_ptr<CudaBlockManager> CudaBlockManager::instance_;
```
rmccorm4, line 34: Is this fine as a global? Or should it belong to server object or similar?

src/cuda_block_manager.cc
```diff
@@ -0,0 +1,149 @@
...
+  std::lock_guard<std::mutex> lock(instance_->mu_);
+  if (instance_->free_blocks_.find(device_id) ==
+      instance_->free_blocks_.end()) {
+    return Status(
+        Status::Code::INTERNAL,
+        (std::string("Invalid device id '") + std::to_string(device_id) +
+         "' to allocate memory from."));
+  }
+
+  // Align the byte size to number of blocks.
```
rmccorm4, line 89: I think I saw some Cuda VMM restriction around alignment and granularity. Could you add a reference or more details on the alignment requirements somewhere in the code where relevant?

src/cuda_block_manager.cc
```diff
@@ -0,0 +1,149 @@
...
+
+  auto& free_blocks = instance_->free_blocks_[device_id];
+  free_blocks.insert(free_blocks.end(), blocks.begin(), blocks.end());
+
+  return Status::Success;
+}
+
+CudaBlockManager::~CudaBlockManager()
+{
+  for (auto& [device, free_blocks] : free_blocks_) {
```
rmccorm4, line 142: Isn't this C++17 structured bindings? I guess we're implicitly compiling with 17 on ubuntu 22.04.

rmccorm4: I'm curious if this part will compile on Windows. I think the rest of the VMM API stuff used here is cross-platform without the IPC components.

src/sequence_state.cc
```diff
@@ -146,15 +199,25 @@ SequenceStates::Initialize(
...
             triton::common::GetByteSize(state.second.data_type(), dims);
       }
-      data = std::make_shared<AllocatedMemory>(
-          state_size, TRITONSERVER_MEMORY_CPU, 0);
+      if (use_growable_memory) {
+        // TODO: use a constant or a flag for virtual address space size.
+        std::unique_ptr<GrowableMemory> growable_memory;
+        RETURN_IF_ERROR(GrowableMemory::Create(
+            state_size, TRITONSERVER_MEMORY_GPU, device_id, growable_memory,
+            1024 * 1024 * 1024));
```
rmccorm4, line 207: Same here https://github.com/triton-inference-server/core/pull/282/files#r1385781160

rmccorm4, line 207: If we stick with a constant default and not expose it as a flag for now, we should document it somewhere too, such as the model_config.proto in common or something.

src/sequence_state.h
```diff
@@ -65,6 +67,17 @@ class SequenceState {
   // The data for this shape.
   std::shared_ptr<Memory>& Data() { return data_; }
 
+  // A boolean indicating whether the state buffer should be reused or not.
+  bool ReuseBuffer() { return reuse_buffer_; }
```
GuanLuo, line 71: Why growable memory requires reuse buffer to be true?

From model config, reuse buffer refers to `use_single_buffer` which control whether the same buffer is used regardless of whether the state is accessed as input or as output. This looks like an optimization config to me (whether there is data copy to update state for input).

src/sequence_batch_scheduler.cc
```diff
@@ -1381,7 +1381,8 @@ SequenceBatch::UpdateImplicitState(
       sequence_states.reset(new SequenceStates);
       sequence_states->Initialize(
           base_->StateOutputConfigMap(), base_->MaxBatchSize(),
-          base_->InitialState());
+          base_->InitialState(), model_instance_->Kind(),
+          model_instance_->DeviceId());
```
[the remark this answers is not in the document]
Tabrizian: I think we can do something like that. I have filed this ticket as a follow up: https://jirasw.nvidia.com/browse/DLIS-5795

src/cuda_block_manager.h
```diff
@@ -0,0 +1,89 @@
...
+namespace triton { namespace core {
+
+class Allocation;
+
+///
+/// CudaBlockManager uses CUDA Virtual Memory Management APIs to allow the
+/// buffers to grow without having to create a new allocation. Currently,
+/// it is only used with implicit state management.
+///
+class CudaBlockManager {
```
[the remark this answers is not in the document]
Tabrizian: Yes, I think we can I added this information to the ticket we used to have for this topic: https://jirasw.nvidia.com/browse/DLIS-1242

src/cuda_block_manager.cc
Tabrizian: This is just to make the static instance work. I think it is still created per server object.

src/sequence_state.cc
Tabrizian: For this release we will be only supporting KIND_GPU. The issue is that we need to determine the initial state memory type and device ID for each instance. I have filed a ticket to investigate this:

https://jirasw.nvidia.com/browse/DLIS-5795

Tabrizian: It is KIND_CPU.

src/sequence_state.cc
Tabrizian: fixed

src/sequence_state.cc
Tabrizian: Great catch! Fixed.

src/backend_model.cc
Tabrizian: Growable memory uses the same "memory" for both input and output state. Since in Growable memory we're not creating a new buffer it is already pointing to the same buffer so this is not required.

src/cuda_block_manager.cc
Tabrizian: Good point fixed.

src/memory.h
```diff
@@ -172,4 +177,39 @@ class AllocatedMemory : public MutableMemory {
...
+  // satisfied, thus the function caller should always check the actual memory
+  // type and memory type id before use.
+  GrowableMemory(
+      size_t byte_size, TRITONSERVER_MemoryType memory_type,
+      int64_t memory_type_id, std::unique_ptr<Allocation>&& allocation,
+      size_t virtual_page_size);
+
+  static Status Create(
+      size_t byte_size, TRITONSERVER_MemoryType memory_type,
+      int64_t memory_type_id, std::unique_ptr<GrowableMemory>& growable_memory,
```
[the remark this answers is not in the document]
Tabrizian: Put it in the first param.

src/test/memory_test.cc
Tabrizian: Added.

src/sequence_state.cc
Tabrizian: It is now a flag.

src/cuda_block_manager.cc
Tabrizian: I added more description.

src/sequence_state.cc
Tabrizian: fixed.

Tabrizian: I added a flag.

src/sequence_state.h
```diff
@@ -65,6 +67,17 @@ class SequenceState {
...
+
+  // Use growable memory or not
+  bool UseGrowableMemory() { return use_growable_memory_; }
+
+  // Set pointer to the other sequence state
+  void SetOtherState(SequenceState* other) { other_state_ = other; }
+
+  SequenceState* OtherState() { return other_state_; }
```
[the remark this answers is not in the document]
Tabrizian: Added a comment.

src/sequence_state.h
Tabrizian: The main purpose of the growable buffer is to allow the backend to append data to an existing state buffer.

If we don't require use single buffer the backend would always have to write the full output state in each iteration since the output state buffer of step i is different that output state of step i+1. So we can technically allow something like this but I think it defeats the original purpose of this feature.

GuanLuo: Would it make more sense to only expose growable buffer configuration and make single buffer internal to growable buffer?

Another question is that I don't see why the user do not want to use single buffer in any case, which always saves a data copy within Triton if I understand it correctly.

Tabrizian: >Would it make more sense to only expose growable buffer configuration and make single buffer internal to growable buffer?

You mean to not expose single buffer optimization externally at all? I think there are scenarios that you want to use single buffer optimization alone just to reduce memory usage (e.g., CPU memory).

> Another question is that I don't see why the user do not want to use single buffer in any case, which always saves a data copy within Triton if I understand it correctly.

It doesn't save a data copy. It reduces the memory usage. But to your original question, I think the issue with always enabling this is that `TRITONBACKEND_StateUpdate` will no longer have an effect (becomes a no-op) which breaks backward compatibility.

include/triton/core/tritonserver.h
```diff
@@ -1855,6 +1855,18 @@ TRITONSERVER_DECLSPEC struct TRITONSERVER_Error*
 TRITONSERVER_ServerOptionsSetCudaMemoryPoolByteSize(
     struct TRITONSERVER_ServerOptions* options, int gpu_device, uint64_t size);
 
+/// Set the size of the virtual address space that will be used
+/// for growable memories in implicit state.
```
nnshah1, line 1859:
```suggestion
/// for growable memory in implicit state.
```

src/backend_model.cc
```diff
@@ -1356,36 +1356,40 @@ TRITONBACKEND_StateBuffer(
...
-    int64_t current_memory_type_id;
-    void* lbuffer =
-        memory->MutableBuffer(&current_memory_type, &current_memory_type_id);
-
-    // If the requested memory type doesn't match the current buffer, allocate
-    // a new buffer with the requested memory type and memory type id.
-    if (current_memory_type == *memory_type &&
-        current_memory_type_id == *memory_type_id) {
-      *buffer = lbuffer;
+  TRITONSERVER_MemoryType current_memory_type;
```
nnshah1, line 1359: Question: Could we refactor this into the SequenceState class?

Provide a "Resize()" method on SequenceState class that would internally handle the different types of state. 

That would also enable us to reduce the public interface of the class and handle them internally.

src/backend_model.cc
nnshah1: Two questions:
1) Can we expand growable memory to cases where input and output are different buffers? Seems like the two are independent -
2) in the case we don't use the same buffer - do we have to also increase the memory for both? (Is that handled somewhere else?)

src/cuda_block_manager.cc
nnshah1: The singleton would only be created once per program - so even if there were multiple server objects - you would have only one instance of the block manager.

I found this interesting singleton implementation using static within the create function itself: https://stackoverflow.com/a/1008289 

Either is fine - maybe for completeness I would explicitly initialzie insteance_ = nullptr.

If we need a seperate instance per server object - then we will need to create and store the instance in the server object .

src/cuda_block_manager.cc
```diff
@@ -0,0 +1,152 @@
...
+    size_t size, std::unique_ptr<Allocation>& allocation, int device_id)
+{
+  if (instance_ == nullptr) {
+    return Status(
+        Status::Code::INTERNAL,
+        (std::string("CUDA Block manager has not been created.")));
+  }
+
+  std::lock_guard<std::mutex> lock(instance_->mu_);
+  if (instance_->free_blocks_.find(device_id) ==
```
nnshah1, line 81: Another slightly less verbose way can be to use count since we do not use the iterator:

if (instance_->free_blocks.count(device_id) ==0))

src/cuda_block_manager.cc
```diff
@@ -0,0 +1,152 @@
...
+Allocation::Merge(std::unique_ptr<Allocation>&& allocation)
+{
+  for (auto block : allocation->Blocks()) {
+    blocks_.push_back(block);
+  }
+  allocation->Blocks().clear();
+}
+
+Status
+CudaBlockManager::Create(double min_supported_compute_capability)
```
nnshah1, line 46: I think it makes sense to support pre-allocation of the initial free block size

src/cuda_block_manager.cc
```diff
@@ -0,0 +1,152 @@
...
+    return Status(
+        Status::Code::INTERNAL,
+        (std::string("Invalid device id '") + std::to_string(device_id) +
+         "' to allocate memory from."));
+  }
+
+  // Align the byte size to number of blocks. The allocated bytes must be a
+  // multiple of the cudaAllocationGranularity. In the current implementation
+  // the block size is the cuda minimum allocation granularity.
+  size_t num_blocks =
```
nnshah1, line 92: Instead of a while loop checking the free block size on the iteration - you alternatively use up all the free blocks and then allocate any remaining (pseudo code):

```
  size_t free_blocks = instance->free_blocks_[device_id].size();
  for (auto block_index = 0; block_index < num_blocks && block_index <free_blocks; ++block_index) {
      allocation->AddBlock(instance_->free_blocks_[device_id].back());
      instance_->free_blocks_[device_id].pop_back();
  }
  for (;block_index <num_blocks;++block_index) {
  // allocate remaining
  }
```

src/cuda_block_manager.h
```diff
@@ -0,0 +1,89 @@
...
+ public:
+  static Status Create(double min_supported_compute_capability);
```
nnshah1, line 52: A few questions:
1) Generally for a singleton pattern - we would have a "GetInstance()" method that creates or returns the instance, and then all calls would be instance methods and not static methods. If we moved them to instance methods we could avoid the 'instance == null' check.
2) It seems odd that the Allocate() and Free() methods take different objects. Is there a reason why Free() wouldn't take an Allocation? 
3) Not sure I follow the reason the allocate takes a reference to a unique pointer - could it take a reference to an allocation instead?

src/memory.cc
```diff
@@ -236,4 +237,132 @@ AllocatedMemory::~AllocatedMemory()
...
+                    " than the virtual address size. byte_size: ") +
+            std::to_string(byte_size) +
+            ", virtual_address_size:" + std::to_string(virtual_address_size));
+  }
+
+  RETURN_IF_ERROR(
+      CudaBlockManager::Allocate(byte_size, allocation, memory_type_id));
+
+  void* buffer;
+  RETURN_IF_CUDA_DRIVER_ERR(
```
nnshah1, line 287: if there is an error - do we clean up allocated memory properly here?

src/memory.cc
```diff
@@ -236,4 +237,132 @@ AllocatedMemory::~AllocatedMemory()
...
+      cuMemAddressReserve(
+          reinterpret_cast<CUdeviceptr*>(&buffer), virtual_address_size,
+          0 /* alignment */, 0 /* start_address */, 0 /* flags */),
+      std::string("cuMemAddressReserve failed"));
+  growable_memory = std::move(std::make_unique<GrowableMemory>(
```
nnshah1, line 292: could we use:
```
growable_memory.reset() 
```
?

src/memory.cc
```diff
@@ -236,4 +237,132 @@ AllocatedMemory::~AllocatedMemory()
...
+      CudaBlockManager::BlockSize() * allocation->Blocks().size(), memory_type,
+      memory_type_id, std::move(allocation), virtual_address_size));
+  growable_memory->buffer_ = reinterpret_cast<char*>(buffer);
+
+  for (auto& block : growable_memory->allocation_->Blocks()) {
```
nnshah1, line 297: could this be done in the constructor?

src/memory.h
```diff
@@ -172,4 +177,39 @@ class AllocatedMemory : public MutableMemory {
...
+  Status Map(CUmemGenericAllocationHandle& block);
+
+  ~GrowableMemory() override;
+
+  Status Resize(size_t size);
+
+  std::unique_ptr<Allocation>& GetAllocation();
+
+ private:
+  std::unique_ptr<Allocation> allocation_;
```
nnshah1, line 207: Does allocation need to be a unique ptr? could it be a instance variable instead - and initialized in the constructor?
from my understanding the allocation is empty to begin with and just has the device id?

src/sequence_state.cc
```diff
@@ -146,15 +199,24 @@ SequenceStates::Initialize(
         state_size =
             triton::common::GetByteSize(state.second.data_type(), dims);
       }
-      data = std::make_shared<AllocatedMemory>(
-          state_size, TRITONSERVER_MEMORY_CPU, 0);
+      if (use_growable_memory) {
+        std::unique_ptr<GrowableMemory> growable_memory;
+        RETURN_IF_ERROR(GrowableMemory::Create(
```
nnshah1, line 204: Could we remove create and use a constructor for growable memory?

src/sequence_state.h
nnshah1: @Tabrizian ,

I'm wondering if it should be growable automatically if you are using CUDA memory?

My thought is:

1) Input / Output having the same buffer - that seems definitely like an option that a backend may or may not support. That is it's not clear to me that input and output wouldn't be read and written in parallel under some models / architectures. The other option is we can require that they can't be read and written to at the same time and deprecate stateupdate.

2) Growing memory vs re-allocating and copying is transparent to the functionality - so could always be enabled if supported for a buffer type?

src/sequence_state.cc
```diff
@@ -85,7 +90,13 @@ SequenceState::SetStringDataToZero()
   const std::shared_ptr<MutableMemory>& memory =
       reinterpret_cast<const std::shared_ptr<MutableMemory>&>(Data());
   char* buffer = memory->MutableBuffer(&memory_type, &memory_type_id);
-  memset(buffer, 0, Data()->TotalByteSize());
+  if (memory_type == TRITONSERVER_MEMORY_GPU) {
+    RETURN_IF_CUDA_ERR(
+        cudaMemset(buffer, 0, Data()->TotalByteSize()),
```
nnshah1, line 95: can we use:
```
 memory->TotalByteSize()
```

src/sequence_state.cc
```diff
@@ -85,7 +90,13 @@ SequenceState::SetStringDataToZero()
...
+        std::string("failed to set the data to zero."));
+  } else {
+    memset(buffer, 0, Data()->TotalByteSize());
```
nnshah1, line 98: It almost seems like ClearMemory / SetMemory should be methods on the memory object and it would encapsulate the difference for different memory types.

It's odd for the sequence state to have to know details about the type of memory?

src/sequence_state.cc
```diff
@@ -118,23 +152,42 @@ SequenceStates::Initialize(
...
-      memcpy(
-          dst_buffer, initial_state_buffer,
-          initial_state_it->second.data_->TotalByteSize());
+      if (use_growable_memory) {
+        std::unique_ptr<GrowableMemory> growable_memory;
+        RETURN_IF_ERROR(GrowableMemory::Create(
+            growable_memory, initial_state_it->second.data_->TotalByteSize(),
+            TRITONSERVER_MEMORY_GPU, device_id,
+            cuda_virtual_address_size.at(device_id)));
+        data = std::move(growable_memory);
```
nnshah1, line 164: I'm a little confused at what is happening here:

1) we move the growable_memory to data (this implies that the contents of growable_memory are now moved to data) 

2) we get the buffer from data and call this dst_buffer

3) we copy from the dst_buffer into the growable_memory MutableBuffer()

Isn't dst_buffer and growable_memory.MutableBuffer() the same (since we moved growable_buffer into data?)

src/sequence_state.cc
```diff
@@ -226,55 +288,62 @@ SequenceStates::OutputState(
   auto input_states_itr = input_states_.begin();
   std::advance(input_states_itr, iter_advance);
   auto& input_state_r = input_states_[input_states_itr->first];
+  bool reuse_buffer = output_states_[name]->ReuseBuffer();
```
nnshah1, line 291: in some places we say "use single buffer" in other places we use 'reuse_buffer"
could we normalize the boolean name?

src/sequence_state.cc
```diff
@@ -226,55 +288,62 @@ SequenceStates::OutputState(
...
-    return Status::Success;
-  });
+  output_state_r->SetStateUpdateCallback(
+      [&output_state_r, &input_state_r, reuse_buffer]() {
+        // Swap the internal memory if the size of the input and output state is
+        // equal
+
+        if (output_state_r->Data()->TotalByteSize() ==
+            input_state_r->Data()->TotalByteSize()) {
+          if (!reuse_buffer) {
```
nnshah1, line 304: can we move this condition to the outside?
```
if (!reuse_buffer) {
  if (output_state ==)
}
```

src/sequence_state.cc
```diff
@@ -226,55 +288,62 @@ SequenceStates::OutputState(
...
+            std::shared_ptr<Memory> temp_memory = input_state_r->Data();
+            RETURN_IF_ERROR(input_state_r->RemoveAllData());
+            RETURN_IF_ERROR(input_state_r->SetData(output_state_r->Data()));
+            RETURN_IF_ERROR(output_state_r->RemoveAllData());
+            RETURN_IF_ERROR(output_state_r->SetData(temp_memory));
+          }
+        } else {
+          if (!reuse_buffer) {
```
nnshah1, line 312: see above - seems like we don't execute either the if or else clause if we are reusing the buffer - can we move that check outside?

src/sequence_state.h
```diff
@@ -44,10 +44,12 @@ class SequenceState {
   SequenceState();
   SequenceState(
       const std::string& name, const inference::DataType datatype,
-      const std::vector<int64_t>& shape);
+      const std::vector<int64_t>& shape, bool reuse_buffer,
```
nnshah1, line 47: should this be "use single buffer"

src/sequence_state.cc
Tabrizian: The reason we're using Create is to communicate any error that might have happened during the creation process. It is not possible to return an error using constructor.

src/sequence_state.cc
Tabrizian: they're exactly the same object.

src/sequence_state.cc
Tabrizian: It is not the details/differences of memory type. The reason that we have this in SequenceState is that this special condition is only required for SequenceState with String datatype. By default for initial state we can have garbage data which is not acceptable for string data type (since the first element is the number of bytes which means only certain values are allowed for the rest of the buffer).

I see your point regarding having utilities in Memory to clear the memory though and I can make that change.

src/memory.cc
Tabrizian: Since we want to be able to return an error it is not possible to do this in constructor.

src/memory.cc
Tabrizian: Yes, since Allocation is a unique_ptr it will free all the memory upon destruction (i.e., when we return from this function).

src/sequence_state.h
Tabrizian: 1. yes, exactly, that's what I mentioned to you earlier.
2. I think there are still some edge cases that we want this to be an option. For example, the minimum allocation for growable memory is 2MBs (depending on the systems) which could be a very large amount esp if you have a lot of sequences. So, I think the user should only use this option for sequences with large memory.

src/cuda_block_manager.h
Tabrizian: 1. I agree but I followed the existing pattern that exists in `cuda_memory_manager`. I have filed a ticket to revisit this in the future.
2. I think we should be able to make that change.
3. It is just that the move semantics for `Allocation` are clearer in my head and I don't have to worry about the copies (since unique ptr doesn't allow copies). Happy to change it to a reference if you think it is less confusing.

src/cuda_block_manager.cc
Tabrizian: Yeah, I just followed what we have in the cuda_memory_manager. We follow the pattern mentioned in stack overflow in Python backend for singleton: https://github.com/triton-inference-server/python_backend/blob/889585ce0fc4d4c88eeb69dddbe0f26d08af4b6f/src/pb_utils.h#L262-L266

src/cuda_block_manager.cc
Tabrizian: I'm still not super clear on the benefits of supporting pre-allocation. Do you mind elaborating more? I think it only provides marginal benefits compared to the current scheme.

src/backend_model.cc
Tabrizian: I think growable memory doesn't provide a lot of value if you are not using a single buffer. The main goal of the growable memory is to provide speedups for the cases where you're able to append state to the existing state object. When you're using using a single state buffer and the state size changes, you have to copy the whole state even if you have an option to grow an existing allocation. I think in that scenario, just getting a new allocation isn't any different than getting a new buffer from user's perspective.

src/memory.cc
Tabrizian: fixed.

src/backend_model.cc
Tabrizian: Done.

src/sequence_state.cc
Tabrizian: fixed.

src/sequence_state.cc
Tabrizian: great catch! Fixed.

rmccorm4 (commented):
I'll defer to Neelay/Guan for most remaining comments

src/backend_model.cc
nnshah1: won't it cut down memory copies in a similar way to single buffer?

1) Request larger state buffer size
2) append data
3) switch input and output buffers
4) append data to output buffer for next update

include/triton/core/tritonserver.h
```diff
@@ -1855,6 +1855,18 @@ TRITONSERVER_DECLSPEC struct TRITONSERVER_Error*
...
 
+/// Set the size of the virtual address space that will be used
+/// for growable memory in implicit state.
+///
+/// \param options The server options object.
+/// \param gpu_device The GPU device to set the CUDA virtual address space size
+/// \param size The size of the CUDA virtual address space.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC TRITONSERVER_Error*
+TRITONSERVER_ServerOptionsSetCudaVirtualAddressSize(
```
nnshah1, line 1866: I believe because we have a singleton - we would only be able to set this once - for the future we would want to see if we can allow multiple server objects with different options (can be handled in the future).

src/backend_model.cc
```diff
@@ -1356,42 +1356,26 @@ TRITONBACKEND_StateBuffer(
...
+      current_memory_type == *memory_type &&
+      current_memory_type_id == *memory_type_id) {
+    *buffer = lbuffer;
   } else {
-    std::shared_ptr<AllocatedMemory> memory = std::make_shared<AllocatedMemory>(
-        buffer_byte_size, *memory_type, *memory_type_id);
-    *buffer = memory->MutableBuffer(memory_type, memory_type_id);
-    to->RemoveAllData();
-    status = to->SetData(memory);
+    RETURN_TRITONSERVER_ERROR_IF_ERROR(
```
nnshah1, line 1375: Question: If the memory type changes does Resize handle this?

src/cuda_block_manager.h
```diff
@@ -0,0 +1,90 @@
...
+  static Status Allocate(
```
nnshah1, line 53: would add comments for methods to describe high level what they do

src/cuda_utils.cc
```diff
@@ -258,6 +263,255 @@ SupportsIntegratedZeroCopy(const int gpu_id, bool* zero_copy_support)
...
+  prop.location.type = CU_MEM_LOCATION_TYPE_DEVICE;
+  prop.location.id = 0;
+
+  RETURN_IF_ERROR(CudaDriverHelper::GetInstance().CuMemGetAllocationGranularity(
+      &aligned_sz, &prop, CU_MEM_ALLOC_GRANULARITY_MINIMUM));
+  return Status::Success;
+}
+
+CudaDriverHelper::CudaDriverHelper()
+{
```
nnshah1, line 280: initialize dl_open_handle_ to nullptr for completeness

src/cuda_utils.h
```diff
@@ -118,6 +134,36 @@ Status GetSupportedGPUs(
...
+/// Set the CUDA context to the specified device ID
+/// It will rollback to the previous device upon destruction.
+class ScopedSetDevice {
+ public:
+  ScopedSetDevice(int device)
+  {
+    overriden_ = false;
+
+    prev_device_ = device;
+    cudaGetDevice(&prev_device_);
```
nnshah1, line 148: question: is device process wide or thread local? if process wide is there anything we can / need to do for thread safety?

If one thread sets a device and another switches it is there any error handling we can do?

src/memory.h
```diff
@@ -146,6 +151,9 @@ class MutableMemory : public Memory {
       TRITONSERVER_MemoryType* memory_type = nullptr,
       int64_t* memory_type_id = nullptr);
 
+  // Set the memory to the specified value.
+  Status SetMemory(int value);
```
nnshah1, line 155: small nit: while the value of the underlying apis use an int but convert to unsigned char - maybe we define this api using an unsigned char to be more precise -

src/memory.cc
```diff
@@ -236,4 +264,133 @@ AllocatedMemory::~AllocatedMemory()
...
+
+  RETURN_IF_ERROR(
+      CudaBlockManager::Allocate(byte_size, allocation, memory_type_id));
+
+  void* buffer;
+  RETURN_IF_ERROR(CudaDriverHelper::GetInstance().CuMemAddressReserve(
+      reinterpret_cast<CUdeviceptr*>(&buffer), virtual_address_size,
+      0 /* alignment */, 0 /* start_address */, 0 /* flags */));
+  growable_memory.reset(new GrowableMemory(
+      CudaBlockManager::BlockSize() * allocation->Blocks().size(), memory_type,
```
nnshah1, line 326: is this byte_size from above?

src/memory.h
```diff
@@ -172,4 +180,39 @@ class AllocatedMemory : public MutableMemory {
...
+      size_t byte_size, TRITONSERVER_MemoryType memory_type,
+      int64_t memory_type_id, std::unique_ptr<Allocation>&& allocation,
+      size_t virtual_page_size);
+
+  static Status Create(
+      std::unique_ptr<GrowableMemory>& growable_memory, size_t byte_size,
+      TRITONSERVER_MemoryType memory_type, int64_t memory_type_id,
+      size_t virtual_page_size);
+
+  Status Map(CUmemGenericAllocationHandle& block);
```
nnshah1, line 201: can map be private?

src/sequence_state.cc
```diff
@@ -79,30 +84,73 @@ SequenceState::SetStringDataToZero()
...
+    RETURN_IF_ERROR(growable_memory->Resize(buffer_byte_size));
+    *buffer = growable_memory->MutableBuffer(memory_type, memory_type_id);
+  } else {
+    std::shared_ptr<AllocatedMemory> memory = std::make_shared<AllocatedMemory>(
+        buffer_byte_size, *memory_type, *memory_type_id);
+    *buffer = memory->MutableBuffer(memory_type, memory_type_id);
+    RemoveAllData();
+    RETURN_IF_ERROR(SetData(memory));
+    if (UseSingleBuffer()) {
+      OtherState()->RemoveAllData();
```
nnshah1, line 111: do we need to check error code here?

nnshah1, line 111: for growable memory we are able to change the underlying memory for both input and output in one step transparently - could we do the same for allocated memory? Is there a way to have the two states input / output share a pointer to the same underlying object and hide that detail?

src/sequence_state.cc
```diff
@@ -226,55 +305,60 @@ SequenceStates::OutputState(
...
+            const std::shared_ptr<MutableMemory>& input_memory =
+                reinterpret_cast<const std::shared_ptr<MutableMemory>&>(
+                    input_state_r->Data());
+
+            input_memory->MutableBuffer(&memory_type, &memory_type_id);
+            std::shared_ptr<AllocatedMemory> memory =
+                std::make_shared<AllocatedMemory>(
+                    output_state_r->Data()->TotalByteSize(), memory_type,
+                    memory_type_id);
+            RETURN_IF_ERROR(input_state_r->RemoveAllData());
```
nnshah1, line 343: possible to combine these into a one operaton - maybe SetData can remove data internaly?

src/sequence_state.h
```diff
@@ -121,7 +144,9 @@ class SequenceStates {
           std::string, const inference::ModelSequenceBatching_State&>&
           state_output_config_map,
       const size_t max_batch_size,
-      const std::unordered_map<std::string, InitialStateData>& initial_state);
+      const std::unordered_map<std::string, InitialStateData>& initial_state,
+      TRITONSERVER_InstanceGroupKind kind, int32_t device_id,
+      const std::map<int, size_t>& cuda_virtual_address_size);
```
nnshah1, line 149: cuda_vitrual_address_size - is this set per model or per server?

src/cuda_utils.h
Tabrizian: It is per thread: https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g159587909ffa0791bbe4b40187a4c6bb

src/cuda_utils.cc
Tabrizian: fixed.

src/cuda_block_manager.cc
```diff
@@ -0,0 +1,153 @@
...
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+
+#include "cuda_block_manager.h"
+
+#include "cuda_utils.h"
+
+
+namespace triton { namespace core {
+std::unique_ptr<CudaBlockManager> CudaBlockManager::instance_;
```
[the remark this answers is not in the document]
Tabrizian: fixed.

src/cuda_block_manager.h
Tabrizian: Used the same object for both Allocate and Free.

src/sequence_state.h
Tabrizian: discussed offline it is per server.

src/sequence_state.cc
Tabrizian: will be addressed as a part of memory refactors.

src/sequence_state.cc
Tabrizian: will be addressed as a part of memory refactors.

Tabrizian: added

src/memory.h
Tabrizian: fixed.

src/memory.h
Tabrizian: fixed.

include/triton/core/tritonserver.h
Tabrizian: discussed offline will be part of future changes.

src/cuda_block_manager.cc
```diff
@@ -0,0 +1,153 @@
...
+        (std::string("CUDA Block manager has already been created.")));
+  }
+
+  std::set<int> supported_gpus;
+  RETURN_IF_ERROR(
+      GetSupportedGPUs(&supported_gpus, min_supported_compute_capability));
+
+  size_t block_size;
+  RETURN_IF_ERROR(GetAllocationGranularity(block_size));
+  instance_.reset(new CudaBlockManager());
```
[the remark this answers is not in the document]
Tabrizian: As discussed for error handling purposes.

src/cuda_block_manager.h
```diff
@@ -0,0 +1,90 @@
...
+  void AddBlock(CUmemGenericAllocationHandle block)
+  {
+    blocks_.push_back(block);
+  }
+
+  void Merge(std::unique_ptr<Allocation>&& allocation);
+
+  std::vector<CUmemGenericAllocationHandle>& Blocks() { return blocks_; }
+
+  ~Allocation() { CudaBlockManager::Free(std::move(blocks_), device_id_); }
```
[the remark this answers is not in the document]
Tabrizian: will be part of future changes to Allocation object.

src/memory.h
Tabrizian: will be part of future changes for Allocation object.

PR #283, opened 2023-11-01
Allow backends to auto-complete model transaction policy

oandreeva-nv:
To set model transaction policy from Python backend's autocomple function, I need to add this field on core-side config, when `TritonModel::UpdateModelConfig` is called. `TritonModel::UpdateModelConfig` seem to be called only for `auto_complete` tasks.


dyastremsky (approved):

krishung5 (approved):

tanmayv25 (approved):

tanmayv25:
@oandreeva-nv  You might want to update the title of the PR to remove the Oandreeva. 
Also, try to make the the title of the PRs more generic.. Such as in this case: `Allow backends to auto-complete model transaction policy`.
In the description, you can call out that this is needed for vLLM auto-complete support.

PR #284, opened 2023-11-03
Don't allow file override outside of model directory in model load API

rmccorm4:
Server/test PR: https://github.com/triton-inference-server/server/pull/6516

```
E1107 03:19:14.402174 16541 model_repository_manager.cc:1323] Poll failed for model directory 'new_model': Invalid file parameter 'file:../../root/new_dir/test.txt' with normalized dir '/root/new_dir' must stay within model directory.
E1107 03:19:14.403816 16541 model_repository_manager.cc:1323] Poll failed for model directory 'new_model': Invalid file parameter 'file:../escape_symlink/new_dir/test.txt' with normalized dir '/root/new_dir' must stay within model directory.
E1107 03:19:14.405993 16541 model_repository_manager.cc:1323] Poll failed for model directory 'new_model': Invalid file parameter 'file:../../root/.bashrc' with normalized dir '/root' must stay within model directory.
E1107 03:19:14.408064 16541 model_repository_manager.cc:1323] Poll failed for model directory 'new_model': Invalid file parameter 'file:../escape_symlink/.bashrc' with normalized dir '/root' must stay within model directory.
```

:warning: **boost::filesystem dependency**

NOTE: `boost::filesystem::weakly_canonical` is used to get an "absolute/normalized" path for a path that **may not yet exist**, and **is cross-platform**. This is available in C++17, but to avoid risk of upgrading C++ version right before code freeze, I'd like to temporarily introduce the boost dependency. After code freeze, I would like to push the C++17 upgrade that @oandreeva-nv started through and then replace this usage with built-in `std:filesystem` instead.

Some other notes on alternatives:
- `realpath` (unix-only) does not support paths that don't yet exist, and may require more complex logic to write by hand. 
- [PathCanonicalize](https://learn.microsoft.com/en-us/windows/win32/api/shlwapi/nf-shlwapi-pathcanonicalizea) is a Windows specific version of something similar to `realpath` that may work if there are any windows issues with boost that take too long to resolve.

### DIFF
--- a/src/model_repository_manager.cc
+++ b/src/model_repository_manager.cc
@@ -129,11 +129,6 @@ class LocalizeRepoAgent : public TritonRepoAgent {
                         .c_str());
               }
 
-
-              // TODO: Remove
-              std::cout << "[DEBUG] dirname: " << dir << std::endl;
-              std::cout << "[DEBUG] weakly_canonical dirname: " << realdir
-                        << std::endl;
               bool dir_exist = false;
               RETURN_TRITONSERVER_ERROR_IF_ERROR(FileExists(dir, &dir_exist));
               if (dir_exist) {
--- a/src/model_repository_manager.cc
+++ b/src/model_repository_manager.cc
@@ -118,9 +118,11 @@ class LocalizeRepoAgent : public TritonRepoAgent {
               // Resolve any relative paths or symlinks, and enforce that target
               // directory stays within model directory for security.
               // DLIS-5149: Can use std::filesystem over boost in C++17.
-              const std::string& real_dir =
+              const std::string& canonical_tmp_dir =
+                  boost::filesystem::weakly_canonical(temp_dir).string();
+              const std::string& canonical_override_dir =
                   boost::filesystem::weakly_canonical(dir).string();
-              if (real_dir.rfind(temp_dir, 0) != 0) {
+              if (canonical_override_dir.rfind(canonical_tmp_dir, 0) != 0) {
                 return TRITONSERVER_ErrorNew(
                     TRITONSERVER_ERROR_INVALID_ARG,
                     (std::string("Invalid file parameter '") + file->Name() +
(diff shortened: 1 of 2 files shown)


nnshah1 (dismissed):

Tabrizian (approved):

src/model_repository_manager.cc
```diff
@@ -109,12 +110,24 @@ class LocalizeRepoAgent : public TritonRepoAgent {
...
+              // temporary model directory as the basepath.
               const std::string file_path =
                   JoinPath({temp_dir, file->Name().substr(file_prefix.size())});
               const std::string dir = DirName(file_path);
+              // Resolve any relative paths or symlinks, and enforce that target
+              // directory stays within model directory for security.
+              // DLIS-5149: Can use std::filesystem over boost in C++17.
+              const std::string& real_dir =
+                  boost::filesystem::weakly_canonical(dir).string();
+              if (real_dir.rfind(temp_dir, 0) != 0) {
```
nnshah1, line 123: question: do we need to also get the real_dir of the temp_dir to be safe?

nnshah1 (approved):

src/model_repository_manager.cc
```diff
@@ -109,12 +110,26 @@ class LocalizeRepoAgent : public TritonRepoAgent {
...
                         .c_str());
               }
-
-              // Save model file to the instructed directory
-              // mkdir
+              // Save model override file to the instructed directory using the
+              // temporary model directory as the basepath.
               const std::string file_path =
                   JoinPath({temp_dir, file->Name().substr(file_prefix.size())});
               const std::string dir = DirName(file_path);
```
nnshah1, line 117: small nit:

instead of creating new variables, could we update the existing?
```
const std::string & dir = boost::filestystem::weakly_canonical(DirName(file_path)).string();

const std::string &temp_dir = boost::filessystem::weakly_canonical(temp_dir_cstr).string();

```

nnshah1 (dismissed):
one small suggestion (but can be follow on / ignored)

nnshah1 (dismissed):

Tabrizian (dismissed):

rmccorm4:
Passes local build and CI on Linux x86 with Neelay's changes.

nnshah1 (approved):

PR #285, opened 2023-11-03
Incorporate runtime into model configuration

kthui:
Related PRs:
* https://github.com/triton-inference-server/server/pull/6518
* https://github.com/triton-inference-server/common/pull/103
* https://github.com/triton-inference-server/backend/pull/94
* https://github.com/triton-inference-server/pytorch_backend/pull/117

Previously, the runtime for each backend is determined automatically when loading the backend. The PR adds the ability for user to choose which runtime to use for each model on the model configuration. For example:
```
runtime: "libtriton_tensorrt.so"
```
```
runtime: "model.py"
```

To retain backward compatibility, the logic previously used to determine runtime is incorporated into autocomplete, which will attempt to fill-in the "runtime" field if left empty. The core will load the exact runtime as specified on the "runtime" field on the model configuration, after any applicable autocomplete.

As a part of the change, the autocomplete logic will fill in Python backend based runtime for backends that are clearly Python backend based backends, for example, vLLM backend.

If a backend provides both C++ and Python backend based runtime, for example, PyTorch backend, the autocomplete will look at the default model filename on the model configuration to try determine whether the C++ or Python backend based runtime is more appropriate. For example, if the default model filename on PyTorch backend is "model.pt", the C++ runtime will be selected. If the default model filename is "model.py", the Python backend based runtime will be selected. In any case, the autocomplete will not alter the runtime if it is explicitly provided by the user.


src/backend_model.cc
```diff
@@ -103,62 +102,35 @@ TritonModel::Create(
...
-      JoinPath({localized_model_path, std::to_string(version)});
-  const std::string global_path =
-      JoinPath({backend_dir, specialized_backend_name});
-  std::vector<std::string> search_paths = {
-      version_path, localized_model_path, global_path};
-
-  std::string backend_libdir;
-  std::string backend_libpath;
-  std::string python_runtime_modeldir;
+  bool is_python_backend_based_backend = false;
```
nnshah1, line 105:
```suggestion
  bool is_python_based_backend = false;
```

kthui: I was thinking about `python_binding_based_backend`, but simplicity is better. Updated: [Rename variable name](https://github.com/triton-inference-server/core/pull/285/commits/7da30c953bf5e058b978ff52a583fb061844fe03)

src/constants.h
```diff
@@ -55,11 +55,14 @@ constexpr char kOpenVINORuntimeBackend[] = "openvino";
 
 constexpr char kPyTorchLibTorchPlatform[] = "pytorch_libtorch";
 constexpr char kPyTorchLibTorchFilename[] = "model.pt";
+constexpr char kPyTorchPythonPlatform[] = "pytorch_python";
```
rmccorm4, line 58: Is this just for symmetry? I thought the `runtime` flag essentially covered what behavior to follow. Are we also introducing this platform flag?

src/model_config_utils.h
```diff
@@ -129,6 +144,15 @@ Status LocalizePythonBackendExecutionEnvironmentPath(
     const std::string& model_path, inference::ModelConfig* config,
     std::shared_ptr<LocalizedPath>* localized_model_dir);
 
+/// Set execution environments for Python backend based backend, if the
+/// execution environments is not already specified and the execution
+/// environment file '<backend_libdir>/pb_exec_<runtime>.tar.gz' exists.
+/// \param backend_libdir The backend runtime library directory path.
+/// \param model_config The model configuration.
+/// \return The error status.
+Status SetPythonBackendBasedBackendExecutionEnvironment(
```
rmccorm4, line 153: "python based" over "python backend based" everywhere applicable
```suggestion
Status SetPythonBasedBackendExecutionEnvironment(
```

src/model_config_utils.h
```diff
@@ -129,6 +144,15 @@ Status LocalizePythonBackendExecutionEnvironmentPath(
     const std::string& model_path, inference::ModelConfig* config,
     std::shared_ptr<LocalizedPath>* localized_model_dir);
 
+/// Set execution environments for Python backend based backend, if the
```
rmccorm4, line 147:
```suggestion
/// Set execution environments for Python based backend, if the
```

src/backend_model.cc
```diff
@@ -317,75 +288,140 @@ TritonModel::GetExecutionPolicy(const inference::ModelConfig& model_config)
...
+  for (const auto& path : search_paths) {
+    const auto full_path = JoinPath({path, backend_libname});
+    bool exists = false;
+    RETURN_IF_ERROR(FileExists(full_path, &exists));
+    if (exists) {
+      *backend_libdir = path;
+      *backend_libpath = full_path;
+      break;
+    }
+  }
```
rmccorm4, line 424: I think we may need to introduce some canonicalization/absolute pathing here to avoid arbitrary code execution - CC @nnshah1 .

Assume there is some directory mounted at `/mnt` for whatever reason.

One thing to do a quick check, can you provide something like this:
```
# config.pbtxt

# try to escape default search path `/opt/tritonserver/backends`
runtime: "../../../mnt/my_malicious_lib.so"
```

We probably shouldn't allow escaping the expected search directories, regardless of how likely it is in practice.

src/constants.h
kthui: This is a good point. After autocomplete, the [model config is validated](https://github.com/triton-inference-server/core/pull/285/files#diff-f947d03afcf01249667bedc5b955404da083564c22809b6666ba923356f7c829L1310), which [takes the platform value and compare it againsts the backend value](https://github.com/triton-inference-server/core/pull/285/files#diff-f947d03afcf01249667bedc5b955404da083564c22809b6666ba923356f7c829R1425). If the runtime is Python, it would be strange to use the LibTorch platform, but [keeping it empty will fail the check](https://github.com/triton-inference-server/core/pull/285/files#diff-f947d03afcf01249667bedc5b955404da083564c22809b6666ba923356f7c829L2263), so it is there to keep the validation logic happy.

rmccorm4: Does this mean if we wanted to add tensorflow python runtime implementation, we'd also need to introduce `platform: "tensorflow_python"` for it to work? Is there a clean way to not require the additional platform flag for these?

I think it also checks for backend field right?

I believe expectation is that users will specify:
```
backend: "pytorch"
```
and there is backwards compatibility to support:
```
platform: "pytorch_libtorch" # --> backend: pytorch
```

If both are left unspecified, we would check file extension somewhere like `.pt` or something to infer pytorch backend. Then if all else fails, we cannot infer the backend and they need to specify it in the config.

I don't think we need to add `platform: pytorch_python` to satisfy this flow, do we?

src/backend_model.cc
rmccorm4: I agree it would probably cover most cases. Something like this may still be possible, not sure:
```
runtime: ".."
```

`<backend_dir>/../libtriton_...so` :sweat_smile: 

But then you could argue that you just add `..` to your excluded substrings.

---

I tried to just check for special characters like this for the model load CVE initially, but I got push back to just fully canonicalize the path we're using and assert that path lives within the expected search paths. This would be a bit more futureproof to account for some unexpected case of symlinks or something like that.

I'll let @nnshah1 share his thoughts.

src/model_config_utils.h
kthui: Updated: [Shorten function name](https://github.com/triton-inference-server/core/pull/285/commits/f22a84c452a7cf6eac8d103a393b890e8e62ac60)

src/model_config_utils.h
kthui: Updated: [Shorten function name](https://github.com/triton-inference-server/core/pull/285/commits/f22a84c452a7cf6eac8d103a393b890e8e62ac60)

src/backend_model.cc
kthui: Add `..` to be safe: [Add .. to runtime library name check](https://github.com/triton-inference-server/core/pull/285/commits/1c2c618df891abbb480eff814e7e57369a589c10)

nnshah1: @kthui , @rmccorm4 - We should follow the same logic / utilities as in the model load api. 

I'm not in favor of searching for specific strings as that can be error prone and os specific.

Better to canonicalize the backend directory and the final path and ensure it is within the x/backend directory.

This also allows for relative / absolute paths as well if the user wants - but ensures that the final path is correct.

src/model_config_utils.cc
```diff
@@ -933,6 +933,26 @@ LocalizePythonBackendExecutionEnvironmentPath(
...
+    std::string env_name = "pb_exec_env_" + model_config->runtime() + ".tar.gz";
+    std::string env_path = JoinPath({backend_libdir, std::move(env_name)});
+    bool env_path_exist;
+    RETURN_IF_ERROR(FileExists(env_path, &env_path_exist));
+    if (env_path_exist) {
+      inference::ModelParameter model_param;
+      model_param.set_string_value(env_path);
+      (*model_config->mutable_parameters())["EXECUTION_ENV_PATH"] =
+          std::move(model_param);
+      LOG_ERROR << "\n\n" << env_path << "\n";
```
nnshah1, line 950: Left over debug message?

src/backend_model.cc
```diff
@@ -317,75 +288,142 @@ TritonModel::GetExecutionPolicy(const inference::ModelConfig& model_config)
...
 {
-  for (const auto& path : search_paths) {
-    const auto full_path = JoinPath({path, backend_libname});
-    bool exists = false;
-    RETURN_IF_ERROR(FileExists(full_path, &exists));
-    if (exists) {
-      *backend_libdir = path;
-      *backend_libpath = full_path;
-      break;
+  std::string python_backend_based_backend_libdir;
```
nnshah1, line 311:
```suggestion
  std::string python_based_backend_libdir;
```

src/constants.h
```diff
@@ -55,11 +55,14 @@ constexpr char kOpenVINORuntimeBackend[] = "openvino";
...
 constexpr char kPyTorchBackend[] = "pytorch";
 
 constexpr char kPythonFilename[] = "model.py";
 constexpr char kPythonBackend[] = "python";
 
+constexpr char kVLLMBackend[] = "vllm";
```
nnshah1, line 64: Do we need these for any new python based backend? Would be good to avoid that if possible - we shouldn't need to update the constants if someone adds a backend in the future

src/constants.h
kthui: I removed the "pytorch_python" platform, and added a comment above saying new runtimes should ignore this field, which defaults for "pytorch_libtorch", as the platform may not be LibTorch.

src/constants.h
kthui: We do not need them. `constexpr char kVLLMBackend[] = "vllm";` is removed, and they are handled as custom backend.

src/backend_model.cc
```diff
@@ -317,12 +288,140 @@ TritonModel::GetExecutionPolicy(const inference::ModelConfig& model_config)
...
+        Status::Code::INVALID_ARG,
+        "backend library name '" + cpp_backend_libname +
+            "' escapes backend directory '" + *backend_libdir +
+            "', for model '" + model_config->name() +
+            "', check model config runtime field");
+  }
+
+  // Both 'backend_libdir' and 'backend_libpath' are now pointing to the C++
+  // backend library, 'backend_libdir' needs adjustment for Python backend based
+  // backend.
```
rmccorm4, line 362:
```suggestion
  // backend library, 'backend_libdir' needs adjustment for Python based backend.
```

rmccorm4 (approved):
Very clean implementation 🚀

rmccorm4 (approved):

PR #287, opened 2023-11-04
Resolve thread contention when waiting for instance slot

tanmayv25:
Before this change the mu_ may be held by the batcher thread while waiting for instance consumer to be available. This would cause the handler thread calling Enqueue on the model to block as well cascading the response back to the user. 

Follow-up item is to enhance the stress tests to these special cases.

### DIFF
--- a/src/dynamic_batch_scheduler.cc
+++ b/src/dynamic_batch_scheduler.cc
@@ -365,9 +365,9 @@ DynamicBatchScheduler::BatcherThread(const int nice)
           // outer lock to allow Enqueue threads above to make progress.
           lock.unlock();
           // Use slot lock to wait for the slot availability.
-          std::unique_lock<std::mutex> slot_lock(slot_mu_);
+          std::mutex slot_mu;
+          std::unique_lock<std::mutex> slot_lock(slot_mu);
           cv_.wait(slot_lock, wait_for_slots);
-
           // Recapture the outer most lock to keep making progress.
           lock.lock();
         }
--- a/src/dynamic_batch_scheduler.cc
+++ b/src/dynamic_batch_scheduler.cc
@@ -358,7 +358,20 @@ DynamicBatchScheduler::BatcherThread(const int nice)
         if (payload_saturated_) {
           continue;
         }
-        cv_.wait(lock, wait_for_slots);
+
+        {
+          // The wait_for_slots conditional can be blocking till the slots
+          // are available for execution. Need to explicitly release the
+          // outer lock to allow Enqueue threads above to make progress.
+          lock.unlock();
+          // Use slot lock to wait for the slot availability.
+          std::unique_lock<std::mutex> slot_lock(slot_mu_);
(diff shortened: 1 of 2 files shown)


src/dynamic_batch_scheduler.cc
```diff
@@ -358,7 +358,20 @@ DynamicBatchScheduler::BatcherThread(const int nice)
...
-        cv_.wait(lock, wait_for_slots);
+
+        {
+          // The wait_for_slots conditional can be blocking till the slots
+          // are available for execution. Need to explicitly release the
+          // outer lock to allow Enqueue threads above to make progress.
+          lock.unlock();
+          // Use slot lock to wait for the slot availability.
+          std::unique_lock<std::mutex> slot_lock(slot_mu_);
+          cv_.wait(slot_lock, wait_for_slots);
```
rmccorm4, line 369: As discussed offline, because `wait_for_slots()` is blocking until payload slot is available, I think we should remove the lock/cv usage here, as it introduces some confusion. There is no other threads here trying to acquire slot_lock/slot_mu

This chunk should basically:
```cpp
unlock()
block_until_slot/consumer_available
lock()
```

tanmayv25: In our discussion today, we missed the case where the queue supports pre-fetching. The payload callback would need to notify the batcher thread in this case as soon as the slot is made available. I think it is best to keep a unified interface for both the variants.

rmccorm4: Makes sense. Do you think the comments here or some other kind of docs could better summarize the high level process involved here between the handler <-> batcher <-> backend threads for future readers that we spent a while talking about?

nnshah1: One potential here is a "retrospective design doc" that would describe an area in detail

tanmayv25: Created DLIS-5785 to write a architecture doc for the current implementation of dynamic batcher and the thread interactions there in.

kthui (dismissed):
If I understand correctly the only goal is release the lock and allow `enqueue` to make progress, then the following should also work?
```
{
  lock.unlock();
  
  std::mutex tmp_mu;
  std::unique_lock<std::mutex> tmp_lock(tmp_mu);
  cv_.wait(tmp_lock, wait_for_slots);
  
  lock.lock();
}
```

tanmayv25:
> If I understand correctly the only goal is release the lock and allow `enqueue` to make progress, then the following should also work?
> 
> ```
> {
>   lock.unlock();
>   
>   std::mutex tmp_mu;
>   std::unique_lock<std::mutex> tmp_lock(tmp_mu);
>   cv_.wait(tmp_lock, wait_for_slots);
>   
>   lock.lock();
> }
> ```

Great suggestion! Done! https://github.com/triton-inference-server/core/pull/287/commits/e869f92befbcd3abafeeafd18e839236b460118c

rmccorm4 (approved):

PR #288, opened 2023-11-07
nnshah1 suggestion

nnshah1:
In our conversation I realized that the path we write to is not the one we cananocalize. While not a high priority - I think we want to normalize the file path, check it, and then write to that same file path - just for consistency.

A future consideration: we can add checking to JoinPath(root, leaf) to:

1) return full absolute path
2) optional error if leaf exits the roof 

This would help in any place we use Join to put paths together that may escape their intended location.

### DIFF
--- a/src/model_repository_manager.cc
+++ b/src/model_repository_manager.cc
@@ -115,13 +115,11 @@ class LocalizeRepoAgent : public TritonRepoAgent {
               // Resolve any relative paths or symlinks, and enforce that target
               // directory stays within model directory for security.
               // DLIS-5149: Can use std::filesystem over boost in C++17.
-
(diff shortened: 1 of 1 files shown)


src/model_repository_manager.cc
```diff
@@ -130,6 +129,9 @@ class LocalizeRepoAgent : public TritonRepoAgent {
                         .c_str());
               }
 
+              // Save model override file to the instructed directory using the
+              // temporary model directory as the basepath.
+              const std::string dir = DirName(file_path);
```
rmccorm4, line 134: I think dir is undefined in the error above after moving it.

rmccorm4 (approved):

PR #294, opened 2023-11-18
Add more NVTX markers to core API

szalpal:
This PR proposed adding two new NVTX markers to core API. The point of adding them is that it will help profiling execution within Backends and Backend optimization.


dyastremsky (approved):

rmccorm4:
Seems fine to me. @szalpal are you able to run a pipeline to confirm this builds and runs OK?

PR #299, opened 2023-12-07
Python API Beta

nnshah1:
Triton Inference Server In-Process Python API [BETA]


python/CMakeLists.txt
```diff
@@ -66,4 +67,5 @@ install(
 # Test
 install(
   CODE "file(INSTALL ${CMAKE_CURRENT_BINARY_DIR}/test/test_binding.py DESTINATION \"${CMAKE_INSTALL_PREFIX}/python\")"
+  CODE "file(INSTALL ${CMAKE_CURRENT_BINARY_DIR}/test/test_api.py DESTINATION \"${CMAKE_INSTALL_PREFIX}/python\")"
```
mc-nv, line 70: My assumption that `install` method of `CMake` would be better to install files if it works: 
```CMake
install(
  FILES 
  ${CMAKE_CURRENT_BINARY_DIR}/test/test_api.py
  ${CMAKE_CURRENT_BINARY_DIR}/test/test_binding.py
  DESTINATION ${CMAKE_INSTALL_PREFIX}/python
  )
  ```

File: python/test/test_api.py [hunk 2]
```diff
@@ -0,0 +1,146 @@
+import asyncio
```
nv-hwoo, line 1: Missing copyright

python/tritonserver/_api/__init__.py
```diff
@@ -0,0 +1 @@
+
```
nv-hwoo, line 1: Missing copyright

python/tritonserver/_api/_datautils.py
```diff
@@ -0,0 +1,327 @@
+from __future__ import annotations
```
nv-hwoo, line 1: Missing copyright

python/CMakeLists.txt
```diff
@@ -58,12 +59,20 @@ add_custom_target(
     "${wheel_stamp_file}"
 )
 
+
+# Wheel
+file(GLOB _Wheel "${CMAKE_CURRENT_BINARY_DIR}/generic/triton*.whl")
```
Tabrizian, line 64: nit: WHEEL
```suggestion
file(GLOB WHEEL "${CMAKE_CURRENT_BINARY_DIR}/generic/triton*.whl")
```

python/CMakeLists.txt
```diff
@@ -30,7 +30,8 @@ add_subdirectory(tritonserver)
 file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/TRITON_VERSION ${TRITON_VERSION})
 configure_file(../LICENSE LICENSE.txt COPYONLY)
 configure_file(setup.py setup.py @ONLY)
-file(COPY test/test_binding.py DESTINATION ./test/.)
+file(GLOB _Tests test/*.py)
```
Tabrizian, line 33:
```suggestion
file(GLOB TESTS test/*.py)
```

python/build_wheel.py
```diff
@@ -93,11 +93,11 @@ def sed(pattern, replace, source, dest=None):
     print("=== Building in: {}".format(os.getcwd()))
     print("=== Using builddir: {}".format(FLAGS.whl_dir))
     print("Adding package files")
-
     mkdir(os.path.join(FLAGS.whl_dir, "tritonserver"))
     shutil.copy("tritonserver/__init__.py", os.path.join(FLAGS.whl_dir, "tritonserver"))
-
+    shutil.copy("tritonserver/py.typed", os.path.join(FLAGS.whl_dir, "tritonserver"))
```
Tabrizian, line 98: Would be good to add a comment about what this file does.

python/test/test_api.py
```diff
@@ -0,0 +1,146 @@
...
+        count = 0
+        while count < 2:
+            response = response_queue.get()
+            count += 1
+            print(response, count)
+            print(response.outputs["text_output"])
+            print(bytes(response.outputs["text_output"][0]))
+            print(type(response.outputs["text_output"][0]))
+            print(response.outputs["fp16_output"])
+            print(type(response.outputs["fp16_output"][0]))
```
Tabrizian, line 84: Need to cleanup the tests

python/tritonserver/__init__.py
```diff
@@ -23,3 +23,44 @@
 # OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""In process Python API for Triton Inference Server."""
+
+import tritonserver._c as triton_bindings
```
Tabrizian, line 29:
```suggestion
import tritonserver._c as _triton_bindings
```

python/tritonserver/__init__.py
```diff
@@ -23,3 +23,44 @@
...
+    ModelBatchFlag,
+    ModelControlMode,
+    ModelIndexFlag,
+    ModelTxnPropertyFlag,
+    Options,
+    RateLimitMode,
+    Server,
+)
+
+exceptions = [
```
Tabrizian, line 47:
```suggestion
_exceptions = [
```

python/tritonserver/__init__.py
```diff
@@ -23,3 +23,44 @@
...
+    triton_bindings.InvalidArgumentError,
+    triton_bindings.UnavailableError,
+    triton_bindings.UnsupportedError,
+    triton_bindings.AlreadyExistsError,
+]
+
+
+# Rename module for exceptions to simplify stack trace
+for exception in exceptions:
+    exception.__module__ = "tritonserver"
```
Tabrizian, line 61: question: what does the exception look like without this change?

python/tritonserver/_api/_datautils.py
```diff
@@ -0,0 +1,327 @@
...
+import tritonserver._api._dlpack as _dlpack
+from tritonserver import _c as triton_bindings
+
+DLPACK_DEVICE_TYPE_TO_TRITON_MEMORY_TYPE = defaultdict(
+    lambda: None,
+    {
+        _dlpack.DLDeviceType.kDLCUDA: triton_bindings.TRITONSERVER_MemoryType.GPU,
+        _dlpack.DLDeviceType.kDLCPU: triton_bindings.TRITONSERVER_MemoryType.CPU,
+    },
+)
```
Tabrizian, line 20:
```suggestion
DLPACK_DEVICE_TYPE_TO_TRITON_MEMORY_TYPE = {
        _dlpack.DLDeviceType.kDLCUDA: triton_bindings.TRITONSERVER_MemoryType.GPU,
        _dlpack.DLDeviceType.kDLCPU: triton_bindings.TRITONSERVER_MemoryType.CPU,
    }
```

python/tritonserver/_api/_datautils.py
```diff
@@ -0,0 +1,327 @@
...
+        (
+            _dlpack.DLDataTypeCode.kDLFloat,
+            16,
+        ): triton_bindings.TRITONSERVER_DataType.FP16,
+    },
+)
+
+
+NUMPY_TO_TRITON_DTYPE = defaultdict(
+    lambda: triton_bindings.TRITONSERVER_DataType.INVALID,
```
Tabrizian, line 35: question why we are using defaultdict instead of a regular dictionary?

python/tritonserver/_api/wrapper.py
```diff
@@ -0,0 +1,725 @@
...
+        return self._server.is_live()
+
+    def is_ready(self):
+        return self._server.is_ready()
+
+    def get_model(self, model_name, model_version=-1):
+        return Model(self._server, model_name, model_version)
+
+    @property
+    def models(self):
```
Tabrizian, line 376: is there a way to access a specific model version in this API or the user has to use `get_model`?

python/test/test_api.py
```diff
@@ -0,0 +1,146 @@
...
+
+        for response in responses:
+            print(response)
+
+        responses = server.models["test"].infer(
+            inputs={"fp16_input": fp16_input}, request_id="1"
+        )
+
+        for response in responses:
+            print(response)
```
Tabrizian, line 39: what if the response is on GPU? Is it still going to be returned in numpy?

python/tritonserver/_api/_datautils.py
```diff
@@ -0,0 +1,327 @@
...
+
+        if hasattr(self, "set_attributes"):
+            self._allocator.set_buffer_attributes_function(self.set_buffer_attributes)
+        return self._allocator
+
+
+try:
+    import cupy
+
+    class CupyAllocator(MemoryAllocator):
```
Tabrizian, line 106: question: where is this Allocator class used?

python/tritonserver/_api/_datautils.py
```diff
@@ -0,0 +1,327 @@
...
+    def data_type(self):
+        return (self._tensor.dtype.type_code, self._tensor.dtype.bits)
+
+    @property
+    def data_ptr(self):
+        return self._tensor.data + self._tensor.byte_offset
+
+
+@dataclass
+class MemoryBuffer:
```
Tabrizian, line 221: perhaps rename to Tensor?

python/tritonserver/_api/wrapper.py
```diff
@@ -0,0 +1,725 @@
...
+        options.set_server_id(self.server_id)
+
+        if not isinstance(self.model_repository, list):
+            self.model_repository = [self.model_repository]
+        for model_repository_path in self.model_repository:
+            options.set_model_repository_path(model_repository_path)
+        options.set_model_control_mode(self.model_control_mode)
+
+        for startup_model in self.startup_models:
+            options.set_startup_model(startup_model)
```
Tabrizian, line 221: Does `set_startup_model` append to the startup models?

python/test/test_api.py
```diff
@@ -0,0 +1,146 @@
...
+        response_queue = asyncio.Queue()
+        responses = test.async_infer(
+            inputs=inputs, response_queue=response_queue, request_id="1"
+        )
+        responses_2 = test.async_infer(
+            inputs=inputs, response_queue=response_queue, request_id="2"
+        )
+        responses_3 = test.async_infer(
+            inputs=inputs, response_queue=response_queue, request_id="3"
+        )
```
Tabrizian, line 125: The API looks very clean! I was wondering why we support both response_queue and returning a list of responses. I feel like just returning a list of responses is sufficient.

python/tritonserver/_api/_datautils.py
nnshah1: wanted to avoid considering it a tensor specifically because I felt the term gets overloaded - but can revisit.

python/tritonserver/_api/_datautils.py
nnshah1: Used when the request is created and callbacks are set. Currently always set to default Numpy - but plan to add an option to choose Cupy. 

Question: currently there is no negotiation - meaning the client sets the memory type - would it be better to try to meet the backends request?

python/test/test_api.py
nnshah1: in GPU case it will be cupy (except for BYTES case which didn't have an easy way to represent that in cupy)

python/tritonserver/_api/wrapper.py
nnshah1: models[("test",1)] 

You can give either a string or a tuple.

I also considered models["test"][1] - but the mechanics became a little convoluted and I felt "string or tuple" was fine enough

python/test/test_api.py
nnshah1: This after considering the case where you would fire X requests in parallel and wanted to handle responses as they came.

Tabrizian (commented):
Some minor comments. I think apart from the tensor abstraction and the default allocator changes that we discussed offline it looks good to me.

We definitely need to add more testing but can be part of future PRs.

python/test/test_api.py
```diff
@@ -0,0 +1,238 @@
...
+import pytest
+import tritonserver
+
+try:
+    import cupy
+except Exception:
+    cupy = None
+
+
+class TrtionServerAPITest(unittest.TestCase):
```
rmccorm4, line 42:
```suggestion
class TritonServerAPITest(unittest.TestCase):
```

python/tritonserver/_api/_datautils.py
```diff
@@ -0,0 +1,400 @@
...
+
+    @staticmethod
+    def from_value(value):
+        if type(value) in MemoryBuffer._supported_conversions:
+            return MemoryBuffer._supported_conversions[type(value)](value)
+        elif hasattr(value, "__dlpack__"):
+            return MemoryBuffer._from_dlpack(value)
+        else:
+            raise triton_bindings.InvalidArgumentError(
+                f"Input type {type(value)} not supported. Must be one of {list(MemoryBufer._supported_conversions.keys())} or the type must support __dlpack__"
```
rmccorm4, line 323:
```suggestion
                f"Input type {type(value)} not supported. Must be one of {list(MemoryBuffer._supported_conversions.keys())} or the type must support __dlpack__"
```

python/test/test_api.py
```diff
@@ -0,0 +1,238 @@
...
+    def test_not_started(self):
+        server = tritonserver.Server()
+        with self.assertRaises(tritonserver.InvalidArgumentError):
+            server.is_ready()
+
+    @pytest.mark.skipif(cupy is None, reason="Skipping gpu memory, cpupy not installed")
```
tanmayv25, line 48:
```suggestion
    @pytest.mark.skipif(cupy is None, reason="Skipping gpu memory, cupy not installed")
```

python/test/test_api.py
```diff
@@ -0,0 +1,238 @@
...
+
+        responses = server.models["test"].infer(
+            inputs={"fp16_input": fp16_input}, request_id="1"
+        )
+
+        for response in responses:
+            print(response)
+        try:
+            pass
+        #            server.stop()
```
tanmayv25, line 73: Why is this commented? Did you forget to uncomment it?

python/test/test_api.py
```diff
@@ -0,0 +1,238 @@
...
+        server = tritonserver.Server()
+        with self.assertRaises(tritonserver.InvalidArgumentError):
+            server.is_ready()
+
+    @pytest.mark.skipif(cupy is None, reason="Skipping gpu memory, cpupy not installed")
+    def test_gpu_memory(self):
+        import cupy
+
+        server = tritonserver.Server(
+            model_repository="/workspace/models", exit_timeout=5
```
tanmayv25, line 53: Can we set these default values(like "/workspace/models") in a single place and use everywhere else?
May be setUp?

python/test/test_api.py
```diff
@@ -0,0 +1,238 @@
...
+        )
+        server.start(blocking=True)
+
+        model = server.models["test"]
+
+        responses = model.infer(
+            inputs={"fp16_input": numpy.array([[0.5]], dtype=numpy.float16)}
+        )
+
+        print(list(responses)[0])
```
tanmayv25, line 95: Add assert to all these calls to fail instead of printing here and below?

python/test/test_api.py
```diff
@@ -0,0 +1,238 @@
...
+        responses_3 = test_2.infer(inputs=inputs)
+
+        for response in responses_3:
+            print(response)
+
+        count = 0
+        while count < 2:
+            response = response_queue.get()
+            count += 1
+            print(response, count)
```
tanmayv25, line 163: I would advice checking for the `request_id` in the response from the queue.

python/test/test_api.py
```diff
@@ -0,0 +1,238 @@
...
+
+        test = server.get_model("test")
+        test_2 = server.get_model("test_2")
+
+        inputs = {
+            "text_input": numpy.array(["hello"], dtype=numpy.object_),
+            "fp16_input": numpy.array([["1"]], dtype=numpy.float16),
+        }
+
+        responses_1 = test.infer(
```
tanmayv25, line 147: Why does infer call returns an iterator even when  response_queue is specified? 
If the responses are directed to the response_queue, then infer should not return the iterator.
Is there a way to merge the individual iterators into a single combined iterator object instead of passing a response queue as an explicit option?

python/test/test_api.py
```diff
@@ -0,0 +1,238 @@
...
+import time
+import unittest
+
+import numpy
+import pytest
+import tritonserver
+
+try:
+    import cupy
+except Exception:
```
rmccorm4, line 38: `ImportError`

python/test/test_api.py
rmccorm4, line 48:
```suggestion
    @pytest.mark.skipif(cupy is None, reason="Skipping gpu memory, cupy not installed")
```

python/tritonserver/_api/wrapper.py
```diff
@@ -0,0 +1,782 @@
...
+
+        options.set_strict_model_config(self.strict_model_config)
+        options.set_rate_limiter_mode(self.rate_limiter_mode)
+
+        for rate_limiter_resource in self.rate_limiter_resources:
+            options.set_rate_limiter_resouces(
```
rmccorm4, line 228:
```suggestion
            options.set_rate_limiter_resources(
```

python/test/test_api.py
```diff
@@ -0,0 +1,238 @@
...
+            ]:
+                print(list(server.models.items()))
+                time.sleep(5)
+                continue
+            break
+        print(server.models[model.name])
+        print(list(server.models.items()))
+
+        print(model.is_ready())
+        server.stop()
```
rmccorm4, line 124: Will this be called implicitly when object goes out of scope?

python/test/test_api.py
GuanLuo: Check response v.s. expected response? `print` doesn't really test anything programmatically.

GuanLuo: Same for other `print` statements..

python/tritonserver/__init__.py
```diff
@@ -23,3 +23,51 @@
...
+from tritonserver._api.wrapper import RateLimitMode as RateLimitMode
+from tritonserver._api.wrapper import Server as Server
+from tritonserver._c import AlreadyExistsError as AlreadyExistsError
+from tritonserver._c import InternalError as InternalError
+from tritonserver._c import InvalidArgumentError as InvalidArgumentError
+from tritonserver._c import NotFoundError as NotFoundError
+from tritonserver._c import TritonError as TritonError
+from tritonserver._c import UnavailableError as UnavailableError
+from tritonserver._c import UnknownError as UnknownError
+from tritonserver._c import UnsupportedError as UnsupportedError
```
GuanLuo, line 52: What is the use of `as ...`?

python/tritonserver/_api/_datautils.py
GuanLuo: Should there be key error instead of returning "invalid" type value?

python/tritonserver/_api/_datautils.py
GuanLuo: Maybe better to move the "reference" allocators to dedicated files, user can choose from "allocators.CupyAllocator" etc.

RE negotiation: it would be cool to show a collective allocator where it contains an instance of CupyAllocator and DefaultAllocator, and it delegates the allocation to them based on backend's request.

python/tritonserver/__init__.py
nnshah1: I found a reference that this is preferred way of explicitly exporting for type checkers, might only be for stub files - but will double check. 

https://peps.python.org/pep-0484/#stub-files
https://github.com/microsoft/pylance-release/issues/2574

python/tritonserver/_api/_datautils.py
nnshah1: The defaultallocator actually has both Numpy and Cupy and delegates based on the backend request (if cupy is available). One question: is query required for negotiation or is to respond with the allocation type as requested sufficient?

python/tritonserver/_api/_datautils.py
nnshah1: Will rename to Tensor

python/tritonserver/_api/wrapper.py
```diff
@@ -0,0 +1,782 @@
...
+
+    def start(self, blocking=False, polling_interval=0.1):
+        if not isinstance(self._server, Server._UnstartedServer):
+            raise _triton_bindings.InvalidArgumentError("Server already started")
+
+        self._server = _triton_bindings.TRITONSERVER_Server(
+            self._options._create_server_options()
+        )
+        while blocking and not self.is_ready():
+            time.sleep(polling_interval)
```
GuanLuo, line 326: Should have some timeout setting..

python/tritonserver/_api/wrapper.py
```diff
@@ -0,0 +1,774 @@
...
+        else:
+            self._server.unload_model(model.name)
+        if blocking:
+            model_versions = [key for key in self.models.keys() if key[0] == model.name]
+            while [
+                key
+                for key in model_versions
+                if self.models[key].state not in Server.UNLOADED_STATES
+            ]:
+                time.sleep(polling_interval)
```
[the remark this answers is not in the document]
GuanLuo: I don't think that can be check easily..

python/tritonserver/_api/wrapper.py
```diff
@@ -0,0 +1,782 @@
...
+                model=self, _server=self._server, **kwargs
+            )
+        server_request, response_iterator = inference_request._create_server_request(
+            use_async_iterator=True
+        )
+        self._server.infer_async(server_request)
+        return response_iterator
+
+    def infer(
+        self, inference_request: InferenceRequest = None, **kwargs
```
GuanLuo, line 492: Instead of creating `InferenceRequest` internally by passing through `kwargs`, we may want to explicitly require an request object. Currently `kwargs` is tied to `InferenceRequest` and it is inflexible if we want to configure behavior of `infer` / `async_infer`, i.e. it will have to be consumed by `InferenceRequest` first.

nnshah1: we can add additional args before kwargs to expand / configure behavior - and pass rest to InferenceRequest  that is something like:

```suggestion
        self, inference_request: InferenceRequest = None, additional_arg:int , **kwargs
```

I'd like to keep the simplicity of not needing to create the object ahead if possible

python/tritonserver/__init__.py
```diff
@@ -23,3 +23,61 @@
...
+from tritonserver._api.wrapper import Server as Server
+from tritonserver._api._datautils import MemoryAllocator as MemoryAllocator
+from tritonserver._api._datautils import NumpyAllocator as NumpyAllocator
+from tritonserver._c import TRITONSERVER_DataType as DataType
+from tritonserver._c import TRITONSERVER_MemoryType as MemoryType
+
+try:
+    from tritonserver._api._datautils import CupyAllocator as CupyAllocator
+except ImportError:
+    pass
```
rmccorm4, line 53: I see a lot of this `try import cupy; except pass` throughout. I don't think the user will ever actually know if they're falling back to CPU when they could be using GPU had they just installed cupy. 

Can be log warnings somewhere?

python/tritonserver/_api/_datautils.py
```diff
@@ -0,0 +1,637 @@
...
+DLPACK_DEVICE_TYPE_TO_TRITON_MEMORY_TYPE = _CustomKeyErrorDict(
+    "DLPack device type",
+    "Triton server memory type",
+    {
+        _dlpack.DLDeviceType.kDLCUDA: MemoryType.GPU,
+        _dlpack.DLDeviceType.kDLCPU: MemoryType.CPU,
+    },
+)
+
+# TRITON_MEMORY_TYPE_TO_DLPACK_DEVICE_TYPE = (lambda: None, {})
```
rmccorm4, line 76: needed?

python/tritonserver/_api/_datautils.py
```diff
@@ -0,0 +1,637 @@
...
+        raise triton_bindings.UnsupportedError(f"{self._module_name} not supported")
+
+    def __setattribute__(self, name):
+        raise triton_bindings.UnsupportedError(f"{self._module_name} not supported")
+
+
+try:
+    import cupy
+except ImportError:
+    cupy = _UnsupportedModule("cupy")
```
rmccorm4, line 176: Why do we use this helper class here, but in other places do `cupy = None` ?

python/tritonserver/__init__.py
nnshah1: With the change:
`tritonserver.InternalError: failed to load all models`

Without the change:
`tritonserver._c.triton_bindings.InternalError: failed to load all models`

Small difference -but felt that since the errors are part of the API - that having them aliased made sense ... open to discuss though.

python/test/test_api.py
```diff
@@ -0,0 +1,237 @@
...
+
+    def test_unload(self):
+        server = tritonserver.Server(
+            model_repository="/workspace/models",
+            exit_timeout=5,
+            model_control_mode=tritonserver.ModelControlMode.EXPLICIT,
+            startup_models=["test"],
+            log_verbose=True,
+            log_error=True,
+            log_info=True,
```
rmccorm4, line 84: Are log error/warning/info not true by default?

python/tritonserver/__init__.py
```diff
@@ -23,3 +23,66 @@
...
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""In process Python API for Triton Inference Server."""
+
+import tritonserver._c as _triton_bindings
+from tritonserver._api._datautils import DataType as DataType
+from tritonserver._api._datautils import MemoryAllocator as MemoryAllocator
+from tritonserver._api._datautils import MemoryBuffer as MemoryBuffer
+from tritonserver._api._datautils import MemoryType as MemoryType
+from tritonserver._api._datautils import NumpyAllocator as NumpyAllocator
```
rmccorm4, line 34: The [TorchAllocator](https://github.com/triton-inference-server/tutorials/blob/cca6b1998d40e2e0539dc03adc6cb9ec0ccb67e4/RayServe/server.py#L36-L50) would make a great example for the docs around allocators if it's not going to be shipped by default

python/setup.py
```diff
@@ -61,7 +61,16 @@ def get_tag(self):
 data_files = [
     ("", ["LICENSE.txt"]),
 ]
-platform_package_data = [os.environ["TRITON_PYBIND"]]
+
+# Type checking marker file indicating support for type checkers.
```
rmccorm4, line 65: We should add a `__version__` to the package:
image

rmccorm4: Completely separately, if this is easy to do, we should also add it to the `tritonclient` package. There was an open github issue about it in the past.
image

python/test/test_api.py
nnshah1: I kept both the iterator and response queue to keep the API consistent for cancellation. Felt that was better than having the return type be different in this case between the two.

In terms of merging the iterators - it can be done but you'd have to check 'if async do this' on each response - as they need different handling of the queues (async requires to push to the event loop thread).

thought it would be cleanest to keep them seperated for now

python/test/test_api.py
nnshah1: yes

python/setup.py
nnshah1: @debermudez for context

python/tritonserver/_api/wrapper.py
```diff
@@ -0,0 +1,955 @@
...
+    request_id: Optional[str] = None
+    flags: int = 0
+    correlation_id: Optional[int | str] = None
+    priority: int = 0
+    timeout: int = 0
+    inputs: dict[str, _datautils.Tensor | Any] = field(default_factory=dict)
+    parameters: dict[str, str | int | bool] = field(default_factory=dict)
+    output_memory_type: Optional[_datautils.DeviceOrMemoryType] = None
+    output_memory_allocator: Optional[_datautils.MemoryAllocator] = None
+    output_array_module: Optional[ModuleType] = None
```
rmccorm4, line 886: Can you explain this array_module stuff being passed around? I saw it in the rayserve example and it feels like an antipattern.

If we need some kind of info at runtime to properly convert, can't we just check the object?
```python
if isinstance(array, np.ndarray):
  ...
elif isinstance(array, cupy.ndarray):
  ...
```

Is it only needed for dlpack logic?

rmccorm4: I also figured some of the awareness would be handled in the Allocators as well. Are these disjoint?

nnshah1: the design here is to allow the user to select the framework they want to use for output (which may be different or the same as input). They also have the option of just using the opaque Tensor object and using dlpack.

So its mostly a convenience - the idea is kind of borrowed from the python array api where you can access the array module from an object like you can get the __dl_pack__ to understand how to transfer the memory.

I think of it as, "output_memory_type" is to select the device

"output_array_module" is to select the framework to use the tensor.

in my head there are three distinct but related concepts:

1) device where memory is allocated
2) allocation framework
3) usage framework

these map to: output_memory_type, output_memory_allocator, and output_array_module

nnshah1: > I also figured some of the awareness would be handled in the Allocators as well. Are these disjoint?

Yes - disjoint. To keep things like RMM as a potential - memory is allocated independent of type and shape - just size and device.

rmccorm4: I feel it would be more intuitive for the user to explicitly control the usage framework portion you described, assuming we provide the necessary interface / conversion magic methods like `__dlpack__` to make it easy for them:
```python
    responses = model.infer(
        inputs={
            "prompt": np.array(
                ["boy with balloon, gritty, urban, charcoal sketch"], dtype="object"
            ).reshape(-1, 1)
        },
        output_memory_type="GPU",
        # output_array_module=torch
    )

    for response in responses:
        output_tensor = response.outputs["generated_image"]
        # generated_image = output_tensor.squeeze().type(torch.uint8)
        generated_image = torch.from_dlpack(output_tensor).squeeze().type(torch.uint8)
```

nnshah1: that's exactly how it works! if you don't specify an output_array_module - you get a Tensor (see example #1)

nnshah1: I added 'output array module' for completeness in my head -but we can easily remove it and force user to use the tensor object. Just thought it would be nice to have "torch in" "torch out" usability, as opposed to only have "torch in", "tensor out" (but that is the default)

rmccorm4: Right - I just don't think we should provide another option with output_array_module that passes a literal module around. I think that's an unnecessary extra step for the first iteration. I think our job is just to provide the necessary interfaces to make the Tensor object easy to use/interop - but not to handle the one-liner conversion call for them with an extra module arg. I think it hurts readability/intuition more than helps.

rmccorm4: The fact that we can call `torch.from_dlpack(Tensor)` or `Tensor.to_ndarray()` etc. on their own looks great to me and I think that's sufficient.

nnshah1: sounds good - I think __dlpack__ and to_ndarray are sufficient. 

to_ndarray will still be needed because of BYTES serialization / deserialization - but fallback is __dlpack__ - can remove the output_array_module

python/test/test_api.py
Tabrizian: I'm leaning toward just removing the "queue" functionality for beta and see whether there are any requests for it. If I remember correctly, from our previous conversation it sounded like the use-case was niche (i.e., to provide an ability for processing multiple responses in parallel).

python/tritonserver/_api/wrapper.py
```diff
@@ -0,0 +1,953 @@
...
+        return Model(self._server, model_name, model_version)
+
+    def models(self, exclude_not_ready: bool = False) -> ModelDictionary:
+        return ModelDictionary(self._server, self._model_index(exclude_not_ready))
+
+    def get_models(self, exclude_not_ready: bool = False) -> ModelDictionary:
+        return ModelDictionary(self._server, self._model_index(exclude_not_ready))
+
+    def model_index(self, exclude_not_ready: bool = False) -> ModelDictionary:
+        return ModelDictionary(self._server, self._model_index(exclude_not_ready))
```
Tabrizian, line 431: Let's only have a single method for each functionality (i.e., either use `live` or `is_live` but not both.

python/tritonserver/_api/_datautils.py
```diff
@@ -0,0 +1,753 @@
...
+
+        return MemoryBuffer(
+            int(dlpack_object.data_ptr),
+            dlpack_object.memory_type,
+            dlpack_object.memory_type_id,
+            dlpack_object.byte_size,
+            owner,
+        )
+
+    def _create_TRITONSERVER_BufferAttributes(self) -> BufferAttributes:
```
Tabrizian, line 238: small nit: `_create_tritonserver_bufferattributes` for consistent style.

python/tritonserver/_api/_datautils.py
```diff
@@ -0,0 +1,753 @@
...
+        buffer_attributes.memory_type_id = self.memory_type_id
+        buffer_attributes.byte_size = self.size
+        #        buffer_attributes.cuda_ipc_handle = None
+        return buffer_attributes
+
+
+class MemoryAllocator(ABC):
+    @abstractmethod
+    def allocate(
+        self, size: int, memory_type: MemoryType, memory_type_id: int, tensor_name: str
```
Tabrizian, line 250: I think name is probably not required for the `allocate` function.

python/tritonserver/_api/_datautils.py
```diff
@@ -0,0 +1,753 @@
...
+        return MemoryBuffer.from_dlpack(ndarray)
+
+
+default_memory_allocators: dict[MemoryType, MemoryAllocator] = dict(
+    {
+        MemoryType.CPU: NumpyAllocator(),
+    }
+)
+
+if cupy is not None:
```
Tabrizian, line 295: I am still not fully onboard with having cupy as one of the dependencies for Python APIs. I think using `cuda-python` probably makes more sense since we're just trying to allocate memory.

python/tritonserver/_api/_datautils.py
```diff
@@ -0,0 +1,753 @@
...
+
+    @property
+    def size(self) -> int:
+        return self.memory_buffer.size
+
+    def __dlpack__(self, *, stream=None):
+        #        if not (stream is None or (isinstance(stream, int) and stream == 0)):
+        #           raise UnsupportedError(
+        #              f"DLPack stream synchronization on {stream} not currently supported"
+        #         )
```
Tabrizian, line 541: Let's remove the comment.

python/tritonserver/_api/_datautils.py
```diff
@@ -0,0 +1,753 @@
...
+        numpy_ndarray = self._to_numpy_on_host()
+
+        # Deserialize bytes array and reshape
+        self.shape = original_shape
+        self.data_type = original_data_type
+        return Tensor._deserialize_bytes_array(numpy_ndarray).reshape(self.shape)
+
+    @staticmethod
+    def from_memory_buffer(data_type, shape, memory_buffer):
+        return Tensor(data_type, shape, memory_buffer)
```
Tabrizian, line 593: It looks identical to the constructor.

python/tritonserver/_api/_datautils.py
```diff
@@ -0,0 +1,753 @@
...
+    @staticmethod
+    def from_object(obj: Any):
+        if type(obj) in Tensor._from_converters:
+            return Tensor._from_converters[type(obj)](obj)
+        elif hasattr(obj, "__dlpack__"):
+            return Tensor.from_dlpack(obj)
+        else:
+            raise InvalidArgumentError(
+                f"Input type {type(obj)} not supported. Must be one of {list(Tensor._from_converters.keys())} or the type must support __dlpack__"
+            )
```
Tabrizian, line 604: I still think just having `from_dlpack` is sufficient... happy to discuss more offline.

Tabrizian (commented):
Thank you Neelay for your efforts on this. I think this is looking great.

A couple of items that I think it would be good to address:

* I still think that just having `from_dlpack` is sufficient for the Tensor interface.
* I think it would be good to make sure all the constructor of all the main wrapper objects can work with using only high-level wrapper objects (i.e., you could create an `InferenceRequest` using the constructor without having to use the lower level bindings).
* Cleaning up the tests. I noticed that there are some print statements and commented code. It would be good to clean that up before merging.
* I see that we're using cupy just for allocating memory. I think `cuda-python` is probably a better candidate if we're only interested in the memory allocation.

python/tritonserver/_api/_tensor.py
```diff
@@ -0,0 +1,594 @@
...
+    size : int
+        Size of the allocated memory in bytes.
+    owner : Any
+        Object that owns or manages the memory buffer.  Allocated
+        memory must not be freed while a reference to the owner is
+        held.
+
+    Examples
+    --------
+    >>> buffer = MemoryBuffer.from_dlpack(numpy.array([100],dtype=numpy.uint8))
```
rmccorm4: The examples look great!

Other examples don't have the `>>>`, so removing for consistency
```suggestion
    buffer = MemoryBuffer.from_dlpack(numpy.array([100],dtype=numpy.uint8))
```

python/tritonserver/_api/_tensor.py
```diff
@@ -0,0 +1,594 @@
...
+                     memory_type_id,
+                     tensor_name):
+
+            device = "cpu"
+
+            if memory_type == tritonserver.MemoryType.GPU:
+                device = "cuda"
+
+            tensor = torch.zeros(size,dtype=torch.uint8,device=device)
+            print("torch allocator!")
```
rmccorm4:
```suggestion
            tensor = torch.zeros(size,dtype=torch.uint8,device=device)
```

python/tritonserver/_api/_response.py
```diff
@@ -0,0 +1,386 @@
...
+
+    Response iterators are returned from model inference methods and
+    allow users to process inference responses in the order they were
+    received for a request.
+
+    """
+
+    def __init__(
+        self,
+        _model: Model,
```
Tabrizian, line 64: I think constructor/method arguments shouldn't start with underscore.

nnshah1: mainly wanted to indicate these were private - but will switch back, I think the docstring + import should be enough. Mainly the object is intended to be gotten from a factory method and not directly instantiated by user.

python/tritonserver/_api/_tensor.py
```diff
@@ -0,0 +1,499 @@
...
+        ----------
+        device : DeviceOrMemoryType
+            The target device. Device can be specified as a string,
+            MemoryType, tuple [MemoryType, memory_type__id], or
+            tuple[DLDeviceType, device_id].
+
+        Returns
+        -------
+        Tensor
+            The tensor moved to the specified device.
```
rmccorm4, line 366: Can we add small example for cpu and gpu?
```
       Examples
        --------
        ... to_device("cpu")
        ... to_device("gpu")
        ... to_device("gpu:1") # ? or whatever this syntax is if supported
        """
```

python/tritonserver/_api/_tensor.py
```diff
@@ -0,0 +1,499 @@
...
+            f"Conversion from {self.memory_type} to numpy array not supported."
+        )
+
+    @staticmethod
+    def _deserialize_bytes_array(numpy_ndarray: numpy.ndarray) -> numpy.ndarray:
+        result = []
+        _buffer = memoryview(numpy_ndarray)
+        offset = 0
+        while offset < len(_buffer):
+            (item_length,) = struct.unpack_from("@I", _buffer, offset)
```
rmccorm4, line 406: note: endianness if this doesn't match with our decisions around client/server endian spots

python/tritonserver/_api/_tensor.py
```diff
@@ -0,0 +1,499 @@
...
+        """
+        dlpack_object = DLPackObject(obj)
+        data_type = dlpack_object.triton_data_type
+        shape = dlpack_object.shape
+        memory_buffer = MemoryBuffer._from_dlpack_object(
+            obj, dlpack_object=dlpack_object
+        )
+        return Tensor(data_type, shape, memory_buffer)
+
+    def to_host(self) -> Tensor:
```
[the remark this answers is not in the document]
rmccorm4: Sounds like we're OK with to_host

python/tritonserver/_api/_tensor.py
```diff
@@ -0,0 +1,499 @@
...
+        Returns
+        -------
+        Any
+            A DLPack-compatible object representing the tensor.
+        """
+
+        #        if not (stream is None or (isinstance(stream, int) and stream == 0)):
+        #           raise UnsupportedError(
+        #              f"DLPack stream synchronization on {stream} not currently supported"
+        #         )
```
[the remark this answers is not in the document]
nnshah1: @oandreeva-nv - do you have a recommendation on how to handle this generically?

python/tritonserver/_api/_response.py
```diff
@@ -0,0 +1,456 @@
...
+                tensor = Tensor(data_type, shape, memory_buffer)
+
+                outputs[name] = tensor
+            result.outputs = outputs
+        except Exception as e:
+            error = InternalError(f"Unexpected error in creating response object: {e}")
+            error.args += tuple(asdict(result).items())
+            result.error = error
+
+        # values["classification_label"] = response.output_classification_label()
```
rmccorm4, line 454: Should we add some `# TODO: Handle classification` or something?

python/tritonserver/_api/_server.py
```diff
@@ -0,0 +1,1086 @@
...
+            Metric family that includes MetricKind, name, and
+            description.
+        labels : Optional[dict[str, str]]
+            Additional labels for the metric (returned in reporting to
+            distinguish multiple metrics within a family)
+
+        Examples
+        --------
+
+        Todo
```
rmccorm4, line 1075: Todo

nnshah1: will do

python/tritonserver/_api/_datautils.py
nnshah1: @rmccorm4 - what do you think - I'm on the fence here - I haven't seen a good case for knowing the tensor name in allocate (maybe for tracking?) - but had kept it since it comes from the underlying response allocator request. We could also remove it and add it back if needed ...

python/tritonserver/_api/_tensor.py
rmccorm4: On the topic of doctest or pydantic/pytest-examples mentioned, I don't know if these `>>>` will impact the "running" of the examples as well

python/tritonserver/_api/_allocators.py
```diff
@@ -0,0 +1,323 @@
...
+        _buffer_,
+        _buffer_user_object,
+        _byte_size,
+        _memory_type,
+        _memory_type_id,
+    ):
+        pass
+
+    def start(self, _allocator, _user_object):
+        pass
```
rmccorm4, line 284: what does it mean to "start" an allocator?

python/tritonserver/_api/_datautils.py
rmccorm4: I would remove the name for now to keep it simple, they appear unused. Agreed we can add it later if needed.

python/tritonserver/_api/_allocators.py
```diff
@@ -0,0 +1,323 @@
...
+                device = "cuda"
+
+            tensor = torch.zeros(size,dtype=torch.uint8,device=device)
+            return tritonserver.MemoryBuffer.from_dlpack(tensor)
+
+    """
+
+    @abstractmethod
+    def allocate(
+        self, size: int, memory_type: MemoryType, memory_type_id: int, tensor_name: str
```
rmccorm4, line 163: Attaching comments on removing `tensor_name` here since code has changed: https://github.com/triton-inference-server/core/pull/299#discussion_r1446484855

python/tritonserver/_api/_allocators.py
nnshah1: this is part of the c api - it is to indicate a start of allocation for a request - decided not to expose to public allocator interface but can add it in the future.

python/tritonserver/_api/_tensor.py
nnshah1: will normalize examples after merge if ok

python/tritonserver/_api/_datautils.py
nnshah1: added todo - and discussing with @oandreeva-nv

python/test/test_api.py
nnshah1: will do some testing of use case to resolve if needed. - post merge.

python/tritonserver/_api/_tensor.py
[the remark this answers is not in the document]
oandreeva-nv: discussed steps offline

python/tritonserver/_api/_dlpack.py
```diff
@@ -0,0 +1,272 @@
+# Copyright 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
Tabrizian, line 1: 2024?

python/tritonserver/_api/_dlpack.py
```diff
@@ -0,0 +1,272 @@
...
+            _raise_error(
+                "DLPack expects '__dlpack_device__' if '__dlpack__' has been defined"
+            )
+        device = dlpack_obj.__dlpack_device__()
+        # Have to condition on the device type as, using numpy as example,
+        # some DLPack implementation doesn't accept 'stream' as arguments
+        if device != DLDeviceType.kDLCUDA:
+            return dlpack_obj.__dlpack__()
+        else:
+            return dlpack_obj.__dlpack__(stream)
```
Tabrizian, line 258: I think here we may need to synchronize the default stream here. You've probably discussed this with @oandreeva-nv offline. This is an important issue that needs to be fixed before merging the PR.

python/tritonserver/_api/_model.py
```diff
@@ -0,0 +1,491 @@
...
+
+        """
+        if "model" in kwargs:
+            kwargs.pop("model")
+        return InferenceRequest(model=self, **kwargs)
+
+    def async_infer(
+        self,
+        inference_request: Optional[InferenceRequest] = None,
+        raise_on_error: bool = False,
```
Tabrizian, line 144: I think probably the pythonic way is to always raise an exception on error.

cc @rmccorm4 in case you have any thoughts here.

python/test/test_api.py
```diff
@@ -0,0 +1,123 @@
...
+        pass
+
+
+class TensorTests(unittest.TestCase):
+    def test_cpu_to_gpu(self):
+        cpu_array = numpy.random.rand(1, 3, 100, 100).astype(numpy.float32)
+        cpu_tensor = tritonserver.Tensor.from_dlpack(cpu_array)
+        gpu_tensor = cpu_tensor.to_device("gpu")
+        cupy.from_dlpack(gpu_tensor)
+        torch.from_dlpack(gpu_tensor)
```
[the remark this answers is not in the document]
nnshah1: tests are WIP - please ignore for now

python/tritonserver/_api/_model.py
nnshah1: how about switch default? I think its useful in case of multiple responses where some may have error and some may not - "raise_on_error" is relatively well used (open telemetry, elasticsearch) for different scenarios

Tabrizian: sounds good.

rmccorm4: LGTM

python/test/test_api_models/test/1/model.py
```diff
@@ -0,0 +1,75 @@
...
+            )
+            outputs.append({"name": output_name, "data_type": data_type, "dims": dims})
+
+        for input_ in inputs:
+            auto_complete_model_config.add_input(input_)
+        for output in outputs:
+            auto_complete_model_config.add_output(output)
+
+        auto_complete_model_config.set_max_batch_size(0)
+        if config["platform"] == "decoupled":
```
rmccorm4, line 36: I think it's fine for testing for now - but I think this use of the `platform` field for convenience ("decoupled", "test", etc) may confuse users that look to the tests for examples. Typically platform field is a backend-specific flag with a meaning for that backend. That being said, I thought it was an interesting use to see 🤓

python/test/test_api_models/test/config.pbtxt
```diff
@@ -0,0 +1,28 @@
+name: "test"
```
rmccorm4, line 1: Removing use of `name` field in config will simplify re-use/renaming of this model/config for future tests. When specified, it must match the folder name. When unspecified, it just uses the folder name. So I find it pretty redundant and cumbersome for testing.

python/test/test_api_models/test/config.pbtxt
```diff
@@ -0,0 +1,28 @@
...
+  name: "text_output"
+  data_type: TYPE_STRING
+  dims: [ -1 ]
+},
+{
+  name: "fp16_output",
+  data_type: TYPE_FP16
+  dims: [-1,1]
+}
+]
```
rmccorm4, line 28: do you want these I/O defined here too? or rely on the autocomplete in model.py as single source of truth?

python/tritonserver/_api/_allocators.py
```diff
@@ -0,0 +1,319 @@
...
+            dlpack_object.byte_size,
+            owner,
+        )
+
+    def _create_tritonserver_buffer_attributes(self) -> TRITONSERVER_BufferAttributes:
+        buffer_attributes = TRITONSERVER_BufferAttributes()
+        buffer_attributes.memory_type = self.memory_type
+        buffer_attributes.memory_type_id = self.memory_type_id
+        buffer_attributes.byte_size = self.size
+        #        buffer_attributes.cuda_ipc_handle = None
```
rmccorm4, line 127: Is there a `# TODO` or follow-up needed for CUDA SHM?
```suggestion
        # buffer_attributes.cuda_ipc_handle = None
```

python/tritonserver/_api/_allocators.py
```diff
@@ -0,0 +1,319 @@
...
+            memory type id requested (typically device id)
+
+        Returns
+        -------
+        MemoryBuffer
+            memory buffer with requested size
+
+        Examples
+        --------
+        memory_buffer = allocator.allocate(100,MemoryType.CPU,0)
```
rmccorm4, line 185:
```suggestion
        memory_buffer = allocator.allocate(100, MemoryType.CPU, 0)
```

python/tritonserver/_api/_allocators.py
```diff
@@ -0,0 +1,319 @@
...
+        the requested memory type or memory type id
+
+        Parameters
+        ----------
+        size : int
+            number of bytes requested
+        memory_type : MemoryType
+                type of memory requested (CPU, GPU, etc.)
+        memory_type_id : int
+            memory type id requested (typically device id)
```
rmccorm4, line 176:
```suggestion
        size : int
            number of bytes requested
        memory_type : MemoryType
            type of memory requested (CPU, GPU, etc.)
        memory_type_id : int
            memory type id requested (typically device id)
```

python/tritonserver/_api/_allocators.py
```diff
@@ -0,0 +1,319 @@
...
+        buffer_attributes,
+        _user_object,
+        _buffer_user_object,
+    ):
+        if self._memory_type is not None:
+            buffer_attributes.memory_type = self._memory_type
+            buffer_attributes.memory_type_id = self._memory_type_id
+        return buffer_attributes
+
+    def create_TRITONSERVER_ResponseAllocator(self):
```
rmccorm4, line 310: lowercase `tritonserver` for consistency everywhere

python/tritonserver/_api/_model.py
```diff
@@ -0,0 +1,491 @@
...
+        InvalidArgumentError
+            if any invalid arguments are provided
+
+        Examples
+        --------
+        >>> responses = server.model("test_2").infer(inputs={"text_input":["hello"]})
+        responses = list(server.model("test_2").infer(inputs={"text_input":["hello"]}))
+
+        >>> response = responses[0]
+        print(respponse)
```
rmccorm4, line 259:
```suggestion
        print(response)
```

python/tritonserver/_api/_dlpack.py
```diff
@@ -0,0 +1,272 @@
...
+    pycapsule: ctypes.py_object = ctypes.cast(handle, ctypes.py_object)
+    if ctypes.pythonapi.PyCapsule_IsValid(pycapsule, c_str_dltensor):
+        dl_managed_tensor = ctypes.pythonapi.PyCapsule_GetPointer(
+            pycapsule, c_str_dltensor
+        )
+        managed_tensor_deleter(dl_managed_tensor)
+        ctypes.pythonapi.PyCapsule_SetDestructor(pycapsule, None)
+
+
+def triton_to_dlpack_dtype(dtype):
```
rmccorm4, line 170: This function could probably re-use `TRITON_TO_DLPACK_DTYPE` or `DLPACK_TO_TRITON_DTYPE` in `_datautils.py` for the mappings, or share a source of truth

python/tritonserver/_api/_dlpack.py
```diff
@@ -0,0 +1,272 @@
...
+        type_code = DLDataTypeCode.kDLFloat
+        bits = 64
+    elif dtype == "BF16":
+        type_code = DLDataTypeCode.kDLBfloat
+        bits = 16
+    elif dtype == "BYTES":
+        _raise_error("DLPack currently doesn't suppose BYTES type")
+    else:
+        _raise_error(
+            "Can not covert unknown data type '{}' to DLPack data type".format(dtype)
```
rmccorm4, line 214:
```suggestion
            "Can not convert unknown data type '{}' to DLPack data type".format(dtype)
```

python/tritonserver/_api/_dlpack.py
```diff
@@ -0,0 +1,272 @@
...
+        type_code = DLDataTypeCode.kDLFloat
+        bits = 32
+    elif dtype == "FP64":
+        type_code = DLDataTypeCode.kDLFloat
+        bits = 64
+    elif dtype == "BF16":
+        type_code = DLDataTypeCode.kDLBfloat
+        bits = 16
+    elif dtype == "BYTES":
+        _raise_error("DLPack currently doesn't suppose BYTES type")
```
rmccorm4, line 211:
```suggestion
        _raise_error("DLPack currently doesn't support BYTES type")
```

python/tritonserver/_api/_model.py
```diff
@@ -0,0 +1,491 @@
...
+    def async_infer(
+        self,
+        inference_request: Optional[InferenceRequest] = None,
+        raise_on_error: bool = True,
+        **kwargs: Unpack[InferenceRequest],
+    ) -> AsyncResponseIterator:
+        """Send an inference request to the model for execution
+
+        Sends an inference request to the model. Responses are
+        returned using an async.io compatible iterator. See
```
rmccorm4, line 150:
```suggestion
        returned using an asyncio compatible iterator. See
```

python/tritonserver/_api/_request.py
```diff
@@ -0,0 +1,196 @@
...
+        Memory allocator to use for inference response output. If not
+        provided default allocators will be used to allocate
+        MemoryType.GPU or MemoryType.CPU memory as set in
+        output_memory_type or as requested by backend / model
+        preference.
+    response_queue : Optional[Union[queue.SimpleQueue, asyncio.Queue]], default None
+        Queue for asynchronous handling of inference responses. If
+        provided Inference responses will be added to the queue in
+        addition to the response iterator. Must be queue.SimpleQueue
+        for non async.io requests and asyncio.Queue for asyncio
```
rmccorm4, line 98:
```suggestion
        for non asyncio requests and asyncio.Queue for asyncio
```

python/tritonserver/_api/_server.py
rmccorm4: Just realized there's about ~8 of these "Todo" examples - just mentioning here rather than commenting on each one. I'd be fine with removing the block and adding the example in a follow-up, but I don't think the rendered doc should have the "Todo"s in the rendered examples.

rmccorm4 (commented):
Really great work @nnshah1, this is looking awesome! 🚀 

Most of my comments in this round are nits/typos that would be public-facing in the docs -- so I provided "suggestions" so they can just be fixed with a button press. 

I think it would be good to [remove the "Todo" examples](https://github.com/triton-inference-server/core/pull/299/files#r1447843203) before publishing the rendered docs somewhere.

Most other comments (test nits, etc.) can be addressed in a follow-up PR if easier.

Happy to hit "approve" otherwise, let me know your thoughts.

python/test/test_api_models/test/1/model.py
rmccorm4: The more idiomatic way to provide some arbitrary config value would probably be to use [config parameters](https://github.com/triton-inference-server/tutorials/blob/336a7d713957c4944fb74b88e2f05baaf1a1d0d3/Quick_Deploy/HuggingFaceTransformers/llama7b/config.pbtxt#L6-L9).
```
parameters: {
  key: "decoupled",
  value: {string_value: "True"}
}
```

python/tritonserver/_api/_dlpack.py
nnshah1: after conversation with @oandreeva-nv synchronize on legacy default stream (1)

python/test/test_api_models/test/1/model.py
nnshah1: I was looking for exactly this and couldn't find it -

python/tritonserver/_api/_dlpack.py
nnshah1: currently _dlpack.py is unmodified from the version in the client repo - would like to keep it unmodified until we find a way to merge

python/test/test_api.py
```diff
@@ -0,0 +1,123 @@
...
+            fp16_output = numpy.from_dlpack(response.outputs["fp16_output"])
+            self.assertEqual(fp16_input, fp16_output)
+
+        for response in server.model("test").infer(
+            inputs={"fp16_input": fp16_input},
+            output_memory_type="gpu",
+            raise_on_error=True,
+        ):
+            fp16_output = cupy.from_dlpack(response.outputs["fp16_output"])
+            self.assertEqual(fp16_input[0][0], fp16_output[0][0])
```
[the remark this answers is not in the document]
nnshah1: actually, cupy.from_dlpack will assert if it is not on GPU

python/test/test_api.py
```diff
@@ -0,0 +1,252 @@
...
+        self.assertEqual(cpu_array.ctypes.data, cpu_fp32_array.ctypes.data)
+        self.assertEqual(cpu_fp32_array.dtype, numpy.float32)
+        self.assertEqual(cpu_fp32_array.nbytes, 200)
+
+        torch_fp32_tensor = torch.from_dlpack(tensor)
+        self.assertEqual(torch_fp32_tensor.dtype, torch.float32)
+        self.assertEqual(torch_fp32_tensor.data_ptr(), cpu_array.ctypes.data)
+        self.assertEqual(torch_fp32_tensor.nbytes, 200)
+
+    @pytest.mark.skipif(cupy is None, reason="Skipping gpu memory, cpupy not installed")
```
rmccorm4, line 95:
```suggestion
    @pytest.mark.skipif(cupy is None, reason="Skipping gpu memory, cupy not installed")
```

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -1953,7 +1962,8 @@ PYBIND11_MODULE(triton_bindings, m)
           &PyInferenceRequest::RemoveAllRequestedOutputs)
       .def("set_string_parameter", &PyInferenceRequest::SetStringParameter)
       .def("set_int_parameter", &PyInferenceRequest::SetIntParameter)
-      .def("set_bool_parameter", &PyInferenceRequest::SetBoolParameter);
+      .def("set_bool_parameter", &PyInferenceRequest::SetBoolParameter)
+      .def("cancel", &PyInferenceRequest::Cancel);
```
rmccorm4, line 1966: Would be good to test `cancel()` in follow-up if exposing it

rmccorm4 (dismissed):

python/test/test_api.py
nnshah1: will address post merge

python/tritonserver/_c/tritonserver_pybind.cc
nnshah1: will add post merge as seperate PR

rmccorm4 (approved):

rmccorm4 (approved):

python/tritonserver/_api/_allocators.py
```diff
@@ -0,0 +1,311 @@
...
+    data_ptr: int
+    memory_type: MemoryType
+    memory_type_id: int
+    size: int
+    owner: Any
+
+    @staticmethod
+    def from_dlpack(owner: Any) -> MemoryBuffer:
+        if not hasattr(owner, "__dlpack__"):
+            raise InvalidArgumentError("Object does not support DLpack protocol")
```
oandreeva-nv, line 94: Just a nit: Maybe change to "Object does not have `__dlpack__` method". Older versions of DLPack protocol allow passing a capsule irectly to `from_dlpack`

python/tritonserver/_api/_dlpack.py
oandreeva-nv: Sorry was late to the game, yes, we've discussed this offline

PR #306, opened 2023-12-29
Add Pinned Memory metrics

pskiran1:
tests: https://github.com/triton-inference-server/server/pull/6754


src/test/CMakeLists.txt
```diff
@@ -489,8 +491,19 @@ if(${TRITON_ENABLE_METRICS})
       GTest::gmock
       prometheus-cpp::core
       protobuf::libprotobuf
+      ${CNMEM_LIBRARY}
+      CUDA::cudart
```
rmccorm4, line 495: Why is the cuda link needed?

src/pinned_memory_manager.cc
```diff
@@ -376,4 +391,16 @@ PinnedMemoryManager::Free(void* ptr)
...
+uint64_t
+PinnedMemoryManager::GetTotalPinnedMemoryByteSize()
+{
+  return pinned_memory_byte_size_;
+}
+
+uint64_t
+PinnedMemoryManager::GetUsedPinnedMemoryByteSize()
+{
+  return used_pinned_memory_byte_size_;
```
rmccorm4, line 403: Are these accesses thread-safe? It seems like the regions where it is updated are protected by `info_mtx_`, which implies some concurrent usage to me.

src/test/CMakeLists.txt
pskiran1: Pinned memory modules use CUDA functions. To prevent errors during building, the above CUDA linkage has been added.

src/pinned_memory_manager.cc
pskiran1: Thank you for noting this. Since `info_mtx_` is an instance variable and cannot be used in static methods, created a specific mutex variable for reading and writing `used_pinned_memory_byte_size_` without modifying the existing `info_mtx_`.

src/pinned_memory_manager.cc
```diff
@@ -153,6 +155,15 @@ PinnedMemoryManager::AllocInternal(
     if (status.IsOk()) {
       auto res = memory_info_.emplace(
           *ptr, std::make_pair(is_pinned, pinned_memory_buffer));
+
+      {
+        std::lock_guard<std::mutex> lk(alloc_info_mtx_);
```
rmccorm4, line 160: Can you move this within the `if` block and remove the extra `{ }` scope brackets, if `is_pinned` doesn't need to be protected?

src/pinned_memory_manager.cc
```diff
@@ -192,6 +203,17 @@ PinnedMemoryManager::FreeInternal(void* ptr)
                      << "pinned memory deallocation: "
                      << "addr " << ptr;
       memory_info_.erase(it);
+
+      {
+        std::lock_guard<std::mutex> lk(alloc_info_mtx_);
```
rmccorm4, line 208: Similarly I think this mutex can be moved within the `if (ix != allocated_memory_info_.end())` block.

src/pinned_memory_manager.cc
```diff
@@ -153,6 +155,13 @@ PinnedMemoryManager::AllocInternal(
     if (status.IsOk()) {
       auto res = memory_info_.emplace(
           *ptr, std::make_pair(is_pinned, pinned_memory_buffer));
+
+      if (is_pinned) {
+        std::lock_guard<std::mutex> lk(alloc_info_mtx_);
+        used_pinned_memory_byte_size_ += size;
+        allocated_memory_info_.emplace(*ptr, size);
```
tanmayv25, line 162: Shouldn't we account for the pinned memory de-allocation performed in L180?
as in decrement and remove the used_pinned_memory_byte_size?

rmccorm4 (commented):
Seems generally OK to me, but pinged @tanmayv25 for second opinion on the extra locking introduced here.

src/pinned_memory_manager.cc
tanmayv25: How about moving the used_pinned_memory_byte_size within PinnedMemory class?
Then we can replace
```
*ptr = pinned_memory_buffer->managed_pinned_memory_.allocate(
        size, std::nothrow_t{});
```
and
```
pinned_memory_buffer->managed_pinned_memory_.deallocate(*ptr);
```
with `Allocate` and `Deallocate` calls respectively. These calls can internally keep record of how much memory is in use by this PinnedMemory object. 
I believe this should make the code more maintainable.

src/pinned_memory_manager.h
```diff
@@ -77,6 +77,9 @@ class PinnedMemoryManager {
   // Return Status object indicating success or failure.
   static Status Free(void* ptr);
 
+  static uint64_t GetTotalPinnedMemoryByteSize();
+  static uint64_t GetUsedPinnedMemoryByteSize();
```
tanmayv25, line 81: Building upon my previous suggestion, as this static function can operate on static variable, you can create a static set in this class that holds all the PinnedMemoryManager objects created and individually accumulates the values from the specific objects.

This suggestion ensures that allocate and deallocate are always accompanied by allocated_memory_info update.

tanmayv25: Additionally in the destructor you'd have to ensure that object is removed from this static set.

src/pinned_memory_manager.cc
```diff
@@ -168,7 +199,7 @@ PinnedMemoryManager::AllocInternal(
   if ((!status.IsOk()) && (*ptr != nullptr)) {
     if (is_pinned) {
       std::lock_guard<std::mutex> lk(pinned_memory_buffer->buffer_mtx_);
```
tanmayv25, line 201: Move these mutex locks within Allocate and Deallocate function.

src/pinned_memory_manager.h
```diff
@@ -99,6 +108,8 @@ class PinnedMemoryManager {
 
   static std::unique_ptr<PinnedMemoryManager> instance_;
   static uint64_t pinned_memory_byte_size_;
+  static std::vector<std::shared_ptr<PinnedMemory>>
```
tanmayv25, line 111: You should add a static mutex to protect the allocated_pinned_memory_buffers_  vector.

pskiran1: Done 👍

src/pinned_memory_manager.cc
pskiran1: Done 👍

rmccorm4 (approved):

PR #308, opened 2024-01-03
Add retry on model loading. Expose option to set model retry count

GuanLuo:


python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -1282,6 +1282,12 @@ class PyServerOptions : public PyWrapper<struct TRITONSERVER_ServerOptions> {
         triton_object_, thread_count));
   }
 
+  void SetModelLoadRetryCount(unsigned int retry_count)
```
rmccorm4, line 1285: Unrelated, but do you think we could have some kind of test that asserts the python bindings fully cover the C API?

I'm thinking long term, folks may not always remember to add the binding equivalent for any API changes, and it might be good to automate that check.

src/model_repository_manager/model_lifecycle.cc
```diff
@@ -513,7 +513,15 @@ ModelLifeCycle::AsyncLoad(
     // Load model asynchronously via thread pool
     load_pool_->Enqueue([this, model_id, version, model_info, OnComplete,
                          load_tracker, is_config_provided]() {
-      CreateModel(model_id, version, model_info, is_config_provided);
+      for (size_t retry = 0; retry <= options_.load_retry; ++retry) {
+        model_info->state_ = ModelReadyState::LOADING;
+        CreateModel(model_id, version, model_info, is_config_provided);
+        // Model state will be changed to NOT loading if failed to load,
+        // so the model is loaded if state is LOADING.
+        if (model_info->state_ == ModelReadyState::LOADING) {
```
nnshah1, line 521: question - is there a pause / is there one needed between retry attempts?

GuanLuo: Why would one be needed?

nnshah1: typically you after some time in order to let a possible transient error resolve - instead of retrying immediately

Tabrizian (approved):

src/model_repository_manager/model_lifecycle.h
```diff
@@ -45,21 +45,23 @@ struct ModelLifeCycleOptions {
...
+  const double min_compute_capability;
   // The backend configuration settings specified on the command-line
-  const triton::common::BackendCmdlineConfigMap& backend_cmdline_config_map_;
+  const triton::common::BackendCmdlineConfigMap& backend_cmdline_config_map;
   // The host policy setting used when loading models.
-  const triton::common::HostPolicyCmdlineConfigMap& host_policy_map_;
+  const triton::common::HostPolicyCmdlineConfigMap& host_policy_map;
   // Number of the threads to use for concurrently loading models
-  const unsigned int model_load_thread_count_;
+  const unsigned int model_load_thread_count;
```
kthui, line 62: Nice refactoring!

kthui (approved):

PR #311, opened 2024-01-12
added additional error handling and test cases

nnshah1:
Added error handling for memory allocation failures
Added faillback to cpu memory if requested memory not supported
Added testing to cover these scenarios

### DIFF
--- a/python/test/test_api.py
+++ b/python/test/test_api.py
@@ -71,6 +71,119 @@ class ModelTests(unittest.TestCase):
 
 
 class AllocatorTests(unittest.TestCase):
+    class MockMemoryAllocator(tritonserver.MemoryAllocator):
+        def __init__(self):
+            pass
+
+        def allocate(self, *args, **kwargs):
+            raise Exception("foo")
+
+    @pytest.mark.skipif(cupy is None, reason="Skipping gpu memory, cupy not installed")
+    def test_memory_fallback_to_cpu(self):
+        server = tritonserver.Server(server_options).start(wait_until_ready=True)
+
+        self.assertTrue(server.ready())
+
+        allocator = tritonserver.default_memory_allocators[tritonserver.MemoryType.GPU]
+
+        del tritonserver.default_memory_allocators[tritonserver.MemoryType.GPU]
+
+        server.load(
+            "test",
+            {
+                "config": json.dumps(
+                    {
+                        "backend": "python",
+                        "parameters": {
+                            "decoupled": {"string_value": "False"},
+                            "request_gpu_memory": {"string_value": "True"},
+                        },
+                    }
+                )
+            },
+        )
+
+        fp16_input = numpy.random.rand(1, 100).astype(dtype=numpy.float16)
+
+        for response in server.model("test").infer(
+            inputs={"fp16_input": fp16_input},
+        ):
+            self.assertEqual(
+                response.outputs["fp16_output"].memory_type, tritonserver.MemoryType.CPU
+            )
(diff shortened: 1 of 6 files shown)


python/tritonserver/_api/_logging.py
```diff
@@ -0,0 +1,61 @@
...
+from tritonserver._c.triton_bindings import TRITONSERVER_LogMessage
+
+
+def LogMessage(level: LogLevel, message: str):
+    """Log Message using Triton Inference Server Logger
+
+    Parameters
+    ----------
+    level : LogLevel
+        log level one of LogLevel.WARN, LogLevel.ERROR, LogLeve.INFO
```
nnshah1, line 41:
```suggestion
        log level one of LogLevel.WARN, LogLevel.ERROR, LogLevel.INFO
```

python/tritonserver/_api/_logging.py
```diff
@@ -0,0 +1,61 @@
...
+
+"""Logging Utilities"""
+import inspect
+import os
+
+from tritonserver._c.triton_bindings import TRITONSERVER_LogLevel as LogLevel
+from tritonserver._c.triton_bindings import TRITONSERVER_LogMessage
+
+
+def LogMessage(level: LogLevel, message: str):
```
rmccorm4, line 35: Might be nice to have some class or syntactic sugar to do `logger.error(message)` or `log_error()` or similar in the future. Not pertaining to this PR or release though.

python/tritonserver/_api/_allocators.py
```diff
@@ -238,24 +247,38 @@ def allocate(
...
+            return (
+                memory_buffer.data_ptr,
+                memory_buffer,
+                memory_buffer.memory_type,
+                memory_buffer.memory_type_id,
+            )
+        except Exception as e:
+            message = f"Catastrophic failure in allocator: {e}, returning NULL"
+            LogMessage(LogLevel.ERROR, message)
+            return (0, None, MemoryType.CPU, 0)
```
rmccorm4, line 281: Not sure if it's possible, but might be nice to have some namedtuple or something like `INVALID_MEMORY` for use/comparison purposes that is just this tuple under the hood (in the future)

python/test/test_api.py
```diff
@@ -71,6 +71,119 @@ def test_create_request(self):
...
+
+        server.load(
+            "test",
+            {
+                "config": json.dumps(
+                    {
+                        "backend": "python",
+                        "parameters": {
+                            "decoupled": {"string_value": "False"},
+                            "request_gpu_memory": {"string_value": "True"},
```
rmccorm4, line 99: If you set the `requested_output_memory_type = GPU`, would python backend also give you back a GPU buffer without having to use this custom parameter?

rmccorm4 (approved):
LGTM, comments are just a question, and some future ideas/thoughts

python/test/test_api.py
nnshah1: My understanding is that the backend requests memory from the client for each output tensor - and it can query the allocator to know what type of memory the client will provide. So the requested memory type is the backend requesting a memory type from the client - and exactly the behavior we are relying on for testing. The python backend (turns out) will request GPU memory from the client if the output tensor produced by the model is on the GPU .... another way to think of it is that the client / response allocator always has the final say in the memory type - the backend can only 'request' it.

python/tritonserver/_api/_allocators.py
nnshah1: we could subclass from a tuple - or actually we can just create a singleton for invalid memory

PR #317, opened 2024-01-17
updated to save server logs and separate stop into separate test

nnshah1:


python/test/test_api.py
```diff
@@ -311,14 +338,51 @@ def test_invalid_repo(self):
...
+            inputs={"fp16_input": fp16_input},
+            output_memory_type="cpu",
+            raise_on_error=True,
+        ):
+            fp16_output = numpy.from_dlpack(response.outputs["fp16_output"])
+            numpy.testing.assert_array_equal(fp16_input, fp16_output)
+        try:
+            server.stop()
+        except tritonserver.InternalError as e:
+            pytest.xfail(f"Server failed to stop! {e}")
```
rmccorm4, line 373: I think it would be easier to find/analyze the failures for debugging if we let the test fail on gitlab side of things. I think we can let the test fail as a known failure for release purposes.

rmccorm4 (commented):
LGTM, just one comment on the `xfail`

python/test/test_api.py
nnshah1: updated to use xfail only for versions <= "2.42.0" to indicate the test is expected to fail / is a known issue.

python/test/test_api.py
```diff
@@ -384,5 +452,3 @@ def test_basic_inference(self):
         ):
             fp16_output = numpy.from_dlpack(response.outputs["fp16_output"])
             numpy.testing.assert_array_equal(fp16_input, fp16_output)
-
-        server.stop()
```
Tabrizian, line 388: question: why is stop removed? Is the server automatically destructed if object goes out of scope? If so, you might need explicit `del` otherwise you are relying on garbage collector behaviour which may not happen immediately.

nnshah1: server does get destroyed and stop called when it goes out of scope - garbage collector would only come into picture if there is a dangling reference - so don't think explicit del will alter behavior in this case -

rmccorm4 (approved):
If the test is actually intermittent, then I think it will fail when the xfail subtest passes now, compared to what you had before with the inline `try/except->xfail`. But if it fails most of the time, this is fine, we can tweak it later if needed. 🤷

Tabrizian (approved):
Hopefully the logs can help us identify the real issue.

nnshah1:
> If the test is actually intermittent, then I think it will fail when the xfail subtest passes now, compared to what you had before with the inline `try/except->xfail`. But if it fails most of the time, this is fine, we can tweak it later if needed. 🤷

quick note for reference - xfail by default allows for test to pass - there is a 'strict' option https://docs.pytest.org/en/7.1.x/how-to/skipping.html#strict-parameter that will fail the test if it passes ....

nnshah1:
> Hopefully the logs can help us identify the real issue.

it might be as you suggested that timing of stops of other tests ... not sure - we can experiment with explicit del and explicit stop to see if it resolves. Currently - for whatever reason the basic inference and the stop test have the issue - but the test gpu output does not ... - which is odd to me.

PR #322, opened 2024-01-22
Moving model repo None check to align with other checks

nnshah1:
Currently all errors on options are thrown at server.start() - moved the check and exception to align with that and moved test to server section.

### DIFF
--- a/python/test/test_api.py
+++ b/python/test/test_api.py
@@ -376,6 +376,10 @@ class ServerTests(unittest.TestCase):
 
         server.stop()
 
+    def test_model_repository_not_specified(self):
+        with self.assertRaises(tritonserver.InvalidArgumentError):
+            tritonserver.Server(model_repository=None).start()
+
 
 class InferenceTests(unittest.TestCase):
     def setup_method(self, method):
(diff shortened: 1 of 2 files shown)


Tabrizian (approved):

Tabrizian:
> Currently all errors on options are thrown at server.start() - moved the check and exception to align with that and moved test to server section.

I think if we are able to detect it when the server is being created rather than started it might be better to move those checks to server creation.

nnshah1:
> > Currently all errors on options are thrown at server.start() - moved the check and exception to align with that and moved test to server section.
> 
> I think if we are able to detect the error when the server is being created rather than started it might be better to move those checks to server creation.

currently `server()` will not throw any errors and `server.start()` will create options and start the server - throwing errors if either fails.

We can move the option creation to `server()` and fail on `option errors` there - but also like the simplicity of all initialization errors being in one place. 

(For example if model repo is None or non existant both errors would be returned at `start()`. If we move the option creation around then model repo None would be through at `server()` and non existant repo at `start()`

nnshah1:
@Tabrizian - Will go ahead and merge this small update and we can address the other separately ?

PR #325, opened 2024-02-07
Add response statistics

kthui:
Related PRs:
- https://github.com/triton-inference-server/server/pull/6869
- https://github.com/triton-inference-server/common/pull/112
- https://github.com/triton-inference-server/square_backend/pull/16

Add response statistics API.

Example statistics endpoint output to client:
```
{
  'model_stats': [{
    'name': 'square_int32',
    ...
    'inference_stats': {
      ...
      'response_stats': {
        '0': {
          'compute_infer': {'count': 3, 'ns': 1200514434},
          'compute_output': {'count': 3, 'ns': 600346215},
          'success': {'count': 2, 'ns': 1200431061},
          'fail': {'count': 1, 'ns': 600429588},
          'empty_response': {'count': 0, 'ns': 0}
        },
        '1': {
          'compute_infer': {'count': 3, 'ns': 1200250644},
          'compute_output': {'count': 3, 'ns': 600358040},
          'success': {'count': 1, 'ns': 600202056},
          'fail': {'count': 2, 'ns': 1200406628},
          'empty_response': {'count': 0, 'ns': 0}
        },
        '2': {
          'compute_infer': {'count': 3, 'ns': 1200252841},
          'compute_output': {'count': 2, 'ns': 400214433},
          'success': {'count': 1, 'ns': 600183751},
          'fail': {'count': 1, 'ns': 600188918},
          'empty_response': {'count': 1, 'ns': 400094605}
        },
        '3': {
          'compute_infer': {'count': 2, 'ns': 800160632},
          'compute_output': {'count': 1, 'ns': 200133237},
          'success': {'count': 0, 'ns': 0},
          'fail': {'count': 1, 'ns': 600214152},
          'empty_response': {'count': 1, 'ns': 400079717}
        },
        '4': {
          'compute_infer': {'count': 1, 'ns': 400086093},
          'compute_output': {'count': 1, 'ns': 200104114},
          'success': {'count': 0, 'ns': 0},
          'fail': {'count': 1, 'ns': 600190207},
          'empty_response': {'count': 0, 'ns': 0}
        },
        '5': {
          'compute_infer': {'count': 1, 'ns': 400089434},
          'compute_output': {'count': 0, 'ns': 0},
          'success': {'count': 0, 'ns': 0},
          'fail': {'count': 0, 'ns': 0},
          'empty_response': {'count': 1, 'ns': 400089434}}
        }
      }
    ...
    }
  }]
}
```

### DIFF
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -1337,16 +1337,16 @@ TRITONBACKEND_ModelInstanceReportMemoryUsage(
 /// TRITONBACKEND_ModelInstanceExecute.
 ///
 ///   TRITONBACKEND_ModelInstanceExecute()
-///     CAPTURE TIMESPACE (exec_start_ns)
+///     CAPTURE TIMESTAMP (exec_start_ns)
 ///     < process input tensors to prepare them for inference
 ///       execution, including copying the tensors to/from GPU if
 ///       necessary>
-///     CAPTURE TIMESPACE (compute_start_ns)
+///     CAPTURE TIMESTAMP (compute_start_ns)
 ///     < perform inference computations to produce outputs >
-///     CAPTURE TIMESPACE (compute_end_ns)
+///     CAPTURE TIMESTAMP (compute_end_ns)
 ///     < allocate output buffers and extract output tensors, including
 ///       copying the tensors to/from GPU if necessary>
-///     CAPTURE TIMESPACE (exec_end_ns)
+///     CAPTURE TIMESTAMP (exec_end_ns)
 ///     return
 ///
 /// Note that these statistics are associated with a valid
@@ -1385,12 +1385,12 @@ TRITONBACKEND_ModelInstanceReportStatistics(
 ///
 ///   TRITONBACKEND_ModelInstanceExecute()
 ///     < start of this response >
-///     CAPTURE TIMESPACE (response_start)
+///     CAPTURE TIMESTAMP (response_start)
 ///     < generate this response >
-///     CAPTURE TIMESPACE (compute_output_start)
+///     CAPTURE TIMESTAMP (compute_output_start)
 ///     < allocate output buffers and extract output tensors, including copying
 ///       the tensors to/from GPU if necessary >
-///     CAPTURE TIMESPACE (response_end)
+///     CAPTURE TIMESTAMP (response_end)
 ///     < end of this response >
 ///     return
 ///
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -1,4 +1,4 @@
-// Copyright 2020-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2020-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
@@ -1355,6 +1355,51 @@ TRITONBACKEND_ModelInstanceReportStatistics(
     const bool success, const uint64_t exec_start_ns,
     const uint64_t compute_start_ns, const uint64_t compute_end_ns,
     const uint64_t exec_end_ns);
+
+/// Record statistics for a decoupled inference response.
+///
+/// All timestamps should be reported in nanonseconds and collected using
+/// std::chrono::steady_clock::now().time_since_epoch() or the equivalent.
+///
+/// Call this function after calling TRITONBACKEND_ResponseSend, and pass the
+/// same send_flags and error object passed to the TRITONBACKEND_ResponseSend.
+///
+/// For consistency of measurement across different backends, the
+/// timestamps should be collected at the following points during
+/// TRITONBACKEND_ModelInstanceExecute.
+///
+///   TRITONBACKEND_ModelInstanceExecute()
+///     < start of this response >
+///     CAPTURE TIMESPACE (response_start)
+///     < generate this response >
+///     CAPTURE TIMESPACE (compute_output_start)
+///     < allocate output buffers and extract output tensors, including copying
+///       the tensors to/from GPU if necessary >
+///     CAPTURE TIMESPACE (response_end)
+///     < end of this response >
+///     return
+///
+/// \param instance The model instance.
+/// \param response_factory The response factory associated with the inference
+/// request.
+/// \param response_start Timestamp for the start of execution for this
+/// response.
+/// \param compute_output_start Timestamp for the start of extracting output
+/// tensors for this response. Set this to 0 for reporting empty response.
+/// \param response_end Timestamp for the end of extracting output tensors for
+/// this response.
+/// \param send_flags Flags associated with the response. \see
+/// TRITONBACKEND_ResponseSend \see
+/// \param error The TRITONSERVER_Error to send if the response is an error, or
+/// nullptr if the response is successful. \see TRITONBACKEND_ResponseSend \see
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONBACKEND_DECLSPEC TRITONSERVER_Error*
+TRITONBACKEND_ModelInstanceReportResponseStatistics(
+    TRITONBACKEND_ModelInstance* instance,
+    TRITONBACKEND_ResponseFactory* response_factory,
+    const uint64_t response_start, const uint64_t compute_output_start,
+    const uint64_t response_end, const uint32_t send_flags,
+    TRITONSERVER_Error* error);
 
 /// Record statistics for the execution of an entire batch of
 /// inference requests.
--- a/include/triton/core/tritonbackend.h
+++ b/include/triton/core/tritonbackend.h
@@ -68,6 +68,23 @@ struct TRITONBACKEND_ModelInstance;
 struct TRITONBACKEND_BackendAttribute;
 struct TRITONBACKEND_Batcher;
 
+struct TRITONBACKEND_ModelInstanceResponseStatistics {
+#ifdef TRITON_ENABLE_STATS
+  TRITONBACKEND_ModelInstanceResponseStatistics()
+      : model_instance(nullptr), response_factory(nullptr), response_start(0),
+        compute_output_start(0), response_end(0), send_flags(0), error(nullptr)
+  {
+  }
+  TRITONBACKEND_ModelInstance* model_instance;
+  TRITONBACKEND_ResponseFactory* response_factory;
+  uint64_t response_start;
+  uint64_t compute_output_start;
+  uint64_t response_end;
+  uint32_t send_flags;
+  TRITONSERVER_Error* error;
+#endif  // TRITON_ENABLE_STATS
+};
+
 ///
 /// TRITONBACKEND API Version
 ///
@@ -1362,11 +1379,8 @@ TRITONBACKEND_ModelInstanceReportStatistics(
 /// All timestamps should be reported in nanonseconds and collected using
 /// std::chrono::steady_clock::now().time_since_epoch() or the equivalent.
 ///
-/// Call this function after calling TRITONBACKEND_ResponseSend, and pass the
-/// same send_flags and error object passed to the TRITONBACKEND_ResponseSend.
-///
-/// For consistency of measurement across different backends, the
-/// timestamps should be collected at the following points during
+/// For consistency of measurement across different backends, the timestamps
+/// should be collected at the following points during
 /// TRITONBACKEND_ModelInstanceExecute.
 ///
 ///   TRITONBACKEND_ModelInstanceExecute()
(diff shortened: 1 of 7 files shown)


include/triton/core/tritonbackend.h
```diff
@@ -1356,6 +1356,51 @@ TRITONBACKEND_ModelInstanceReportStatistics(
...
+/// \param response_factory The response factory associated with the inference
+/// request.
+/// \param response_start Timestamp for the start of execution for this
+/// response.
+/// \param compute_output_start Timestamp for the start of extracting output
+/// tensors for this response. Set this to 0 for reporting empty response.
+/// \param response_end Timestamp for the end of extracting output tensors for
+/// this response.
+/// \param send_flags Flags associated with the response. \see
+/// TRITONBACKEND_ResponseSend \see
```
oandreeva-nv, line 1392: do we need both `\see`?

include/triton/core/tritonbackend.h
```diff
@@ -1356,6 +1356,51 @@ TRITONBACKEND_ModelInstanceReportStatistics(
...
+/// \param error The TRITONSERVER_Error to send if the response is an error, or
+/// nullptr if the response is successful. \see TRITONBACKEND_ResponseSend \see
```
oandreeva-nv, line 1394: ditto

src/infer_response.h
```diff
@@ -61,6 +61,10 @@ class InferenceResponseFactory {
         alloc_userp_(alloc_userp), response_fn_(response_fn),
         response_userp_(response_userp), response_delegator_(delegator),
         is_cancelled_(false)
+#ifdef TRITON_ENABLE_STATS
+        ,
+        response_stats_index_(0)
```
oandreeva-nv, line 66: did pre-commit hook put comma and parameter on two lines?

src/infer_response.h
```diff
@@ -94,6 +98,11 @@ class InferenceResponseFactory {
   void ReleaseTrace() { trace_ = nullptr; }
 #endif  // TRITON_ENABLE_TRACING
 
+#ifdef TRITON_ENABLE_STATS
+  // Return the current response statistics index and increment it.
+  uint64_t ResponseStatsIndex() { return response_stats_index_++; };
```
oandreeva-nv, line 103: Could you briefly explain the logic behind why we need to increment it here?

kthui: We want to increment to the next index, atomically. I think we can adjust the function naming if it makes it better?

src/infer_response.h
kthui: Yes, pre-commit did it.

include/triton/core/tritonbackend.h
kthui: No, It's updated.

include/triton/core/tritonbackend.h
kthui: Updated.

include/triton/core/tritonbackend.h
```diff
@@ -1356,6 +1356,51 @@ TRITONBACKEND_ModelInstanceReportStatistics(
...
+/// TRITONBACKEND_ResponseSend
+/// \param error The TRITONSERVER_Error to send if the response is an error, or
+/// nullptr if the response is successful. \see TRITONBACKEND_ResponseSend
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONBACKEND_DECLSPEC TRITONSERVER_Error*
+TRITONBACKEND_ModelInstanceReportResponseStatistics(
+    TRITONBACKEND_ModelInstance* instance,
+    TRITONBACKEND_ResponseFactory* response_factory,
+    const uint64_t response_start, const uint64_t compute_output_start,
+    const uint64_t response_end, const uint32_t send_flags,
```
rmccorm4, line 1401: Not a main reviewer here, but @Tabrizian @kthui @oandreeva-nv do you think we should use another Opaque object here for stats/info if we need to add more fields to this in the future? Currently appending another field would require a new API.

If we sent a `TRITONBACKEND_ResponseStats` object or something, we can append to it?

This could also be the object to denote if the request was cancelled for phase 2 , etc

rmccorm4: Also for backends that implement inflight batching themselves, or perhaps some other logic I'm not currently thinking of, backends may maintain their own queues. This could also be a place to report how long a request was "queued" from the backend's perspective, rather than triton core scheduler perspective.

Just food for thought - may not make total sense yet.

oandreeva-nv: Yes, me and @kthui had an offline discussion that in future there may be a need for new API wich handles `key` and does not use `response_factory`. I think some opaque object could help

src/infer_response.h
oandreeva-nv: Yes I agree with the name adjustment, my concern is that as is `ResponseStatsIndex` implies that the function just returns index. But with the current logic we increment it as well.

oandreeva-nv: cc @Tabrizian in case he has a feedback

include/triton/core/tritonbackend.h
kthui: I think you are on the right track @rmccorm4. My thought ultimately is to allow the backend to specify a "key" which is used to group/aggregate the "same" response statistics across different requests. I will leave the decision to @Tabrizian if we want to switch to / implement the "key" now.

include/triton/core/tritonbackend.h
```diff
@@ -1356,6 +1356,51 @@ TRITONBACKEND_ModelInstanceReportStatistics(
...
+/// request.
+/// \param response_start Timestamp for the start of execution for this
+/// response.
+/// \param compute_output_start Timestamp for the start of extracting output
+/// tensors for this response. Set this to 0 for reporting empty response.
+/// \param response_end Timestamp for the end of extracting output tensors for
+/// this response.
+/// \param send_flags Flags associated with the response. \see
+/// TRITONBACKEND_ResponseSend
+/// \param error The TRITONSERVER_Error to send if the response is an error, or
```
Tabrizian, line 1393: would be good to clarify the ownership of the error object (i.e. it is owned the caller and should be freed if it is provided).

include/triton/core/tritonbackend.h
Tabrizian: @rmccorm4 what is the use-case you have in mind? Is it possible for the user to use custom metrics for anything that doesn't fit into this structure?

> The distinction between request-specific and response-specific is a little fuzzy for me atm.

For decoupled models, currently there is no way to access the latency breakdown for individual responses. These statistics help the user to get a better idea about the latency breakdown of individual responses.

src/infer_response.h
Tabrizian: Agreed with Olga. Perhaps we can rename it to `GetAndIncrementResponseIndex` to make it more clear.

include/triton/core/tritonbackend.h
kthui: Good point. The ownership is clarified on the comments.

src/infer_response.h
kthui: Sure. Renamed to `GetAndIncrementResponseIndex`.

include/triton/core/tritonbackend.h
rmccorm4: I think custom metrics *can* be used for things that fall outside of the current stats, but was just curious if we had any ideas of other stats that may be useful to add to the "core" set in the near-mid term. If so, we would have to add a new C API for new stats if we didn't use some opaque object.

For example, if the use of iterative scheduling / IFB handled in Triton core had more adoption - would we want some notion of per-step stats for rescheduled requests? I think this case might be analogous to the "empty/flag-only" intermediate responses, where we don't actually produce a response but want to somehow signify that we have done some work in between.

Tabrizian: > I think this case might be analogous to the "empty/flag-only" intermediate responses, where we don't actually produce a response but want to somehow signify that we have done some work in between.

I don't think response stats is the right place for this. If you are sending a response, you can use response stats.

include/triton/core/tritonbackend.h
```diff
@@ -1356,6 +1357,53 @@ TRITONBACKEND_ModelInstanceReportStatistics(
...
+/// Call this function after calling TRITONBACKEND_ResponseSend, and pass the
+/// same send_flags and error object passed to the TRITONBACKEND_ResponseSend.
+///
+/// For consistency of measurement across different backends, the
+/// timestamps should be collected at the following points during
+/// TRITONBACKEND_ModelInstanceExecute.
+///
+///   TRITONBACKEND_ModelInstanceExecute()
+///     < start of this response >
+///     CAPTURE TIMESPACE (response_start)
```
oandreeva-nv, line 1374: timespace -> timestamp?

kthui: Good catch! It was copied from the `TRITONBACKEND_ModelInstanceReportResponseStatistics()` API above, it is also fixed there.

include/triton/core/tritonbackend.h
```diff
@@ -1374,7 +1358,46 @@ TRITONBACKEND_ModelInstanceReportStatistics(
     const uint64_t compute_start_ns, const uint64_t compute_end_ns,
     const uint64_t exec_end_ns);
 
-/// Record statistics for a decoupled inference response.
+/// Create a new decoupled inference response statistics object.
```
rmccorm4, line 1361: Is mention of "decoupled" necessary in these docs? I think a non-decoupled backend could choose to publish for a single response if they wanted, right? (though maybe not that useful currently)

kthui: Sure, removed the word decoupled from comments.

src/infer_stats.cc
```diff
@@ -210,6 +210,132 @@ InferenceStatsAggregator::UpdateSuccessCacheMiss(
...
 
+Status
+InferenceStatsAggregator::UpdateResponseSuccess(
+    const std::string& key, const uint64_t response_start_ns,
+    const uint64_t compute_output_start_ns, const uint64_t response_end_ns)
+{
+  if (response_start_ns > compute_output_start_ns) {
+    return Status(
+        Status::Code::INVALID_ARG,
+        "Response start cannot happen after compute output start");
```
oandreeva-nv, line 221: Should we make it more descriptive? Something like:
"Unexpected behavior: response start timestamp was recorded after compute output start"

src/infer_stats.cc
```diff
@@ -210,6 +210,132 @@ InferenceStatsAggregator::UpdateSuccessCacheMiss(
...
+  }
+  if (compute_output_start_ns > response_end_ns) {
+    return Status(
+        Status::Code::INVALID_ARG,
+        "Compute output start cannot happen after response end");
```
oandreeva-nv, line 226: ditto

oandreeva-nv (dismissed):
TMLG, but I would also let @rmccorm4 take a look

src/backend_model_instance.cc
```diff
@@ -795,6 +795,23 @@ TritonModelInstance::TritonBackendThread::BackendThread()
   LOG_VERBOSE(1) << "Stopping backend thread for " << name_ << "...";
 }
 
+// Opaque object for the response statistics C-API
```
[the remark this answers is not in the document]
kthui: We can skip mentioning it is an opaque object, but it is important to mention this struct is only for the response statistics C-API and nothing else. When we refactor the `InferenceStatsAggregator()` in the future (which was written in a non-object-orientated style unfortunately but I don't want to mess with it on this PR), we can remove this struct and all the C-API set calls should interact directly with the `InferenceStatsAggregator()` (or a helper class object within).

Maybe we can just change the comment to `// For response statistics C-API only`?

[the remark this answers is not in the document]
oandreeva-nv: I'm ok with keeping it as is in this PR then

src/infer_stats.cc
kthui: I don't know if this is cleaner, because the user could have recorded the `response start` first and then followed by the `compute output start`, but reported the `compute output start` as the `response start` and `response start` as the `compute output start`. Then, the message could send the user looking at the wrong place, which is why I want to avoid speculating on what the user did, but simply say what is passed into the API.

Maybe we can write
```
Response start cannot happen after compute output start, check response statistics timestamps.
```
?

oandreeva-nv: We ca keep it as is then=)

include/triton/core/tritonbackend.h
```diff
@@ -1356,6 +1358,154 @@ TRITONBACKEND_ModelInstanceReportStatistics(
...
+/// \param response_statistics The response statistics object.
+/// \param response_start The response start time.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONBACKEND_DECLSPEC TRITONSERVER_Error*
+TRITONBACKEND_ModelInstanceResponseStatisticsSetResponseStart(
+    TRITONBACKEND_ModelInstanceResponseStatistics* response_statistics,
+    uint64_t response_start);
+
+/// Set compute output start time to an inference response statistics object.
+/// Set this to 0 for reporting an empty response.
```
rmccorm4, line 1429:
```suggestion
/// NOTE: Set this to 0 for reporting an empty response.
```

rmccorm4, line 1429: Given that we have an opaque object now, I feel like the correct way to specify empty would some kind of bool or flag. I think this is probably fine as-is, but I wish we had a some example use cases for this feature. In my mind, there could be models that do some compute work, but decide it's not ready for a full response yet. Having some way to specify "work was done" (compute > 0), but also "empty response" may be useful..

It's hard to say without some concrete use case for this feature. What do you think?

rmccorm4 (commented):
LGTM overall, main comment is on the use of "zero" compute time for "empty".

src/tritonserver.cc
```diff
@@ -2933,6 +2935,35 @@ TRITONSERVER_ServerModelStatistics(
           metadata, inference_stats, "cache_miss",
           infer_stats.cache_miss_count_, infer_stats.cache_miss_duration_ns_);
 
+      // Add response statistics
+      triton::common::TritonJson::Value response_stats(
```
rmccorm4, line 2939: Can you add a sample output of stats endpoint including these new stats to the PR description?

kthui: Sure.

include/triton/core/tritonbackend.h
kthui: The user actually doesn't need to set this value if reporting an empty response.

kthui: Updated the comment.

kthui: The response statistics collects 3 timestamps and translates into 2 durations:
```
response start    -->    compute output start    -->    response end
        v                           v                         v
        |       compute infer       |      compute output     |
```
On you example, if there is no compute output start time, which implies the response is empty, then there is no compute output duration which the compute infer duration is from response start to response end, so the graph for empty becomes:
```
response start    -->   response end
        v                     v
        |    compute infer    |
```
I think your example is the generic case where users should use empty response, because some compute work is done but there is no output tensor, so there is no coping of output tensor, which there is no compute output duration. The only purpose for the compute output start timestamp is to break the total duration from response start to end into compute infer and output durations. Since there is no compute output duration for empty response (if and only if), the compute output start timestamp can be repurposed for denoting empty response.

I think it is more natural to use `compute output start = 0` to denote an empty response, because the user should not record the compute output start timestamp if copying output tensors never takes place. Then, the user can simply skip setting compute output start time (or set it to 0 if it fits the workflow better), and it is automatically classified as an empty response, without needing the user to additionally report it is an empty response or not. What do you think?

kthui: I think `#5` is an example for this:
```
        '5': {
          'compute_infer': {'count': 1, 'ns': 400089434},
          'compute_output': {'count': 0, 'ns': 0},
          'success': {'count': 0, 'ns': 0},
          'fail': {'count': 0, 'ns': 0},
          'empty_response': {'count': 1, 'ns': 400089434}}
        }
```
There is an empty response with compute duration, but not output duration.

kthui: I think `#3` is an interesting example:
```
        '3': {
          'compute_infer': {'count': 2, 'ns': 800160632},
          'compute_output': {'count': 1, 'ns': 200133237},
          'success': {'count': 0, 'ns': 0},
          'fail': {'count': 1, 'ns': 600214152},
          'empty_response': {'count': 1, 'ns': 400079717}
        },
```
There is a fail and an empty response. Regardless of result, compute infer always take 4 seconds and compute output always take 2 seconds (if any). On the empty response, there is no compute output but compute infer, so the total duration is 4 seconds. On the fail response, there is compute infer and output, so the total duration is 6 seconds. There is only 1 compute output from the fail response, so there is only 1 compute output totaling 2 second. There is a compute infer from fail and empty, so there are 2 compute infer totaling 8 seconds.

rmccorm4 (approved):
Nice work! :rocket:

include/triton/core/tritonbackend.h
rmccorm4: Looks like "success" here also implies "non-empty" btw.
>          'success': {'count': 0, 'ns': 0},
>          'empty_response': {'count': 1, 'ns': 400079717}

kthui: Yes, this is expected.

A summary of where the response should be grouped into (i.e. success, fail, empty):

|   | no failure (success) | has failure |
| --- | --- | --- |
| **sent response** | success | fail |
| **no response sent** | empty | fail |

PR #327, opened 2024-02-20
updating to skip torch tests if torch.cuda not available

nnshah1:
- Updated to skip tests requiring torch for gpu tensors if torch.cuda is not available.
- Updated to expect failure on stop test as not yet root caused.

### DIFF
--- a/python/test/test_api.py
+++ b/python/test/test_api.py
@@ -44,6 +44,9 @@ except ImportError:
 
 try:
     import torch
+
+    if not torch.cuda.is_available():
+        torch = None
 except ImportError:
     torch = None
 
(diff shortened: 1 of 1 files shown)


python/test/test_api.py
```diff
@@ -343,7 +346,7 @@ def test_ready(self):
         self.assertTrue(server.ready())
 
     @pytest.mark.xfail(
-        tritonserver.__version__ <= "2.42.0",
+        tritonserver.__version__ <= "2.43.0",
```
mc-nv, line 349: I see we sequentially changing version. <br>
Is this something that will have place in for version `2.44.0`?

nnshah1: Not yet scheduled / prioritized.

mc-nv (approved):

PR #331, opened 2024-02-29
updating to skip torch tests if torch.cuda not available

mc-nv:


python/test/test_api.py
```diff
@@ -44,6 +44,9 @@
 
 try:
     import torch
+
+    if not torch.cuda.is_available():
+        torch = None
```
jbkyang-nvi, line 49: what does this fix?

mc-nv: @nnshah1 can you help by providing comment on your change

nnshah1: The test relies on not just torch, but torch.cuda to be available. There was a configuration / build in which torch was installed but cuda portion was not.

To safeguard I added the check - and if torch.cuda is not available treat that the same as torch being unavailable.

nnshah1 (approved):

PR #333, opened 2024-03-04
Promptly respond to timeout requests under reject policy

kthui:
Related PR: https://github.com/triton-inference-server/server/pull/6938

Currently, the dynamic batch scheduler only rejects timed-out requests when a payload slot is available and a new batch of requests is formed and submitted for inference. The payload slot availability is based on the completion of the previous batched inference request, which can take a long time depending on the model. Thus, the rejection of timed-out requests can be significantly delayed.

This change adds the ability for the dynamic batch scheduler to reject timed-out requests while waiting for the availability of a payload slot.

### DIFF
--- a/src/dynamic_batch_scheduler.cc
+++ b/src/dynamic_batch_scheduler.cc
@@ -461,28 +461,29 @@ DynamicBatchScheduler::WaitForPayloadSlotAvailable(
   lock->unlock();
 
   std::chrono::microseconds wait_timeout(wait_microseconds);
+  std::mutex slot_mu;
+  std::unique_lock<std::mutex> slot_lock(slot_mu);
 
-  auto slot_available_future = std::async(std::launch::async, [this]() {
-    return model_->Server()->GetRateLimiter()->PayloadSlotAvailable(
-        model_, model_instance_, queue_.SupportPrefetching());
-  });
-  while (slot_available_future.wait_for(wait_timeout) !=
-         std::future_status::ready) {
-    // Reject and release timed-out requests while waiting.
-    std::vector<std::deque<std::unique_ptr<InferenceRequest>>>
-        rejected_requests, cancelled_requests;
-    {
-      std::lock_guard<std::mutex> lock(mu_);
-      queue_.RejectTimeoutRequests();
-      queue_.ReleaseSkippedRequests(&rejected_requests, &cancelled_requests);
+  cv_.wait(slot_lock, [this, &wait_timeout]() {
+    auto slot_available_future = std::async(std::launch::async, [this]() {
+      return model_->Server()->GetRateLimiter()->PayloadSlotAvailable(
+          model_, model_instance_, queue_.SupportPrefetching());
+    });
+    while (slot_available_future.wait_for(wait_timeout) !=
+           std::future_status::ready) {
+      // Reject and release timed-out requests while waiting.
+      std::vector<std::deque<std::unique_ptr<InferenceRequest>>>
+          rejected_requests, cancelled_requests;
+      {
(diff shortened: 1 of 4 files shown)


src/scheduler_utils.cc
```diff
@@ -237,6 +237,32 @@ PriorityQueue::PolicyQueue::ApplyPolicy(
...
+  if (timeout_action_ != inference::ModelQueuePolicy::REJECT) {
+    return false;
+  }
+
+  size_t rejected_count = 0;
+  uint64_t now_nanoseconds =
+      std::chrono::duration_cast<std::chrono::nanoseconds>(
+          std::chrono::steady_clock::now().time_since_epoch())
+          .count();
+  for (size_t idx = 0; idx < queue_.size();) {
```
oandreeva-nv, line 252: maybe while loop with some end condition flag instead of a for loop? I was concerned/confused in the beginning regarding the running index, but then noticed you control it yourself

kthui: Sure, switched to while loop.

src/dynamic_batch_scheduler.cc
```diff
@@ -444,17 +443,62 @@ DynamicBatchScheduler::BatcherThread(const int nice)
...
+  const std::chrono::microseconds wait_timeout(wait_microseconds);
+  std::mutex slot_mu;
+  std::unique_lock<std::mutex> slot_lock(slot_mu);
+  bool slot_available = false;
+
+  while (!slot_available) {
+    slot_available = cv_.wait_for(
+        slot_lock, wait_timeout,
+        [this, &wait_timeout, &reject_and_release_timeout_requests]() {
+          auto slot_available_future = std::async(std::launch::async, [this]() {
```
oandreeva-nv, line 483: aha, this is the `async` you mentioned, which is calling a thread. Nice!

oandreeva-nv (dismissed):
Nice approach!

oandreeva-nv (approved):

src/dynamic_batch_scheduler.cc
```diff
@@ -359,18 +369,7 @@ DynamicBatchScheduler::BatcherThread(const int nice)
...
-          // outer lock to allow Enqueue threads above to make progress.
-          lock.unlock();
-          // Use slot lock to wait for the slot availability.
-          std::mutex slot_mu;
-          std::unique_lock<std::mutex> slot_lock(slot_mu);
-          cv_.wait(slot_lock, wait_for_slots);
-          // Recapture the outer most lock to keep making progress.
-          lock.lock();
-        }
+        WaitForPayloadSlotAvailable(&lock, default_wait_microseconds);
```
GuanLuo, line 372: default_wait_microseconds is a fixed value, wouldn't there still be delay of 0.5 seconds? Or that is acceptable for rejecting requests?

oandreeva-nv: I think 0.5s should be fine for now. What they were complaining about is timeout of 2s in the scenario, when a model executes for 10s, so timeout requests are waiting for 8 extra seconds to be returned.

We can always adjust it upon their feedback

kthui: Yes. I think we can have them try this version first, we can always reduce the delay interval if they ask. I have filed an enhancement ticket to remove the need of a manually set interval.

kthui: > wouldn't there still be delay of 0.5 seconds? Or that is acceptable for rejecting requests?

I think it is acceptable, since they mentioned "[some large model] could last for tens of seconds" and it becomes a concern, which 0.5 seconds is significantly less than tens of seconds.

PR #334, opened 2024-03-14
Exposing trace context to python backend

oandreeva-nv:
In this PR I 1) introduce context field to `InferenceTrace` 2) add 2 C APIs.

For this task I simply replace context with a new one. It may make sense in future to update APIs to extend and not replace  context. Open question is where should we do it: on the core side, or on the other side (frontend or backend) through GetContext -> localy update -> SetContext.

related PRs:
pbe: https://github.com/triton-inference-server/python_backend/pull/346
server: https://github.com/triton-inference-server/server/pull/6985

### DIFF
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -1,4 +1,4 @@
-// Copyright 2020-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2020-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -907,11 +907,11 @@ TRITONSERVER_InferenceTraceSetContext(
 /// Get TRITONSERVER_InferenceTrace context.
 ///
 /// \param trace The trace.
-/// \param context Returns the context associated with the trace.
+/// \param trace_context Returns the context associated with the trace.
 /// \return a TRITONSERVER_Error indicating success or failure.
 TRITONSERVER_DECLSPEC struct TRITONSERVER_Error*
 TRITONSERVER_InferenceTraceContext(
-    struct TRITONSERVER_InferenceTrace* trace, const char** context);
+    struct TRITONSERVER_InferenceTrace* trace, const char** trace_context);
 
 /// TRITONSERVER_InferenceRequest
 ///
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -91,7 +91,7 @@ struct TRITONSERVER_MetricFamily;
 ///   }
 ///
 #define TRITONSERVER_API_VERSION_MAJOR 1
-#define TRITONSERVER_API_VERSION_MINOR 28
+#define TRITONSERVER_API_VERSION_MINOR 29
 
 /// Get the TRITONBACKEND API version supported by the Triton shared
 /// library. This value can be compared against the
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -894,6 +894,27 @@ TRITONSERVER_InferenceTraceSpawnChildTrace(
     struct TRITONSERVER_InferenceTrace* trace,
     struct TRITONSERVER_InferenceTrace** child_trace);
 
+/// TODO: adjust
+///
+/// \param trace The trace.
+/// \param request_id Returns the version of the model associated
+/// with the trace.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC struct TRITONSERVER_Error*
+TRITONSERVER_InferenceTraceSetContext(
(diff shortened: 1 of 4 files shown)


include/triton/core/tritonserver.h
```diff
@@ -894,6 +894,25 @@ TRITONSERVER_InferenceTraceSpawnChildTrace(
...
+
+/// Get TRITONSERVER_InferenceTrace context.
+///
+/// \param trace The trace.
+/// \param context Returns the context associated with the trace.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC struct TRITONSERVER_Error*
+TRITONSERVER_InferenceTraceContext(
+    struct TRITONSERVER_InferenceTrace* trace, const char** context);
+
```
rmccorm4, line 915: Can we pick `trace_context` or `context` and use same in both places?

src/infer_trace.h
```diff
@@ -63,10 +63,12 @@ class InferenceTrace {
   const std::string& ModelName() const { return model_name_; }
   int64_t ModelVersion() const { return model_version_; }
   const std::string& RequestId() const { return request_id_; }
+  const std::string& Context() const { return context_; }
 
   void SetModelName(const std::string& n) { model_name_ = n; }
   void SetModelVersion(int64_t v) { model_version_ = v; }
   void SetRequestId(const std::string& request_id) { request_id_ = request_id; }
+  void SetContext(const std::string& context) { context_ = context; }
```
rmccorm4, line 71: I'm a little dubious of the user-provided `const char*` from `TRITONSERVER_InferenceTraceSetContext` being passed through to a `const std::string& context` here, but according to [this](https://stackoverflow.com/questions/67480842/does-stdstring-const-parameter-copies-the-passed-const-char) there will be a temp string constructed in between and then we're copying to the class member variable `context_` anyway so I think it's OK.

No action item here, just commenting a new thing I learned, and I was a little surprised the compiler is OK with this. I was surprised you can go from `const char*` to `const std::string&` because the `const char*`'s value can be changed, and then this "const ref string" would not be constant if referring to the original `const char*` provided.

rmccorm4 (commented):
LGTM, just the one nit: https://github.com/triton-inference-server/core/pull/334/files#r1525508414

include/triton/core/tritonserver.h
```diff
@@ -894,6 +894,25 @@ TRITONSERVER_InferenceTraceSpawnChildTrace(
...
+    struct TRITONSERVER_InferenceTrace* trace, const char* trace_context);
+
+
+/// Get TRITONSERVER_InferenceTrace context.
+///
+/// \param trace The trace.
+/// \param trace_context Returns the context associated with the trace.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC struct TRITONSERVER_Error*
+TRITONSERVER_InferenceTraceContext(
```
kthui, line 913: Just for my own understanding, if this function is called exclusively by the Python backend (at least on this change), should it be placed at the `tritonbackend.h`?

kthui:
Don't forget to update the [API version](https://github.com/triton-inference-server/core/blob/15c1e19b3dec953ba857d9c47f09589aad7f63db/include/triton/core/tritonserver.h#L94).

include/triton/core/tritonserver.h
oandreeva-nv: This can be also potentially be called from the front-end if there is a need to update context, i.e.
- context = TRITONSERVER_InferenceTraceContext
- Add value to existing (not `""`) context -> updated_context = context.addEntry(...)
- TRITONSERVER_InferenceTraceSetContext(triton_trace, updated_context)

oandreeva-nv: However, I do see your point and it might be reasonable to move `TRITONSERVER_InferenceTraceContext` to `tritonbackend.h` 

I don't think that a hypothetical scenario described above is valuable.

kthui: Thanks, that makes sense. It looks like we are keeping all the Trace and Metrics APIs in the `tritonserver.h` anyway.

include/triton/core/tritonserver.h
oandreeva-nv: updated

rmccorm4 (approved):

include/triton/core/tritonserver.h
```diff
@@ -894,6 +894,25 @@ TRITONSERVER_InferenceTraceSpawnChildTrace(
...
     struct TRITONSERVER_InferenceTrace** child_trace);
 
+/// Set TRITONSERVER_InferenceTrace context.
+///
+/// \param trace The trace.
+/// \param trace_context A new trace context to associate with the trace.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC struct TRITONSERVER_Error*
+TRITONSERVER_InferenceTraceSetContext(
+    struct TRITONSERVER_InferenceTrace* trace, const char* trace_context);
```
Tabrizian, line 904: Can we make the trace_context a void* with byte size? I think this allows us to be more flexible with the context definition if the future (e.g. if the context for some trace type should be just a binary blob).

PR #335, opened 2024-03-15
HTTP live connections on server shutdown

kthui:
Related PR: https://github.com/triton-inference-server/server/pull/6986

Add a new C-API to update the server exit timeout value and then stop the server.

### DIFF
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -2258,7 +2258,9 @@ TRITONSERVER_DECLSPEC struct TRITONSERVER_Error* TRITONSERVER_ServerDelete(
 TRITONSERVER_DECLSPEC struct TRITONSERVER_Error* TRITONSERVER_ServerStop(
     struct TRITONSERVER_Server* server);
 
-/// Set the exit timeout on the server object.
+/// Set the exit timeout on the server object. This value overrides the 
+/// value initially set through server options and provides a mechanism to 
+/// update the exit timeout while the serving is running.
 ///
 /// \param server The inference server object.
 /// \param timeout The exit timeout, in seconds.
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -91,7 +91,7 @@ struct TRITONSERVER_MetricFamily;
 ///   }
 ///
 #define TRITONSERVER_API_VERSION_MAJOR 1
-#define TRITONSERVER_API_VERSION_MINOR 29
+#define TRITONSERVER_API_VERSION_MINOR 30
 
 /// Get the TRITONBACKEND API version supported by the Triton shared
 /// library. This value can be compared against the
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -2239,14 +2239,13 @@ TRITONSERVER_DECLSPEC struct TRITONSERVER_Error* TRITONSERVER_ServerDelete(
 TRITONSERVER_DECLSPEC struct TRITONSERVER_Error* TRITONSERVER_ServerStop(
     struct TRITONSERVER_Server* server);
 
-/// Set the exit timeout on the server object, and then stop the server object.
-/// A server can't be restarted once it is stopped.
+/// Set the exit timeout on the server object.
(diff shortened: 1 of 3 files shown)


src/tritonserver.cc
```diff
@@ -2542,6 +2542,18 @@ TRITONSERVER_ServerStop(TRITONSERVER_Server* server)
   return nullptr;  // Success
 }
 
+TRITONAPI_DECLSPEC TRITONSERVER_Error*
+TRITONSERVER_ServerStopWithTimeout(
+    TRITONSERVER_Server* server, unsigned int timeout)
+{
+  tc::InferenceServer* lserver = reinterpret_cast<tc::InferenceServer*>(server);
+  if (lserver != nullptr) {
+    lserver->SetExitTimeoutSeconds(timeout);
```
nnshah1, line 2551: @kthui - I didn't realize that the internal object had a method `SetExitTimeoutSeconds` - this being the case I think also we could expose this method directly as well - that is add a new API
```
TRITONSERVER_ServerSetExitTimeoutSeconds()
```
I think in the design doc I thought we had something like "update" and that seemed odd - sorry for the thrash on this.

kthui: Sure, updated to expose the `SetExitTimeoutSeconds` method instead. The API name is
```
TRITONSERVER_ServerSetExitTimeout()
```
to align with the API name that sets the opaque option object when creating the server object
```
TRITONSERVER_ServerOptionsSetExitTimeout()
```

include/triton/core/tritonserver.h
```diff
@@ -2258,6 +2258,15 @@ TRITONSERVER_DECLSPEC struct TRITONSERVER_Error* TRITONSERVER_ServerDelete(
 TRITONSERVER_DECLSPEC struct TRITONSERVER_Error* TRITONSERVER_ServerStop(
     struct TRITONSERVER_Server* server);
 
+/// Set the exit timeout on the server object.
```
nnshah1, line 2261:
```suggestion
/// Set the exit timeout on the server object. This value will override the 
/// value initially set through server options and provides a mechanism to 
/// update the exit timeout while the serving is running.
```

nnshah1 (approved):
Slight tweak to description of API - to indicate that it overrides the original value if set.

src/tritonserver.cc
```diff
@@ -2573,13 +2573,12 @@ TRITONSERVER_ServerStop(TRITONSERVER_Server* server)
 }
 
 TRITONAPI_DECLSPEC TRITONSERVER_Error*
-TRITONSERVER_ServerStopWithTimeout(
+TRITONSERVER_ServerSetExitTimeout(
```
rmccorm4, line 2576: Why do we need both `TRITONSERVER_ServerOptionsSetExitTimeout` and `TRITONSERVER_ServerSetExitTimeout`?

kthui: `TRITONSERVER_ServerOptionsSetExitTimeout` is setting the timeout on an opaque object which is used when creating a new server object.
`TRITONSERVER_ServerSetExitTimeout` is to set the exit timeout on an exiting server object.

rmccorm4: So we're using this in server PR to logically separate the two parts of shutdown, but overall respect the --exit-timeout setting originally set?

ex: 30 sec timeout

- wait 30 secs for connections
- update core timeout to 0 seconds
- core shutdown exit immediately, no waiting

or

- wait 20 seconds connections
- update core timeout to 10 seconds for sum total of 30
- core shutdown

?

kthui: > overall respect the --exit-timeout setting originally set?

Yes, we respect the `--exit-timeout` command line option value overall.

The value is passed to the core with `TRITONSERVER_ServerOptionsSetExitTimeout` opaque option object when the core starts, but it is no longer accurate by the time of shutdown, because part of it can be consumed by the server frontend shutdown which happens ahead of the core shutdown, so it is always updated immediately with `TRITONSERVER_ServerSetExitTimeout` before the core shutdown.

nnshah1 (approved):
LGTM

PR #341, opened 2024-04-18
Fix double slot release after request cancellation

HennerM:
It looks like a regression got introduced with https://github.com/triton-inference-server/server/issues/273. 

Cancelled requets release a slot in the sequence batcher, but the very same request still get reaped after idle which leads to the same slot (now potentially already taken by another sequence) to be released again. This can have severe consequences when using the implicit state management, because the state is associated with the slot.

This is the root cause for the problem described in https://github.com/triton-inference-server/server/issues/7117. 

I have tested this change against a very simple example with the gRPC Python client and a simple sequence model that has max_candidate_sequences 2. I sent two requests, one that is currently executed and another that is queued up in the sequence batcher, now I cancel request two and wait for max_sequence_idle_microseconds. Now I send two new sequences and see in the logs that they both get assigned slot 0

### DIFF
--- a/src/sequence_batch_scheduler/sequence_batch_scheduler.cc
+++ b/src/sequence_batch_scheduler/sequence_batch_scheduler.cc
@@ -884,6 +884,29 @@ SequenceBatchScheduler::ReleaseSequenceSlot(
 {
   std::unique_lock<std::mutex> lock(mu_);
 
+  // If we are releasing the slot for a cancelled sequence,
+  // we have to clean up the sequence
+  // otherwise the reaper will try to clean it up again.
+  if (!requests->empty() && requests->front()->IsCancelled()) {
+    // clean-up the cancelled sequences
+    // Normal sequences will already be
+    const InferenceRequest::SequenceId& old_correlation_id =
+        requests->front()->CorrelationId();
+    LOG_VERBOSE(1) << "Releasing canceled "
+                      "CORRID "
+                   << old_correlation_id;
+
+    if (sequence_to_batcherseqslot_map_.find(old_correlation_id) !=
+        sequence_to_batcherseqslot_map_.end()) {
+      sequence_to_batcherseqslot_map_.erase(old_correlation_id);
+    }
+    // remove this correlation ID from expiration checks
+    if (correlation_id_timestamps_.find(old_correlation_id) !=
+        correlation_id_timestamps_.end()) {
+      correlation_id_timestamps_.erase(old_correlation_id);
+    }
+  }
+
   // If there are any remaining requests on the releasing sequence slot, those
   // requests will be cancelled.
   MarkRequestsCancelled(requests);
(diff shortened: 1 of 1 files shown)


src/sequence_batch_scheduler/sequence_batch_scheduler.cc
```diff
@@ -884,6 +884,29 @@ SequenceBatchScheduler::ReleaseSequenceSlot(
...
+                   << old_correlation_id;
+
+    if (sequence_to_batcherseqslot_map_.find(old_correlation_id) !=
+        sequence_to_batcherseqslot_map_.end()) {
+      sequence_to_batcherseqslot_map_.erase(old_correlation_id);
+    }
+    // remove this correlation ID from expiration checks
+    if (correlation_id_timestamps_.find(old_correlation_id) !=
+        correlation_id_timestamps_.end()) {
+      correlation_id_timestamps_.end();
```
rakeshv247, line 906: should this be `correlation_id_timestamps_.erase();`

HennerM: yes 100%

rmccorm4:
Thanks for the submission @HennerM ! We're looking into this PR and the underlying root causes or edge cases throughout the Sequence Batch Scheduler.

src/sequence_batch_scheduler/sequence_batch_scheduler.cc
```diff
@@ -884,6 +884,29 @@ SequenceBatchScheduler::ReleaseSequenceSlot(
...
+    // remove this correlation ID from expiration checks
+    if (correlation_id_timestamps_.find(old_correlation_id) !=
+        correlation_id_timestamps_.end()) {
+      correlation_id_timestamps_.erase(old_correlation_id);
+    }
+  }
+
   // If there are any remaining requests on the releasing sequence slot, those
   // requests will be cancelled.
   MarkRequestsCancelled(requests);
```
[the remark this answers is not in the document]
HennerM: Thanks for the review @kthui 
Your suggestions make sense, I wasn't entirely sure that erase behaves that way, good to know, it's a lot simpler now.
I basically took in your suggestion as you said, I still left in the log message though because in our debugging for this issue we found the logging invaluable and since this situation shouldn't occurr too often I don't think it adds much noise to the logs.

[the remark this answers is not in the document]
kthui: I think the logs are good and give us some good insight into some critical stages of the scheduler when running.

kthui (approved):
Thanks for the contribution @HennerM It is awesome we can get this bug fixed!

src/sequence_batch_scheduler/sequence_batch_scheduler.cc
```diff
@@ -884,6 +884,24 @@ SequenceBatchScheduler::ReleaseSequenceSlot(
 {
   std::unique_lock<std::mutex> lock(mu_);
 
+  // If we are releasing the slot for a cancelled sequence,
+  // we have to clean up the sequence
+  // otherwise the reaper will try to clean it up again.
+  if (!requests->empty() && requests->front()) {
+    const InferenceRequest::SequenceId& corr_id =
+        requests->front()->CorrelationId();
```
Tabrizian, line 892: @kthui is there a possiblity for requests other than the front request to be cancelled?

rmccorm4: I believe cancelling any request in the sequence should cancel the entire sequence. The `requests->front()` check is just to make sure that the request object is not null (a null request is inserted by reaper thread to signal a timed out sequence). Otherwise I'll let @kthui comment if you're asking something else.

tanmayv25 (approved):

PR #343, opened 2024-04-19
Add model namespacing into metrics

kthui:
Related PR: https://github.com/triton-inference-server/server/pull/7141

Include "namespace" label into metrics if the server is launched with "--model-namespacing=true". The "namespace" label will not be added if the server is launched without "--model-namespacing=true".

closes https://github.com/triton-inference-server/server/issues/6138

### DIFF
--- a/src/ensemble_scheduler/ensemble_scheduler.cc
+++ b/src/ensemble_scheduler/ensemble_scheduler.cc
@@ -1373,8 +1373,9 @@ EnsembleScheduler::EnsembleScheduler(
   if (Metrics::Enabled()) {
     // Ensemble scheduler doesn't currently support response cache at top level.
     MetricModelReporter::Create(
-        model_id, 1, METRIC_REPORTER_ID_CPU, false /* response_cache_enabled */,
-        config.metric_tags(), &metric_reporter_);
+        model_id, 1 /* model_version */, METRIC_REPORTER_ID_CPU,
+        false /* response_cache_enabled */, config.metric_tags(),
+        &metric_reporter_);
   }
 #endif  // TRITON_ENABLE_METRICS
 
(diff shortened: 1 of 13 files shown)


src/ensemble_scheduler/ensemble_scheduler.cc
```diff
@@ -1371,9 +1373,8 @@ EnsembleScheduler::EnsembleScheduler(
   if (Metrics::Enabled()) {
     // Ensemble scheduler doesn't currently support response cache at top level.
     MetricModelReporter::Create(
-        config.name(), 1, METRIC_REPORTER_ID_CPU,
-        false /* response_cache_enabled */, config.metric_tags(),
-        &metric_reporter_);
+        model_id, 1, METRIC_REPORTER_ID_CPU, false /* response_cache_enabled */,
```
oandreeva-nv, line 1376:
```suggestion
        model_id, 1 /* model_version */, METRIC_REPORTER_ID_CPU, false /* response_cache_enabled */,
```
I know, that it is not your change, but I would like to elaborate this magic number =)

oandreeva-nv: @rmccorm4 could you please clarify why we hardcode `model_version`?

rmccorm4: Looks like that goes way back to [May 2020](https://github.com/triton-inference-server/server/commit/42fc28f5462ef2c4dd3b4c26d5a13ab315556c35#diff-d2cd2b262fd3de620e660fd32738363c38759afbfb3cd3e6987fcf10b462060d).

If I were to guess, it's because the ensemble model has no `TritonModel*` object associated with it to query a version, and I don't think the model config protobuf object has a version field to query either (like the `config.name()` it uses for model name). There may be a way around it to get some programattic now :shrug: 

Don't think I've ever seen someone raise an issue about this, maybe people don't mess with ensemble versions :sweat_smile:

kthui: > I know, that it is not your change, but I would like to elaborate this magic number =)

Sure, just adding a comment: [Comment on model version number](https://github.com/triton-inference-server/core/pull/343/commits/b9ec4c5b4dc273f42464e85926b8cb0bc390aca1)

oandreeva-nv (approved):
LGTM

PR #346, opened 2024-04-23
Add metrics support for rejected requests

rmccorm4:


src/failure_reasons.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+enum class FailureReason {
+    REJECTED,
+    CANCELED,
+    BACKEND,
+    TIMEOUT,
```
rmccorm4, line 7: What's the difference or scenarios between REJECTED and TIMEOUT?

src/infer_request.cc
```diff
@@ -442,6 +450,10 @@ InferenceRequest::RespondIfError(
           std::move(response), TRITONSERVER_RESPONSE_COMPLETE_FINAL, status),
       (request->LogRequest() + "failed to send error response").c_str());
 
+  // Report error with type.
+  FailureReason reason = stringToFailureReason(status.Message());  // Convert error message to reason
```
rmccorm4, line 454: I don't think this `status` will contain the expected string for conversion here

rmccorm4: We can probably pass the `FailureReason` directory to the call to RespondIfError depending on the part of the code we're calling from

src/infer_request.cc
```diff
@@ -1272,20 +1285,31 @@ InferenceRequest::ReportStatistics(
...
   }
 }
 
+void 
+InferenceRequest::ReportStatistics(
+  MetricModelReporter* metric_reporter, bool success,
+  const uint64_t compute_start_ns, const uint64_t compute_input_end_ns,
+  const uint64_t compute_output_start_ns, const uint64_t compute_end_ns) {
+  // Call the updated ReportStatistics with a default error_type
+  this->ReportStatistics(metric_reporter, success, compute_start_ns, compute_input_end_ns,
```
rmccorm4, line 1302: Don't need `this`
```suggestion
  ReportStatistics(metric_reporter, success, compute_start_ns, compute_input_end_ns, 
```

src/infer_stats.cc
```diff
@@ -48,7 +60,8 @@ InferenceStatsAggregator::UpdateFailure(
 
 #ifdef TRITON_ENABLE_METRICS
   if (metric_reporter != nullptr) {
-    metric_reporter->IncrementCounter("inf_failure", 1);
+    std::string reason_str = failureReasonToString(reason);
+    metric_reporter->IncrementCounter("inf_failure_" + reason_str, 1);
```
rmccorm4, line 64: Didn't we agree to use prometheus `labels` for these metrics instead of embedding the reason within the metric name?

src/metric_model_reporter.cc
```diff
@@ -341,9 +358,9 @@ MetricModelReporter::IncrementCounter(const std::string& name, double value)
     // No counter metric exists with this name
     return;
   }
-
+  
   auto counter = iter->second;
-  if (!counter) {
+  if (!counter){
```
rmccorm4, line 363: You should setup the pre-commit hooks like `pre-commit install` and `pre-commit run` to auto-run the linter/formatters when you commit the code.

src/failure_reasons.h
rmccorm4: Let's just go with rejected for now

itskevinwang: Keep rejected, remove timeout since timeout fits underneath rejected. There are multiple ways something can be rejected.

src/infer_request.cc
itskevinwang: Look into how many times RespondIfError is called, do some research here.

src/metric_model_reporter.cc
```diff
@@ -183,6 +184,14 @@ MetricModelReporter::~MetricModelReporter()
   }
 }
 
+std::map<FailureReason, std::string> failure_reasons_map = {
+    {FailureReason::REJECTED, "REJECTED"},
+    {FailureReason::CANCELED, "CANCELED"},
+    {FailureReason::BACKEND, "BACKEND"},
+    {FailureReason::TIMEOUT, "TIMEOUT"},
+    {FailureReason::OTHER, "OTHER"}
+};
```
rmccorm4, line 193: We should probably make this `static const` somewhere, not sure where yet. Let me know if you have any ideas. I'm thinking either `constants.h` or maybe a member variable of the MetricModelReporter

src/infer_request.cc
```diff
@@ -420,6 +420,14 @@ InferenceRequest::Run(std::unique_ptr<InferenceRequest>& request)
   return status;
 }
 
+FailureReason stringToFailureReason(const std::string& error_type) {
+  if (error_type == "REJECTED") return FailureReason::REJECTED;
+  if (error_type == "CANCELED") return FailureReason::CANCELED;
+  if (error_type == "BACKEND") return FailureReason::BACKEND;
+  if (error_type == "TIMEOUT") return FailureReason::TIMEOUT;
+  return FailureReason::OTHER;
+}
```
rmccorm4, line 429: I don't have a link to back it up right now, but I think we should generally avoid these c-style 1-line if statements, and instead have the brackets `{}`

src/infer_request.cc
```diff
@@ -442,6 +455,8 @@ InferenceRequest::RespondIfError(
           std::move(response), TRITONSERVER_RESPONSE_COMPLETE_FINAL, status),
       (request->LogRequest() + "failed to send error response").c_str());
 
+  request->ReportStatistics(request->model_raw_->MetricReporter().get(), false, 0, 0, 0, 0, reason);
```
rmccorm4, line 458: Add inline comments for what these hard-coded values are:
```cpp
request->ReportStatistics(..., false /* success */, 0 /* compute_start_ns */, ...)
```

src/infer_request.cc
```diff
@@ -1272,20 +1274,31 @@ InferenceRequest::ReportStatistics(
...
 
+void 
+InferenceRequest::ReportStatistics(
+  MetricModelReporter* metric_reporter, bool success,
+  const uint64_t compute_start_ns, const uint64_t compute_input_end_ns,
+  const uint64_t compute_output_start_ns, const uint64_t compute_end_ns) {
+  // Call the updated ReportStatistics with a default error_type
+  ReportStatistics(metric_reporter, success, compute_start_ns, compute_input_end_ns, 
+                         compute_output_start_ns, compute_end_ns, FailureReason::BACKEND);
+}
```
rmccorm4, line 1293: I think this overload is handled by your change to `TRITONBACKEND_ModelInstanceReportStatistics` to always pass `FailureReason::BACKEND` - is this one needed?

itskevinwang: Correct, removed the overload.

src/infer_request.cc
itskevinwang: Add comments.

src/infer_request.cc
itskevinwang: Corrected.

src/infer_request.cc
itskevinwang: Corrected.

src/failure_reasons.h
itskevinwang: Corrected.

src/infer_request.cc
```diff
@@ -442,6 +455,16 @@ InferenceRequest::RespondIfError(
...
 
+  request->ReportStatistics(
+      request->model_raw_->MetricReporter().get(),
+      false, // success: Indicates the operation did not succeed
+      0, // compute_start_ns: Start time of the compute operation, 0 because there was an error before computation
+      0, // compute_input_end_ns: End time of input processing, 0 because there was an error before input processing
+      0, // compute_output_start_ns: Start time of output processing, 0 because there was an error before output processing
+      0, // compute_end_ns: End time of the compute operation, 0 because there was an error before computation ended
+      reason // reason: The specific reason for the failure
+  );
```
rmccorm4, line 466: No need for full descriptions here, just mentioning which vars they refer to. ex: https://github.com/triton-inference-server/core/blob/34d2650598ba394c75c368de99d39462d596f79a/src/dynamic_batch_scheduler.cc#L262

```suggestion
  // compute times set to zero because there was an error
  request->ReportStatistics(
      request->model_raw_->MetricReporter().get(),
      false /* success */,
      0 /* compute_start_ns */,
      0 /* compute_input_end_ns */, 
      0 /* compute_output_start_ns */,
      0 /* compute_end_ns */,
      reason
  );
```

rmccorm4: Also please run the pre-commit hooks so all the style stuff will automatically be linted

src/backend_model_instance.cc
```diff
@@ -558,7 +559,7 @@ TritonModelInstance::PrepareRequestsOrRespond(
   // If any errors occurred, respond with error for each request.
   if (!status.IsOk()) {
     for (auto& r : requests) {
-      InferenceRequest::RespondIfError(r, status, true /* release_requests */);
+      InferenceRequest::RespondIfError(r, status, true /* release_requests */, FailureReason::BACKEND);
```
rmccorm4, line 562: Can you add a test to confirm whether or not this `RespondIfError` call and the `TRITONBACKEND_ModelInstanceReportStatistics(..., success=false, ...)` is double counting the same error?

src/metric_model_reporter.cc
```diff
@@ -183,6 +184,13 @@ MetricModelReporter::~MetricModelReporter()
   }
 }
 
+std::map<FailureReason, std::string> failure_reasons_map = {
```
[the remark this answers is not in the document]
Tabrizian: Can we make it a static variable of the `MetricModelReporter` class?

src/infer_stats.cc
```diff
@@ -36,10 +36,21 @@ namespace triton { namespace core {
 
 #ifdef TRITON_ENABLE_STATS
 
+// This function converts FailureReason enum values to std::string
+std::string failureReasonToString(FailureReason reason) {
+  switch (reason) {
+    case FailureReason::REJECTED: return "REJECTED";
```
Tabrizian, line 42: instead of introducing a separate data structure can we use the TRITON error types?

src/infer_request.cc
```diff
@@ -420,10 +420,23 @@ InferenceRequest::Run(std::unique_ptr<InferenceRequest>& request)
   return status;
 }
 
+FailureReason stringToFailureReason(const std::string& error_type) {
+    if (error_type == "REJECTED") {
```
rmccorm4, line 424: @Tabrizian what do you think about "Rejected" vs "Timeout" for this category of failures?
```
nv_inference_request_failure{model="llama-3-8b-instruct",version="1",reason="TIMEOUT"} 7
```
vs.
```
nv_inference_request_failure{model="llama-3-8b-instruct",version="1",reason="REJECTED"} 7
```

Originally I opted for "REJECTED" because of:
```
  enum TimeoutAction {
    REJECT = 0;
    DELAY = 1;
}
```
and only the REJECTED timeout would be a "failure", but TIMEOUT may be more intuitive.

src/infer_stats.cc
rmccorm4: note to @itskevinwang (who came by to my desk with some questions), let's focus on the following first:
- serve a model that loads successfully
- make sure the new failure metrics appear in output of `curl localhost:8002/metrics`
- have some tests that would test the expected logic - it's OK if the tests themselves fail (ex: expected failure count > 1, but got zero, or something like that)

Once you have server, model, and tests that can run together, we can iterate on the feedback more easily.

src/infer_request.cc
Tabrizian: Is it possible to merge the [error types](https://github.com/triton-inference-server/common/blob/00b3a71519e32e3bc954e9f0d067e155ef8f1a6c/include/triton/common/error.h#L39-L49) with failure reason data structure? I think ideally we want the backend to be able to report these categories as well. I may not have to full context on this feature so feel free to correct my understanding. I slightly prefer timeout since it could be used for connection timeout as well.

PR #347, opened 2024-05-01
Change the default value for strict model config for parity with server

Tabrizian:
Auto-complete is enabled by default on the server side while it was disabled on the Python API because of this default value.

### DIFF
--- a/python/tritonserver/_api/_server.py
+++ b/python/tritonserver/_api/_server.py
@@ -137,7 +137,7 @@ class Options:
         List of models to load at startup. Only relevant with ModelControlMode.EXPLICIT.
         See :c:func:`TRITONSERVER_ServerOptionsSetStartupModel`
 
-    strict_model_config : bool, default True
+    strict_model_config : bool, default False
         Enable or disable strict model configuration.
         See :c:func:`TRITONSERVER_ServerOptionsSetStrictModelConfig`
 
@@ -275,7 +275,7 @@ class Options:
     server_id: str = "triton"
     model_control_mode: ModelControlMode = ModelControlMode.NONE
     startup_models: list[str] = field(default_factory=list[str])
-    strict_model_config: bool = True
+    strict_model_config: bool = False
 
     rate_limiter_mode: RateLimitMode = RateLimitMode.OFF
     rate_limiter_resources: list[RateLimiterResource] = field(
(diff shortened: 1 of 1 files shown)


rmccorm4 (commented):
LGTM - I was looking for this change too, but will leave approval to Neelay

nnshah1:
I'm a little conflicted on the best approach here. During the initial implementation the goal was to be consistent with the C in process api and honor the defaults as set in the code. This value for strict model config was taken from:

https://github.com/triton-inference-server/core/blob/5fc928af68082896d49e385c49c478e76088532d/src/server.cc#L104 

That being said we can adopt the defaults as set by the server application as an alternative - however we also are now recommending users to set strict_model_config to True unless needed for their use case (opt-in instead of opt-out):

https://docs.nvidia.com/deeplearning/triton-inference-server/user-guide/docs/customization_guide/deploy.html#disable-auto-complete-config-default-enabled 

let's discuss as part of python in process api productization epic (i.e. revisit default values)

nnshah1 (changes requested):
pending discussion

Tabrizian:
> however we also are now recommending users to set strict_model_config to True unless needed for their use case (opt-in instead of opt-out):

https://docs.nvidia.com/deeplearning/triton-inference-server/user-guide/docs/customization_guide/deploy.html#disable-auto-complete-config-default-enabled

Can you elaborate about the security risk of auto-complete?

nnshah1:
> > however we also are now recommending users to set strict_model_config to True unless needed for their use case (opt-in instead of opt-out):
> 
> https://docs.nvidia.com/deeplearning/triton-inference-server/user-guide/docs/customization_guide/deploy.html#disable-auto-complete-config-default-enabled
> 
> Can you elaborate about the security risk of auto-complete?

auto complete can more easily lead to non obvious results - since we offer the option to disable and it is not a necessary requirement for many use cases - the default for the server application likely should be to disable as basic principle of least priviledge / least feature - and make the team deploying the application responsible to enabling it.

nnshah1 (commented):
In offline discussion we have added a separate task to review the default options for the in process APIs (so they all are consistent) and for the server application seperately. Unblocking this request in the mean time with the understanding that the default may change in the future after review.

nnshah1 (approved):

rmccorm4 (approved):
My $0.02 for the future discussion, maybe the python "high-level" wrapper can be a little more opinionated than the 1:1 bindings below it for ergonomics where we deem fit. 

As another data point for "prior art", we have a C++ high level wrapper around the C API, and it looks like it also has strict model config as `false` by default:
- https://github.com/triton-inference-server/developer_tools/blob/3d71fca07dfa7c68ef574566c809775bcf0312b6/server/src/server_wrapper.cc#L800
- https://github.com/triton-inference-server/developer_tools/blob/3d71fca07dfa7c68ef574566c809775bcf0312b6/server/src/server_wrapper.cc#L1539-L1540

Maybe the productization discussion could consider making the "high level wrappers" match rather than making 1:1 C-API match.

Tabrizian:
@nnshah1 @rmccorm4 I think we're all mostly on the same page. I agree with being more permissive in the API and restrictive in the `tritonserver` binary. Ideally, we need to ensure similar user-experience across all these different APIs. Thanks for the feedback and discussions.

PR #348, opened 2024-05-06
Multiple Model Configurations

yinggeh:
Add `--mode-config-name` option when starting Triton server. Allow users to select custom configurations other than default config.pbtxt.

server: https://github.com/triton-inference-server/server/pull/7185

### DIFF
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -91,7 +91,7 @@ struct TRITONSERVER_MetricFamily;
 ///   }
 ///
 #define TRITONSERVER_API_VERSION_MAJOR 1
-#define TRITONSERVER_API_VERSION_MINOR 30
+#define TRITONSERVER_API_VERSION_MINOR 31
 
 /// Get the TRITONBACKEND API version supported by the Triton shared
 /// library. This value can be compared against the
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -1828,6 +1828,16 @@ TRITONSERVER_DECLSPEC struct TRITONSERVER_Error*
 TRITONSERVER_ServerOptionsSetStrictModelConfig(
     struct TRITONSERVER_ServerOptions* options, bool strict);
 
+/// Set the custom model configuration name to load for all models.
+/// Fall back to default config file if empty.
+///
+/// \param options The server options object.
+/// \param config_name The name of the config file to load for all models.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC struct TRITONSERVER_Error*
+TRITONSERVER_ServerOptionsSetModelConfigName(
+    struct TRITONSERVER_ServerOptions* options, const char* model_config_name);
+
 /// Set the rate limit mode in a server options.
 ///
 ///   TRITONSERVER_RATE_LIMIT_EXEC_COUNT: The rate limiting prioritizes the
--- a/src/model_repository_manager/model_lifecycle.cc
+++ b/src/model_repository_manager/model_lifecycle.cc
@@ -92,8 +92,10 @@ VersionsToLoad(
   RETURN_IF_ERROR(GetDirectorySubdirs(model_path, &subdirs));
   std::set<int64_t, std::greater<int64_t>> existing_versions;
   for (const auto& subdir : subdirs) {
-    if (subdir == kWarmupDataFolder || subdir == kInitialStateFolder ||
-        subdir == kModelConfigFolder) {
+    static const std::vector skip_dirs{
+        kWarmupDataFolder, kInitialStateFolder, kModelConfigFolder};
+    if (std::find(skip_dirs.begin(), skip_dirs.end(), subdir) !=
+        skip_dirs.end()) {
       continue;
     }
     if ((subdir.length() > 1) && (subdir.front() == '0')) {
--- a/src/model_repository_manager/model_lifecycle.cc
+++ b/src/model_repository_manager/model_lifecycle.cc
@@ -92,7 +92,8 @@ VersionsToLoad(
   RETURN_IF_ERROR(GetDirectorySubdirs(model_path, &subdirs));
   std::set<int64_t, std::greater<int64_t>> existing_versions;
   for (const auto& subdir : subdirs) {
-    if (subdir == kWarmupDataFolder || subdir == kInitialStateFolder) {
+    if (subdir == kWarmupDataFolder || subdir == kInitialStateFolder ||
+        subdir == kModelConfigFolder) {
       continue;
     }
     if ((subdir.length() > 1) && (subdir.front() == '0')) {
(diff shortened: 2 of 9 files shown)


include/triton/core/tritonserver.h
```diff
@@ -1828,6 +1828,16 @@ TRITONSERVER_DECLSPEC struct TRITONSERVER_Error*
...
     struct TRITONSERVER_ServerOptions* options, bool strict);
 
+/// Set the custom model configuration name to load for all models.
+/// Fall back to default config file if empty.
+///
+/// \param options The server options object.
+/// \param config_name The name of the config file to load for all models.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC struct TRITONSERVER_Error*
+TRITONSERVER_ServerOptionsSetModelConfigName(
```
rmccorm4, line 1838: Don't forget to update API version: https://github.com/triton-inference-server/core/blob/35555724612df29923007f1da45d2a58f928206c/include/triton/core/tritonserver.h#L93-L94

src/model_repository_manager/model_lifecycle.cc
```diff
@@ -92,7 +92,8 @@ VersionsToLoad(
   RETURN_IF_ERROR(GetDirectorySubdirs(model_path, &subdirs));
   std::set<int64_t, std::greater<int64_t>> existing_versions;
   for (const auto& subdir : subdirs) {
-    if (subdir == kWarmupDataFolder || subdir == kInitialStateFolder) {
+    if (subdir == kWarmupDataFolder || subdir == kInitialStateFolder ||
+        subdir == kModelConfigFolder) {
```
GuanLuo, line 96: Should just compile a list of folders to skip and for-loop over it.

src/model_repository_manager/model_repository_manager.cc
```diff
@@ -218,8 +219,7 @@ CreateAgentModelListWithLoadAction(
       std::unique_ptr<TritonRepoAgentModel> agent_model;
       if (lagent_model_list->Size() != 0) {
         lagent_model_list->Back()->Location(&artifact_type, &location);
-        const auto config_path = JoinPath({location, kModelConfigPbTxt});
-        if (!ReadTextProto(config_path, &model_config).IsOk()) {
+        if (!ReadTextProto(orignial_model_config_path, &model_config).IsOk()) {
```
GuanLuo, line 222: I don't understand this change. The intention is that after creating a repo agent, it may have modified the files / change model repo location, therefore the config should be reloaded from the location reported by the repo agent.

After the change the config from original model repo will always be loaded?

src/model_repository_manager/model_repository_manager.cc
```diff
@@ -1401,17 +1434,21 @@ ModelRepositoryManager::InitializeModelInfo(
     // the override while the local files may still be unchanged.
     linfo->mtime_nsec_ = std::make_pair(0, 0);
     linfo->model_path_ = location;
+    linfo->model_config_path_ = JoinPath({location, kModelConfigPbTxt});
     linfo->agent_model_list_.reset(new TritonRepoAgentModelList());
     linfo->agent_model_list_->AddAgentModel(std::move(localize_agent_model));
   } else {
+    linfo->model_config_path_ = GetModelConfigFullPath(linfo->model_path_);
```
GuanLuo, line 1441: Will this get updated periodically? Say if a `model_config_path_` is set and the file get deleted by the user. Will the usage of `model_config_path_` just always pointing to invalid path instead of falling back to some config based on fall back policy?

src/model_repository_manager/model_repository_manager.cc
yinggeh: Nice catch. I misunderstood the previous line where `location` can be updated.

src/model_repository_manager/model_repository_manager.cc
yinggeh: If `model-management-mode` is set to none, the configuration only loaded once per my understanding.

The `InitializeModelInfo` function is invoked by function `Poll`, which is invoked by `LoadUnloadModels`. So if If `model-management-mode` is set to explicit or poll, the configuration location will update. See tests [qa/L0_lifecycle/test.sh](https://github.com/triton-inference-server/server/pull/7185/files#diff-8ad9d754d4d301c75f1892863f7a7ad02287a882d18bff7460b14faac9cc06a6) and [qa/L0_lifecycle/lifecycle_test.py](https://github.com/triton-inference-server/server/pull/7185/files#diff-27572619e19393af9b72b568ffb9d6b2c4932944aa3141a492176636acb8e508).

src/model_repository_manager/model_lifecycle.cc
yinggeh: Can you elaborate "compile a list of folders", like create a constant vector in `src/constants.h`?

GuanLuo: Something like...
```
skip_folders = [kWarmupDataFolder, kInitialStateFolder, kModelConfigFolder]
if subdir in skip_folders
```

yinggeh: Fixed

src/model_repository_manager/model_repository_manager.cc
yinggeh: Fixed

src/model_repository_manager/model_repository_manager.cc
```diff
@@ -1401,17 +1437,22 @@ ModelRepositoryManager::InitializeModelInfo(
     // the override while the local files may still be unchanged.
     linfo->mtime_nsec_ = std::make_pair(0, 0);
     linfo->model_path_ = location;
+    linfo->model_config_path_ = JoinPath({location, kModelConfigPbTxt});
```
GuanLuo, line 1440: Should be like below as well..
```
linfo->model_config_path_ =
        GetModelConfigFullPath(linfo->model_path_, model_config_name_);`
```

GuanLuo: Can you add a test case where `configs/<some_config_name>.pbtxt` is specified via loading with file override?

yinggeh: Correct me if I am wrong. In this case, `LocalizeRepoAgent` in line 1417 will create the model config file named `kModelConfigPbTxt` aka "config.pbtxt" in a tmp directory. Therefore, the config file path is static.

GuanLuo (approved):
Approving but do check the new comments for concerns on the handling in model file override scenario.

src/model_repository_manager/model_repository_manager.cc
yinggeh: Looks safe to me. See error below when not specifying config for model file override.
`E0516 17:57:40.808915 923786 model_repository_manager.cc:1371] Poll failed for model directory 'onnx_float32_float32_float32': Load parameter 'config' must be specified for model file override`

yinggeh:
Add --mode-config-name option when starting Triton server. Allow users to select custom configurations other than default config.pbtxt.

include/triton/core/tritonserver.h
rmccorm4, line 1838: TODO: Need to propagate changes to python bindings and C++ bindings, etc.

rmccorm4: Filed DLIS-6713

PR #349, opened 2024-05-07
Updated parameter handling

nnshah1:
1) add support for double 
2) add support for mapping numpy.bool_ type to tritonserver data type
3) changed dlpack dtype for bool to kDLBool, 8 (8bits) from KDLBool, 1 (1bits)  

Discovered in sending sampling parameters to vLLM backend.


python/tritonserver/_api/_request.py
```diff
@@ -167,6 +167,10 @@ def _set_parameters(self, request):
                 request.set_int_parameter(key, value)
             elif isinstance(value, bool):
                 request.set_bool_parameter(key, value)
+            elif isinstance(value, float):
+                request.set_double_parameter(key, value)
```
rmccorm4, line 171: Update type hint on line 132
```
    parameters: dict[str, str | int | bool | float] = field(default_factory=dict)
```

rmccorm4: and same docstring on line 78 for parameters to include `float` in `Union`

python/tritonserver/_api/_datautils.py
```diff
@@ -176,6 +176,7 @@ def __getitem__(self, key):
     "Triton server data type",
     {
         bool: DataType.BOOL,
+        numpy.bool_: DataType.BOOL,
```
rmccorm4, line 179: Would be nice to start iteratively adding some unit tests for some of these, would it be an easy addition or copy of [test_request_create](https://github.com/triton-inference-server/core/blob/06b8f6e98fa90f93c02d899886564a6e88f2ae02/python/test/test_api.py#L84) by passing in a `parameters` dict with a float?

python/tritonserver/_api/_request.py
```diff
@@ -167,6 +167,10 @@ def _set_parameters(self, request):
...
+            elif value is None:
```
Tabrizian, line 172: Can you elaborate more about this part and why it was required for OpenAI chat completions?

nnshah1: double was to allow for setting temperature in vllm sampling parameters (and should be added in anycase for completeness)

on the None its more related to requests specified in open api specs like:

 https://github.com/triton-inference-server/tutorials/blob/0774e8454f1d6a9ed995e005d99a24004682a1c8/Triton_Inference_Server_Python_API/examples/fastapi/fastapi-codegen/openai_protocol_types.py#L64 

The pydantic type fills in None for values that aren't present so this was mainly for convience of being able to pass through the request parameters.

https://github.com/triton-inference-server/tutorials/blob/0774e8454f1d6a9ed995e005d99a24004682a1c8/Triton_Inference_Server_Python_API/examples/fastapi/fastapi-codegen/main.py#L153

I think it's easy to workaround - but thought its a nice convience to ignore parameters that are None - I'm a little ambivalent to it now.

Tabrizian: Thanks for explaining. Could we prune these values before passing as request parameters? I also don't have strong opinions. Might be worth mentioning it in the docstrings so that the behaviour is clear to the user.

nnshah1: agree we can prune for now - let me take a stab at that and then revert this - on second thought probably better to error cause it may hide some error

python/tritonserver/_api/_request.py
```diff
@@ -167,6 +167,10 @@ def _set_parameters(self, request):
                 request.set_int_parameter(key, value)
             elif isinstance(value, bool):
                 request.set_bool_parameter(key, value)
+            elif isinstance(value, float):
```
GuanLuo, line 170: Specify a type map?
```
_set_param_func = {str : Request.set_string_parameter, ...}
...
for key, value in ...:
    _set_param_func[type(value)](request, key, value)
```

python/test/test_api.py
```diff
@@ -450,12 +450,93 @@ def test_basic_inference(self):
             },
         )
 
+        inputs = {
+            "fp16_input": numpy.random.rand(1, 100).astype(dtype=numpy.float16),
+            "bool_input": numpy.random.rand(1, 100).astype(dtype=numpy.bool_),
```
GuanLuo, line 455: The model already has these inputs and they weren't connected previously?

python/test/test_api.py
```diff
@@ -450,12 +450,93 @@ def test_basic_inference(self):
...
+        }
+
+        for response in server.model("test").infer(
+            inputs=inputs,
+            output_memory_type="cpu",
+            raise_on_error=True,
+        ):
+            for name, input_ in inputs.items():
```
Tabrizian, line 463:
```suggestion
            for name, input_item in inputs.items():
```

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -1159,6 +1162,11 @@ class PyInferenceRequest
     ThrowIfError(TRITONSERVER_InferenceRequestSetBoolParameter(
         triton_object_, key.c_str(), value));
   }
+  void SetDoubleParameter(const std::string& key, double value)
+  {
+    ThrowIfError(TRITONSERVER_InferenceRequestSetDoubleParameter(
```
Tabrizian, line 1167: nit: `THROW_IF_ERROR`

python/test/test_api.py
```diff
@@ -450,12 +450,93 @@ def test_basic_inference(self):
...
+                output_ = response.outputs[name.replace("input", "output")]
```
Tabrizian, line 464: nit: 
```suggestion
                output_item = response.outputs[name.replace("input", "output")]
```

Tabrizian (commented):
Do we need to add double type here too? https://github.com/triton-inference-server/core/blob/5fb865f46061d607892351c238c46a34023c1d3c/src/tritonserver.cc#L637

python/tritonserver/_c/tritonserver_pybind.cc
nnshah1: we usually use all caps only for macros and constants - in this case ThrowIf Error is an actual function -

nnshah1:
> Do we need to add double type here too?
> 
> https://github.com/triton-inference-server/core/blob/5fb865f46061d607892351c238c46a34023c1d3c/src/tritonserver.cc#L637

did a quick search and ammended a few other places

python/tritonserver/_c/tritonserver_pybind.cc
Tabrizian: Ah, I see, thanks for explaining! I thought it is a macro.

Tabrizian (approved):

GuanLuo (approved):

PR #355, opened 2024-05-14
Updated logging of tables and protobuf to use new macros

nnshah1:
- Use new log table macros

### DIFF
--- a/src/server.cc
+++ b/src/server.cc
@@ -1,4 +1,4 @@
-// Copyright 2018-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2018-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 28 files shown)


rmccorm4 (approved):

tanmayv25 (approved):

src/server.cc
```diff
@@ -673,8 +672,8 @@ InferenceServer::PrintBackendAndModelSummary()
       }
     }
   }
-  std::string models_table_string = models_table.PrintTable();
-  LOG_INFO << models_table_string;
+
+  LOG_TABLE_INFO(models_table);
```
GuanLuo, line 676: `LOG_INFO_NOESCAPE(models_table.PrintTable())`? Seems to be not clear if calling out the object type twice..

nnshah1: resolving based on discussion - will revisit after

rmccorm4 (approved):
Deferring refactors/suggestions such as more type-safe logging functions to follow-up PR due to time constraints

PR #356, opened 2024-05-15
Do not call cudaHostAlloc when byte size is zero

Tabrizian:


src/pinned_memory_manager.cc
```diff
@@ -264,17 +264,20 @@ PinnedMemoryManager::Create(const Options& options)
...
-    if (err != cudaSuccess) {
-      buffer = nullptr;
-      LOG_WARNING << "Unable to allocate pinned system memory, pinned memory "
-                     "pool will not be available: "
-                  << std::string(cudaGetErrorString(err));
-    } else if (options.pinned_memory_pool_byte_size_ != 0) {
-      LOG_INFO << "Pinned memory pool is created at '"
-               << PointerToString(buffer) << "' with size "
-               << options.pinned_memory_pool_byte_size_;
+    if (options.pinned_memory_pool_byte_size_ > 0) {
```
GuanLuo, line 267: Move this one layer up, `!host_policy_map_.empty()` case also touches host alloc
```
if (options.pinned_memory_pool_byte_size_ > 0) {
   if (options.host_policy_map_.empty()) {
       ...
```

PR #361, opened 2024-05-23
Added new flag for GPU peer access API control

indrajit96:
Added a new flag "--enable-peer-access" at Triton startup to control creation of CUDA context at server startup.
Jira : https://jirasw.nvidia.com/browse/DLIS-6705
Server PR : https://github.com/triton-inference-server/server/pull/7261


src/cuda_utils.cc
```diff
@@ -83,39 +83,41 @@ GetDeviceMemoryInfo(const int device_id, size_t* free, size_t* total)
 }
 
 Status
-EnablePeerAccess(const double min_compute_capability)
+EnablePeerAccess(const double min_compute_capability, bool enable_peer_access)
```
nnshah1, line 86: seems odd to have a function that has a bool that to disable it?

Why wouldn't we guard against calling this function instead?

src/pinned_memory_manager.cc
```diff
@@ -264,17 +264,20 @@ PinnedMemoryManager::Create(const Options& options)
...
-      buffer = nullptr;
-      LOG_WARNING << "Unable to allocate pinned system memory, pinned memory "
-                     "pool will not be available: "
-                  << std::string(cudaGetErrorString(err));
-    } else if (options.pinned_memory_pool_byte_size_ != 0) {
-      LOG_INFO << "Pinned memory pool is created at '"
-               << PointerToString(buffer) << "' with size "
-               << options.pinned_memory_pool_byte_size_;
+    if (options.pinned_memory_pool_byte_size_ > 0) {
+      auto err = cudaHostAlloc(
```
nnshah1, line 268: For my understanding - the implication here is that cudahostalloc creates a cuda context even in the requested memory size is 0?

src/server.cc
```diff
@@ -244,8 +244,8 @@ InferenceServer::Init()
   }
 
 #endif  // TRITON_ENABLE_GPU
-
-  status = EnablePeerAccess(min_supported_compute_capability_);
+  status =
+      EnablePeerAccess(min_supported_compute_capability_, enable_peer_access_);
   if (!status.IsOk()) {
     // failed to enable peer access is not critical, just inefficient.
     LOG_WARNING << status.Message();
```
nnshah1, line 251: if the user has explicitly asked for enabling peer access - should we error instead of warn?

src/cuda_utils.cc
```diff
@@ -83,39 +83,41 @@ GetDeviceMemoryInfo(const int device_id, size_t* free, size_t* total)
...
-        for (const auto& peer : supported_gpus) {
-          if (host == peer) {
-            continue;
+  if (enable_peer_access) {
+    std::set<int> supported_gpus;
+    bool all_enabled = false;
+    if (GetSupportedGPUs(&supported_gpus, min_compute_capability).IsOk()) {
+      all_enabled = true;
+      int can_access_peer = false;
+      for (const auto& host : supported_gpus) {
```
nnshah1, line 97: small naming tweak - is `device` better than `host` ? `host` implies the cpu  (i.e. host memory)

include/triton/core/tritonserver.h
```diff
@@ -2028,6 +2028,16 @@ TRITONSERVER_DECLSPEC struct TRITONSERVER_Error*
...
     struct TRITONSERVER_ServerOptions* options, bool enable_namespace);
 
+/// Enable peer access to allow GPU device to directly access the memory of
+/// another GPU device
+///
+/// \param options The server options object.
+/// \param enable_peerAccess Whether to enable peer access or not.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC struct TRITONSERVER_Error*
+TRITONSERVER_ServerOptionsSetPeerAccess(
```
Tabrizian, line 2038:
```suggestion
TRITONSERVER_ServerOptionsSetTryEnablingPeerAccess(
```
```suggestion
TRITONSERVER_ServerOptionsSetPeerAccess(
```

include/triton/core/tritonserver.h
```diff
@@ -2028,6 +2028,16 @@ TRITONSERVER_DECLSPEC struct TRITONSERVER_Error*
...
+    struct TRITONSERVER_ServerOptions* options, bool enable_peerAccess);
```
Tabrizian, line 2039:
```suggestion
    struct TRITONSERVER_ServerOptions* options, bool enable_peer_access);
```

src/pinned_memory_manager.cc
```diff
@@ -264,17 +264,20 @@ PinnedMemoryManager::Create(const Options& options)
...
-    if (err != cudaSuccess) {
-      buffer = nullptr;
-      LOG_WARNING << "Unable to allocate pinned system memory, pinned memory "
-                     "pool will not be available: "
-                  << std::string(cudaGetErrorString(err));
-    } else if (options.pinned_memory_pool_byte_size_ != 0) {
-      LOG_INFO << "Pinned memory pool is created at '"
-               << PointerToString(buffer) << "' with size "
-               << options.pinned_memory_pool_byte_size_;
+    if (options.pinned_memory_pool_byte_size_ > 0) {
```
Tabrizian, line 267: @GuanLuo Asked to move it one layer up so that it covers host policy branch below as well.

src/tritonserver.cc
```diff
@@ -1409,6 +1414,16 @@ TRITONSERVER_ServerOptionsSetModelNamespacing(
   return nullptr;  // Success
 }
 
+TRITONAPI_DECLSPEC TRITONSERVER_Error*
+TRITONSERVER_ServerOptionsSetPeerAccess(
+    TRITONSERVER_ServerOptions* options, bool enable_peeraccess)
```
Tabrizian, line 1419:
```suggestion
    TRITONSERVER_ServerOptions* options, bool enable_peer_access)
```

src/server.cc
indrajit96: We can do that, but I feel as this warning has existed till now changing to ERROR can cause existing customer scripts (if any) and/or our tests to maybe break?

src/cuda_utils.cc
GuanLuo: In this context, it is host v.s. peer, can use different naming but the new naming should highlight the idea of "current" device and "peer" device. To enable peer access between two devices, the call needs to be done in both direction

include/triton/core/tritonserver.h
GuanLuo: Can we document that peer access may not be enabled successfully even if it is set and remove `TryEnabling` in the API?

src/server.cc
GuanLuo: > if the user has explicitly asked for enabling peer access - should we error instead of warn?

Default is on for performance reason. I don't know about error out since it doesn't break functionality.

include/triton/core/tritonserver.h
```diff
@@ -2056,6 +2056,16 @@ TRITONSERVER_DECLSPEC struct TRITONSERVER_Error*
...
     struct TRITONSERVER_ServerOptions* options, bool enable_namespace);
 
+/// Enable peer access to allow GPU device to directly access the memory of
+/// another GPU device
+///
+/// \param options The server options object.
+/// \param enable_peerAccess Whether to enable peer access or not.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC struct TRITONSERVER_Error*
+TRITONSERVER_ServerOptionsSetPeerAccess(
```
nnshah1, line 2066: should it be "SetEnablePeerAccess" ? to match command line option - ok either way - just question. 

@Tabrizian

python/tritonserver/_api/_server.py
```diff
@@ -381,6 +386,7 @@ def _create_tritonserver_server_options(
         options.set_model_load_thread_count(self.model_load_thread_count)
         options.set_model_load_retry_count(self.model_load_retry_count)
         options.set_model_namespacing(self.model_namespacing)
+        options.set_peer_access(self.enable_peer_access)
```
nnshah1, line 389: same as above - should this be `set_enable_peer_access`?

src/server.h
```diff
@@ -271,6 +271,8 @@ class InferenceServer {
     enable_model_namespacing_ = e;
   }
 
+  void SetPeerAccessEnabled(const bool e) { enable_peer_access_ = e; }
```
nnshah1, line 274: should match with naming above, either ``SetEnablePeerAccess` or `SetPeerAccess`

nnshah1:
some questions on consistency for naming - but otherwise looks good to me

indrajit96:
Resolved all the naming comments

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -1303,6 +1303,12 @@ class PyServerOptions : public PyWrapper<struct TRITONSERVER_ServerOptions> {
         triton_object_, enable_namespace));
   }
 
+  void SetPeerAccess(bool enable_peer_access)
```
Tabrizian, line 1306:
```suggestion
  void SetEnablePeerAccess(bool enable_peer_access)
```

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -2037,6 +2043,7 @@ PYBIND11_MODULE(triton_bindings, m)
           "set_model_load_retry_count",
           &PyServerOptions::SetModelLoadRetryCount)
       .def("set_model_namespacing", &PyServerOptions::SetModelNamespacing)
+      .def("set_enable_peer_access", &PyServerOptions::SetPeerAccess)
```
Tabrizian, line 2046:
```suggestion
      .def("set_enable_peer_access", &PyServerOptions::SetEnablePeerAccess)
```

src/server.cc
```diff
@@ -201,15 +201,17 @@ InferenceServer::Init()
...
     return status;
   }
-
-  PinnedMemoryManager::Options options(pinned_memory_pool_size_);
-  status = PinnedMemoryManager::Create(options);
-  if (!status.IsOk()) {
-    ready_state_ = ServerReadyState::SERVER_FAILED_TO_INITIALIZE;
-    return status;
+  if (pinned_memory_pool_size_ > 0) {
+    PinnedMemoryManager::Options options(pinned_memory_pool_size_);
```
Tabrizian, line 205: I think we have host policy options as well. @GuanLuo do you know if it is possible to have host policy configs that require allocating pinned memory even when pinned memory pool size is zero.

include/triton/core/tritonserver.h
```diff
@@ -2056,6 +2056,16 @@ TRITONSERVER_DECLSPEC struct TRITONSERVER_Error*
 TRITONSERVER_ServerOptionsSetModelNamespacing(
     struct TRITONSERVER_ServerOptions* options, bool enable_namespace);
 
+/// Enable peer access to allow GPU device to directly access the memory of
+/// another GPU device
```
Tabrizian, line 2060:
```suggestion
/// another GPU device. Note that even when this option is set Triton will only try to enable peer access and
/// might fail to do so if the underlying system doesn't support peer access.
```

include/triton/core/tritonserver.h
```diff
@@ -2056,6 +2056,18 @@ TRITONSERVER_DECLSPEC struct TRITONSERVER_Error*
 TRITONSERVER_ServerOptionsSetModelNamespacing(
     struct TRITONSERVER_ServerOptions* options, bool enable_namespace);
 
+/// Enable peer access to allow GPU device to directly access the memory of
+/// another GPU device. Note that even when this option is set to True, Triton
+/// will only try to enable peer access and might fail to enable it if the
+/// underlying system doesn't support peer access.
+///
+/// \param options The server options object.
+/// \param enable_peerAccess Whether to enable peer access or not.
```
Tabrizian, line 2065:
```suggestion
/// \param enable_peer_access Whether to enable peer access or not.
```
```suggestion
/// \param enable_peerAccess Whether to enable peer access or not.
```

Tabrizian (approved):

src/server.cc
rmccorm4: It looks like this new `if (pinned_memory_pool_size)` is causing issues when pinned memory pool size is set to zero (`--pinned-memory-pool-byte-size=0`).

This only happens to be used in 2 tests:
- L0_pinned_memory : segfaults
- L0_metrics : OK because it doesn't do any inference with this setting, it just checks the metrics

For some flows (possibly most?), when trying to allocate an output buffer it will try pinned -> cpu fallback order within PinnedMemoryManager, but it looks like never gets to cpu fallback because it returns earlier due to no PinnedMemoryManager instance existing.

indrajit96: @Tabrizian I looked at the internals of PinnedMemoryManager
Should we move the check inside of PinnedMemoryManager::Create(const Options& options)
That way we just avoid the cuda calls and create a PinnedMemoryManager

rmccorm4: +1 

For now, we need the PinnedMemoryManager even for non-pinned CPU memory allocations. It is a simpler WAR to always have the manager, and moved the pinned_memory size checks inside of the manager to avoid unnecessary CUDA calls.

PR #364, opened 2024-06-04
fix: Fix various tests caused by cases byte size validation not handled properly

rmccorm4:
#### What does the PR do?

- Recent byte size validation checks were failing the L0_http HttpTest.test_byte unit test which sends a raw binary tensor.
- Recent input byte size validation checks are segfaulting for the CUDA Shared Memory case because it assumes it is in CPU memory and dereferences an invalid pointer.
- Currently, L0_trt_shape_tensor and L0_trt_reformat_free tests are failing because TensorRT infer requests should not check byte size.

The raw binary tensor in the failing test case consists of two memory chunks by the time it gets validated:
1. The size of the bytes element
2. The bytes element

This PR makes some generic refactoring to try to handle any series of buffers, where each element is not guaranteed to fit within a single buffer, and each buffer is not guaranteed to contain a single element. Also it skips byte size validation if input memory type is GPU or input platform is TensorRT.

#### Caveats: 

1. GPU buffers are currently skipped, as they may need some special handling to be dereferenced/checked compared to CPU buffers in the existing code. This will likely cause the element count checks to fail as well.
2. TensorRT inputs are currently skipped, as format-free IO tensors require communication with the backend to determine the byte size.
3. A check is made that the 4-byte `byte_size` indicator for a given element is contained within a single buffer, and is not split across buffers. If a byte_size is split across buffers, then an error is returned. This was just easier to implement, open to suggestions if anyone has a slick solution. Technically the buffer APIs should support doing this, but I assume it is unlikely to be done. Example of this currently unsupported case: 
    - `buffer1=[<size1><element1><size2_partial>]`
    - `buffer2=[<size2_partial><element2>...]`
    - `<size2_partial>` is split across buffers, and is currently rejected.

#### Open Items

- [ ] Verify test plan passes in CI
- [ ] Decide what to do about GPU tensors - probably need to handle them
- [ ] Decide what to do about TensorRT, e.g. shape tensors and reformat-free I/O tensors.

#### Checklist
- [X] PR title reflects the change and is of format `<commit_type>: <Title>`
- [X] Changes are described in the pull request.
- [X] Verified copyright is correct on all changed files.
- [X] All template sections are filled out.

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
https://github.com/triton-inference-server/server/pull/7326

N/A

#### Test plan:
- L0_http
- L0_input_validation
- L0_infer_cudashm
- L0_sequence_batcher_cudashm
- L0_backend_output_detail
- L0_request_cancellation
- L0_trt_reformat_free
- L0_trt_shape_tensors
- Nightly pipeline

CI Pipeline ID: 15635924

#### Background
None

None

### DIFF
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -1203,10 +1203,10 @@ InferenceRequest::Normalize()
               triton::common::GetByteSize(data_type, input_dims);
         }
 
-        bool byte_size_valid =
+        bool byte_size_invalid =
             (byte_size > INT_MAX) ||
             (static_cast<int64_t>(byte_size) != expected_byte_size);
-        if (!skip_byte_size_check && byte_size_valid) {
+        if (!skip_byte_size_check && byte_size_invalid) {
           return Status(
               Status::Code::INVALID_ARG,
               LogRequest() + "input byte size mismatch for input '" + input_id +
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -1301,6 +1301,13 @@ InferenceRequest::ValidateBytesInputs(
           &buffer_memory_type, &buffer_memory_id));
       *expected_byte_size += remaining_buffer_size;
 
+      if (buffer_idx > buffer_count) {
+        return Status(
+            Status::Code::INVALID_ARG,
+            LogRequest() +
+                "element strings exceed the end of the last buffer.");
+      }
+
       // FIXME: Skip GPU buffers for now, return an expected_byte_size of -1 as
       // a signal to skip validation.
       if (buffer_memory_type == TRITONSERVER_MEMORY_GPU) {
@@ -1318,7 +1325,7 @@ InferenceRequest::ValidateBytesInputs(
         return Status(
             Status::Code::INVALID_ARG,
             LogRequest() +
-                "Element byte size indicator exceeds the end of the buffer.");
+                "element byte size indicator exceeds the end of the buffer.");
       }
 
       // Start the next element and reset the remaining element size.
@@ -1349,8 +1356,8 @@ InferenceRequest::ValidateBytesInputs(
   if (buffer_idx != buffer_count) {
     return Status(
         Status::Code::INVALID_ARG,
-        LogRequest() + "expected to process " + std::to_string(buffer_count) +
-            " buffers for inference input '" + input_id + "', only processed " +
+        LogRequest() + "expected " + std::to_string(buffer_count) +
+            " buffers for inference input '" + input_id + "', got " +
             std::to_string(buffer_idx));
   }
 
@@ -1359,7 +1366,7 @@ InferenceRequest::ValidateBytesInputs(
     return Status(
         Status::Code::INVALID_ARG,
         LogRequest() + "expected " + std::to_string(element_count) +
-            " strings for inference input '" + input_id + "', got " +
+            " string elements for inference input '" + input_id + "', got " +
             std::to_string(element_idx));
   }
 
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -1186,6 +1186,11 @@ InferenceRequest::Normalize()
       if (data_type == inference::DataType::TYPE_STRING) {
         RETURN_IF_ERROR(
             ValidateBytesInputs(input_id, input, &expected_byte_size));
+        // FIXME: -1 is used as a signal to skip total byte size validation for
+        // unhandled cases in ValidateBytesInputs.
+        if (expected_byte_size == -1) {
+          return Status::Success;
+        }
       } else {
         expected_byte_size = triton::common::GetByteSize(data_type, input_dims);
       }
@@ -1294,13 +1299,15 @@ InferenceRequest::ValidateBytesInputs(
           &buffer_memory_id);
       *expected_byte_size += remaining_buffer_size;
 
-      // FIXME: Handle GPU buffers
+      // FIXME: Skip GPU buffers for now, return an expected_byte_size of -1 as
+      // a signal to skip validation.
       if (buffer_memory_type == TRITONSERVER_MEMORY_GPU) {
-        LOG_WARNING << "Validation of GPU byte size buffers is not implemented "
-                       "yet, skipping to the next buffer.";
-        buffer = nullptr;
-        remaining_buffer_size = 0;
-        continue;
+        LOG_WARNING << LogRequest()
+                    << "Validation of GPU byte size buffers is not implemented "
+                       "yet, skipping validation for input: "
+                    << input_id;
+        *expected_byte_size = -1;
+        return Status::Success;
       }
     }
 
@@ -1349,8 +1356,6 @@ InferenceRequest::ValidateBytesInputs(
             std::to_string(buffer_idx));
   }
 
-  // FIXME: If the input contains GPU buffers that get skipped, the element
-  //        count will likely not match expectations.
   // Validate the number of processed elements exactly match expectations.
   if (element_idx != element_count) {
     return Status(
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -1176,10 +1176,7 @@ InferenceRequest::Normalize()
     // Note: Since we're using normalized input.ShapeWithBatchDim() here,
     // make sure that all the normalization is before the check.
     {
-      const size_t& byte_size = input.Data()->TotalByteSize();
       const auto& data_type = input.DType();
-      const auto& input_dims = input.ShapeWithBatchDim();
-      int64_t expected_byte_size = INT_MAX;
 
       // FIXME: Skip byte size validation for TensorRT backend because it breaks
       // shape-size assumption. See DLIS-6805 for proper fix for TRT backend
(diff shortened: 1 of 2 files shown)


src/infer_request.cc
```diff
@@ -1186,6 +1186,11 @@ InferenceRequest::Normalize()
       if (data_type == inference::DataType::TYPE_STRING) {
         RETURN_IF_ERROR(
             ValidateBytesInputs(input_id, input, &expected_byte_size));
+        // FIXME: -1 is used as a signal to skip total byte size validation for
+        // unhandled cases in ValidateBytesInputs.
+        if (expected_byte_size == -1) {
+          return Status::Success;
```
yinggeh, line 1192: I don't think we should return immediately incase more checks are appended to this function.

rmccorm4: Open to changes on the GPU portions if you've already been looking into it. The early return can be changed if new checks are added or if we add support for checking the GPU buffers. 

The early return here is temporary to avoid failing the total element count validation at the end if we skip elements in gpu buffers.

yinggeh: My concern is someone started appending new sections without looking backwards. You can change to
```
bool skip_byte_size_check = (expected_byte_size == -1);

if (!skip_byte_size_check &&
    ((byte_size > INT_MAX) || (static_cast<int64_t>(byte_size) != expected_byte_size))) {
```

yinggeh, line 1192: I don't think we should return immediately incase more checks are appended to this function.
My concern is someone will add new lines without looking backwards. You should consider
```
bool skip_byte_size_check = (expected_byte_size == -1);

if (!skip_byte_size_check &&
    ((byte_size > INT_MAX) || (static_cast<int64_t>(byte_size) != expected_byte_size))) {
```

src/infer_request.cc
```diff
@@ -1269,47 +1274,89 @@ InferenceRequest::ValidateBytesInputs(
...
+          buffer_idx++, &remaining_buffer_size, &buffer_memory_type,
+          &buffer_memory_id);
+      *expected_byte_size += remaining_buffer_size;
+
+      // FIXME: Skip GPU buffers for now, return an expected_byte_size of -1 as
+      // a signal to skip validation.
+      if (buffer_memory_type == TRITONSERVER_MEMORY_GPU) {
+        LOG_WARNING << LogRequest()
+                    << "Validation of GPU byte size buffers is not implemented "
+                       "yet, skipping validation for input: "
```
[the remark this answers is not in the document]
rmccorm4: I can update the message, it's probably best-effort on a per backend basis rather than strictly validated.

Were you working on updating the GPU buffer validation in core? What was your latest status regarding that? Feel free to slack.

src/infer_request.cc
GuanLuo: Agree that you should not return early on "success". And further more, I think you should reorder the code here because `ValidateBytesInputs()` already does "total byte size validation":
```
if (byte_data_type) {
  RETURN_IF_ERROR(ValidateBytesInputs(...));
}
else { // non_byte_data
  // line 1179-1185
  // line 1195
  // line 1197-1205
}
```

src/infer_request.cc
```diff
@@ -1269,47 +1274,89 @@ InferenceRequest::ValidateBytesInputs(
...
+  size_t buffer_idx = 0;
+  const size_t buffer_count = input.Data()->BufferCount();
+
+  const char* buffer = nullptr;
+  size_t remaining_buffer_size = 0;
+  TRITONSERVER_MemoryType buffer_memory_type;
+  int64_t buffer_memory_id;
+
+  // Validate elements until all buffers have been fully processed.
+  while (remaining_buffer_size || buffer_idx < buffer_count) {
```
GuanLuo, line 1293: Why not like below to keep variables in scope
```
for (buf : buffers) {
  buffer_size = ...;
  for (checked_size = 0; checked_size < buffer_size) {
    ..
  }
}
```

src/infer_request.cc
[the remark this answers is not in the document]
yinggeh: Slacked you.

src/infer_request.cc
yinggeh: I also just realized that all these checks are inside of a for loop, so you probably intended to`continue` instead of `return` (neither was a good idea). It's prone to make mistakes which makes me feel like to refactor such a long function `Normalize`.

yinggeh: > Agree that you should not return early on "success". And further more, I think you should reorder the code here because `ValidateBytesInputs()` already does "total byte size validation":
> 
> ```
> if (byte_data_type) {
>   RETURN_IF_ERROR(ValidateBytesInputs(...));
> }
> else { // non_byte_data
>   // line 1179-1185
>   // line 1195
>   // line 1197-1205
> }
> ```

`ValidateBytesInputs` asserts all buffers and all strings are visited. It doesn't check the size of the last string imo. I still think it's a good idea to compare expected_byte_size at one place regardless of input data type.

src/infer_request.cc
```diff
@@ -1269,47 +1274,89 @@ InferenceRequest::ValidateBytesInputs(
...
       element_idx++;
+
+      // Advance pointer and remainder by the indicator size.
+      buffer += kElementSizeIndicator;
+      remaining_buffer_size -= kElementSizeIndicator;
+    }
+
+    // If the remaining buffer fits it: consume the rest of the element, proceed
+    // to the next element.
+    if (remaining_buffer_size >= remaining_element_size) {
```
yinggeh, line 1335: Please revisit the logic in this case. I think it will cause segfault/infinite loop if `remaining_buffer_size == remaining_element_size`.

rmccorm4:
Re-assigning to @yinggeh as this involves a lot of refactoring to the original validation logic feature he wrote and it seems like he has some ideas for further refactoring.

src/infer_request.cc
```diff
@@ -1180,30 +1180,40 @@ InferenceRequest::Normalize()
...
-        return Status(
-            Status::Code::INVALID_ARG,
-            LogRequest() + "input byte size mismatch for input '" + input_id +
-                "' for model '" + ModelName() + "'. Expected " +
-                std::to_string(expected_byte_size) + ", got " +
-                std::to_string(byte_size));
+        bool byte_size_valid =
+            (byte_size > INT_MAX) ||
+            (static_cast<int64_t>(byte_size) != expected_byte_size);
+        if (!skip_byte_size_check && byte_size_valid) {
```
rmccorm4, line 1209: This was my mistake, but let's rename `byte_size_valid` to `byte_size_invalid` or similar.

src/infer_request.cc
GuanLuo: Let me know if there is misalignment in my understanding:

"total byte size" in this function scope is `byte_size` which is set to `input.Data()->TotalByteSize()`.

In `ValidateBytesInputs()`, it iterates all elements in the data to get the actual byte size used in execution, and compare it against the buffer boundary (using buffer count and buffer byte size). Isn't the buffer boundary the reflection of `TotalByteSize()`? If iteration exactly ends at buffer boundary, then it implies the `expected size == got size`

src/infer_request.cc
```diff
@@ -1281,31 +1293,34 @@ InferenceRequest::ValidateBytesInputs(
...
       // Reset remaining buffer size and pointers for next buffer.
-      buffer = input.Data()->BufferAt(
-          buffer_idx++, &remaining_buffer_size, &buffer_memory_type,
-          &buffer_memory_id);
+      RETURN_IF_ERROR(input.DataBuffer(
+          buffer_next_idx++, (const void**)(&buffer), &remaining_buffer_size,
+          &buffer_memory_type, &buffer_memory_id));
       *expected_byte_size += remaining_buffer_size;
 
+      if (buffer_next_idx > buffer_count) {
```
GuanLuo, line 1314: Isn't this already checked by the while loop conditional? The largest value of `buffer_next_idx` at the beginning of the loop will be `buffer_count-1`, after line 1310 will be `buffer_count` and this line will never be true.

src/infer_request.cc
```diff
@@ -1180,28 +1180,40 @@ InferenceRequest::Normalize()
...
-        // unhandled cases in ValidateBytesInputs.
-        if (expected_byte_size == -1) {
-          return Status::Success;
+
+      // Skip byte size validation for TensorRT backend because it breaks
+      // shape-size assumption
+      bool skip_byte_size_check = false;
+      constexpr char trt_prefix[] = "tensorrt_";
+      const std::string& platform = model_raw_->Config().platform();
+      skip_byte_size_check |= (platform.rfind(trt_prefix) == 0);
```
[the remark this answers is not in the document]
rmccorm4: Yes this is a WAR to avoid breaking existing TRT reformat_free features. Shape tensors can be explicitly checked, but reformat_free is not as easy to check without some communication with the backend or something. We wanted to unblock Sai who is working on other TRT features and avoid noise in failing TRT tests until we have a proper fix for this.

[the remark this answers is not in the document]
GuanLuo: Can you or Yingge add some tracking info here if it will be addressed later?

src/infer_request.cc
yinggeh: `buffer_next_idx > buffer_count` statement is true if `remaining_buffer_size>0` and `buffer_next_idx == buffer_count` at the top while loop condition check. This happens when the last string element content is in the last two buffers but the second buffer somehow does not exist.

src/infer_request.cc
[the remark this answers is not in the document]
yinggeh: Adding TODOs?

src/infer_request.cc
yinggeh: If iteration exactly ends at buffer boundary, it only means the number of bytes matches `kElementSizeIndicator`.

src/infer_request.cc
GuanLuo: >This happens when the last string element content is in the last two buffers but the second buffer somehow does not exist.

Can you point me to the test case that illustrate this edge case? I can't see why the 2nd buffer will not exist while input data says it contains 2 buffer.

> if `remaining_buffer_size>0` and `buffer_next_idx == buffer_count`
if `remaining_buffer_size>0`, then line 1307 `if (!remaining_buffer_size)` is false and won't hit this line.

src/infer_request.cc
GuanLuo: "Iteration ends" refer to the point after checking all the elements, does that make sense?

src/infer_request.cc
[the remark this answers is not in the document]
GuanLuo: Add ticket number here and add comment in the ticket itself to revert / address this change as part of definition of done

GuanLuo (commented):
See new comments in opened discussion.

src/infer_request.cc
yinggeh: You are right. This check is redundent.

src/infer_request.cc
yinggeh: Gotcha. Here `byte_size` is comparing to itself (expected_byte_size) which is redundant.

yinggeh: Dumb question: Is `input.Data()->TotalByteSize()` guarenteed to be accurate? I can see this check is useful when there is a size mismatch between `byte_size` and bytes get read from all buffers.

yinggeh:
> Please update the description with the latest information and pipeline ID, I need to see the test status

I believe I have updated with the latest info and pipeline ID. Is there anything particular you think is missing?

src/infer_request.cc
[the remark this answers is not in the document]
rmccorm4: Looks like ticket numbers, details, and follow-ups were added - resolving

rmccorm4 (commented):
I can't approve this PR because I originally opened it, but I am verbally approving this if all the affected CI tests in the test plan description are passing.

I added a follow-up ticket DLIS-6833 to address some of the logical refactoring brought up by Guan after code freeze.

GuanLuo (approved):

src/infer_request.cc
GuanLuo: >Is input.Data()->TotalByteSize() guaranteed to be correct?

This function should return the total size of the buffers.

> mismatch between `byte_size` and bytes get read from all buffers

From the current implementation they are the same thing, right? I think the comparison should be between buffer byte size (from `TotalByteSize()`) and byte size computed from other sources (i.e. shape * data byte size for non-byte tensor, and iterating elements for byte tensor).

yinggeh:
Please also review related server PR github.com/triton-inference-server/server/pull/7326

PR #365, opened 2024-06-05
fix: Move pinned memory size checks to PinnedMemoryManager and update C API version for new EnablePeerAccess API

indrajit96:
1. Update C API version for changes in 
"Added new flag for GPU peer access API control #361"
2. Fix a seg fault in PinnedMemoryManager::AllocInternal(

Seg Fault Details:

Triton tries to access pinned_memory_buffer [here](https://github.com/triton-inference-server/core/blob/190d8e3b7e5cf89234a4e8b7006f8b87bfa4c237/src/pinned_memory_manager.cc#L153)
Which was null because of the new change introduced in #361.

Triton behavior is to verify it failed to allocate pinned memory by looping through pinned_memory_buffer AND then do a malloc. This looks like a design choice made.

### DIFF
--- a/include/triton/core/tritonserver.h
+++ b/include/triton/core/tritonserver.h
@@ -91,7 +91,7 @@ struct TRITONSERVER_MetricFamily;
 ///   }
 ///
 #define TRITONSERVER_API_VERSION_MAJOR 1
-#define TRITONSERVER_API_VERSION_MINOR 31
+#define TRITONSERVER_API_VERSION_MINOR 32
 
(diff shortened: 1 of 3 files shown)


Tabrizian (approved):

Tabrizian:
@indrajit96 please update the title to be more representative of the actual changes.

rmccorm4 (approved):
LGTM, and please update description with more details about the bug and fix

indrajit96:
> LGTM, and please update description with more details about the bug and fix

Have added a brief description here, will add more detailed description in the Jira bug

PR #369, opened 2024-06-10
fix: Preserve Sequence Flags in Ensemble models

indrajit96:
Revert changes made in https://github.com/triton-inference-server/core/commit/e978d80c5138890bc48d798f5187b11be3ac6d23 to ensure START and END flags are propagated correctly in ensemble models.

RCA : Any ensemble inference request where the user sets the flags "sequence_start" and "sequence_end" was getting ignored.
Fix : We now check if the request had the FLAGS set and then pass them forward as is.

Future scope : This was a bug introduced in https://github.com/triton-inference-server/core/commit/e978d80c5138890bc48d798f5187b11be3ac6d23 which was intended to clean up code. We do not expect these changes to come back again.
Have added a TC in https://github.com/triton-inference-server/server/pull/7344 to ensure regression in future.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-# Copyright 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2023-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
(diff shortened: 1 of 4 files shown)


GuanLuo:
Update title

rmccorm4:
Can you elaborate in the description why the reverted changes were breaking the propagation?

PR #370, opened 2024-06-13
fix: Fix response iterator in Python In-Process API

Tabrizian:
#### What does the PR do?
The response iterator destroyed the request object after final response is received which lead to destruction of the queue and responses not being retrieved by the user.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [X] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [x] test

#### Related PRs:
N/A

N/A

#### Test plan:
Updated L0_python_api.

- CI Pipeline ID:
15804851

#### Caveats:
N/A

#### Background
N/A

N/A

### DIFF
--- a/python/test/test_api.py
+++ b/python/test/test_api.py
@@ -455,15 +455,20 @@ class InferenceTests(unittest.TestCase):
             "bool_input": numpy.random.rand(1, 100).astype(dtype=numpy.bool_),
         }
 
-        for response in server.model("test").infer(
+        response_iterator = server.model("test").infer(
             inputs=inputs,
             output_memory_type="cpu",
             raise_on_error=True,
-        ):
-            for input_name, input_value in inputs.items():
-                output_value = response.outputs[input_name.replace("input", "output")]
-                output_value = numpy.from_dlpack(output_value)
-                numpy.testing.assert_array_equal(input_value, output_value)
+        )
+
+        responses = list(response_iterator)
+        self.assertTrue(len(responses), 1)
+        response = responses[0]
+
+        for input_name, input_value in inputs.items():
+            output_value = response.outputs[input_name.replace("input", "output")]
+            output_value = numpy.from_dlpack(output_value)
+            numpy.testing.assert_array_equal(input_value, output_value)
 
         # test normal bool
         inputs = {"bool_input": [[True, False, False, True]]}
--- a/python/tritonserver/_api/_response.py
+++ b/python/tritonserver/_api/_response.py
@@ -29,7 +29,6 @@
 from __future__ import annotations
 
 import asyncio
-import inspect
 import queue
 from dataclasses import asdict, dataclass, field
 from typing import TYPE_CHECKING, Optional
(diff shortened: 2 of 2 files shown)


python/test/test_api.py
```diff
@@ -455,15 +455,20 @@ def test_basic_inference(self):
...
             output_memory_type="cpu",
             raise_on_error=True,
-        ):
-            for input_name, input_value in inputs.items():
-                output_value = response.outputs[input_name.replace("input", "output")]
-                output_value = numpy.from_dlpack(output_value)
-                numpy.testing.assert_array_equal(input_value, output_value)
+        )
+
+        responses = list(response_iterator)
```
rmccorm4, line 464: This is subtle at a glance, but `list(response_iterator)` is actually blocking and iterating on the iterator until it's completely consumed, right?

python/tritonserver/_api/_response.py
```diff
@@ -308,9 +304,6 @@ def _response_callback(self, response, flags, unused):
             self._queue.put(response)
             if self._user_queue is not None:
                 self._user_queue.put(response)
-            if flags == TRITONSERVER_ResponseCompleteFlag.FINAL:
-                del self._request
```
rmccorm4, line 312: Note to self to verify that this might fix the issue with `server.stop()` in `L0_python_api`.

python/test/test_api.py
Tabrizian: Correct. It looks like sometimes the previous version of the test didn't execute since there were no responses observed. This is to make sure that the response length is always one.

python/tritonserver/_api/_response.py
```diff
@@ -182,9 +181,6 @@ def _response_callback(self, response, flags, unused):
                 asyncio.run_coroutine_threadsafe(
                     self._user_queue.put(response), self._loop
                 )
-            if flags == TRITONSERVER_ResponseCompleteFlag.FINAL:
-                del self._request
-                self._request = None
```
rmccorm4, line 187: From bindings / lifecycle perspectives, are there any cons introduced by **not** deleting the request here or somewhere else? Will an object get kept around unexpectedly?

Tabrizian: I added `while True` but didn't notice any memory increase.

Tabrizian: Memory usage appears to be stable. Might need stress tests to make sure it is not leaking any memory.

python/tritonserver/_api/_response.py
rmccorm4: Passed 10/10 reruns, so hoping it's the fix. We'll see for sure later on.

GuanLuo:
What is being held in request object? Aren't the responses being stored in the iterator?

Tabrizian:
@GuanLuo I think the `response_iterator` was relying on a queue inside the request object which might be the root cause for this issue.

PR #376, opened 2024-07-04
fix: Fix cpu-only-py3-base job

yinggeh:
#### What does the PR do?
Update CMakeLists.txt to make sure CUDA libraries are not included in the build process when TRITON_ENABLE_GPU=OFF.

#### Checklist

#### Commit Type:
- [x] fix

#### Related PRs:
N/A

N/A

#### Test plan:
N/A

- CI Pipeline ID:
16347752

#### Caveats:
N/A

#### Background
N/A

N/A

### DIFF
--- a/src/test/CMakeLists.txt
+++ b/src/test/CMakeLists.txt
@@ -650,23 +650,6 @@ if(${TRITON_ENABLE_GPU})
     ${CUDA_MEMORY_MANAGER_SRCS}
     ${CUDA_MEMORY_MANAGER_HDRS}
   )
-else()
-  add_executable(
-    input_byte_size_test
-    input_byte_size_test.cc
-  )
-endif() # TRITON_ENABLE_GPU
-
-set_target_properties(
-  input_byte_size_test
-  PROPERTIES
-    SKIP_BUILD_RPATH TRUE
-    BUILD_WITH_INSTALL_RPATH TRUE
-    INSTALL_RPATH_USE_LINK_PATH FALSE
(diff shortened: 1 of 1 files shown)


src/test/CMakeLists.txt
```diff
@@ -659,39 +666,58 @@ set_target_properties(
...
 
-target_include_directories(
-  input_byte_size_test
-  PRIVATE
-    ${CMAKE_CURRENT_SOURCE_DIR}/..
-    ${CMAKE_CURRENT_SOURCE_DIR}/../../include
-    ${GTEST_INCLUDE_DIRS}
-    ${CNMEM_PATH}/include
-)
+if(${TRITON_ENABLE_GPU})
```
fpetrini15, line 669: My preference would be to merge this if/else with the if/else above. Doing so would also make it more consistent with the rest of the file.

yinggeh: My test cases are partially run in CPU.

fpetrini15: Right, but I think it can be condensed into a single if/else block for clarify. I'll leave it up to you.

fpetrini15 (approved):
Optional comment. Otherwise LGTM, thanks for the quick fix!

src/test/CMakeLists.txt
yinggeh: I agree. Fixed.

fpetrini15 (approved):
LGTM, pending passing pipeline. Great work!

PR #377, opened 2024-07-08
feat: Metrics API enhanced for Failure counters

indrajit96:
#### What does the PR do?
Currently We don't have any mechanism to count any failures that happen in triton core before the request gets to the backend.

This PR enhances existing metrics API to incorporate failure and failure counts in inference requests, along with categories for the types of failures as a "reason" label with initial support for: BACKEND, CANCELED, REJECTED, and OTHER.

**NOTE**: Ensemble failure metrics will reflect the failure counts of their composing models,
but currently do not capture the same granularity for the "reason" label and will default to the "OTHER" reason.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
https://github.com/triton-inference-server/server/pull/7423

RespondIfError function call in core

#### Test plan:

See test plan in https://github.com/triton-inference-server/server/pull/7423

- closes Jira issue: DLIS-5530


src/backend_model_instance.cc
```diff
@@ -558,7 +559,8 @@ TritonModelInstance::PrepareRequestsOrRespond(
   // If any errors occurred, respond with error for each request.
   if (!status.IsOk()) {
     for (auto& r : requests) {
-      InferenceRequest::RespondIfError(r, status, true /* release_requests */);
+      InferenceRequest::RespondIfError(
+          r, status, true /* release_requests */, FailureReason::BACKEND);
```
rmccorm4, line 563: Shouldn't this be `OTHER`?

src/backend_model_instance.cc
```diff
@@ -685,7 +687,8 @@ TritonModelInstance::Execute(
     for (TRITONBACKEND_Request* tr : triton_requests) {
       std::unique_ptr<InferenceRequest> ur(
           reinterpret_cast<InferenceRequest*>(tr));
-      InferenceRequest::RespondIfError(ur, status, true /* release_requests */);
+      InferenceRequest::RespondIfError(
+          ur, status, true /* release_requests */, FailureReason::OTHER);
```
rmccorm4, line 691: Should this be `BACKEND`?

src/backend_model_instance.cc
```diff
@@ -964,7 +967,7 @@ TRITONBACKEND_ModelInstanceReportStatistics(
   InferenceRequest* tr = reinterpret_cast<InferenceRequest*>(request);
   tr->ReportStatistics(
       ti->MetricReporter(), success, exec_start_ns, compute_start_ns,
-      compute_end_ns, exec_end_ns);
+      compute_end_ns, exec_end_ns, FailureReason::BACKEND);
```
rmccorm4, line 970: Is it possible for a backend ModelInstanceExecute() to also return an error, and then this `BACKEND` failure would get double counted?

src/metric_model_reporter.cc
```diff
@@ -101,6 +102,14 @@ MetricReporterConfig::ParseQuantiles(std::string options)
 //
 // MetricModelReporter
 //
+const std::map<FailureReason, std::string>
+    MetricModelReporter::failure_reasons_map = {
+        {FailureReason::SUCCESS, "SUCCESS"},
```
rmccorm4, line 107: should remove `SUCCESS` reason for failures

src/metric_model_reporter.cc
```diff
@@ -227,6 +236,15 @@ MetricModelReporter::InitializeCounters(
       counters_[name] = CreateMetric<prometheus::Counter>(*family_ptr, labels);
     }
   }
+
+  // Initialize failure metrics with reasons
+  for (const auto& reason_pair : failure_reasons_map) {
+    std::map<std::string, std::string> extended_labels = labels;
```
rmccorm4, line 242: Line 201 above is creating an extra failure label with no `reason` field:
```
  counter_families_["inf_failure"] = &Metrics::FamilyInferenceFailure();
```
and should probably be removed and rel-noted with this change.

src/sequence_batch_scheduler/sequence_batch_scheduler.cc
```diff
@@ -1173,7 +1175,8 @@ SequenceBatchScheduler::ReaperThread(const int nice)
           "timeout of the corresponding sequence has been expired");
       for (auto& backlog : expired_backlogs) {
         for (auto& req : *backlog->queue_) {
-          InferenceRequest::RespondIfError(req, rejected_status, true);
+          InferenceRequest::RespondIfError(
+              req, rejected_status, true, FailureReason::REJECTED);
```
rmccorm4, line 1179:
```suggestion
              req, rejected_status, true /* release_requests */, FailureReason::REJECTED);
```

src/sequence_batch_scheduler/sequence_batch_scheduler.cc
```diff
@@ -91,7 +92,8 @@ CancelRequests(std::vector<std::unique_ptr<InferenceRequest>>&& requests)
       LOG_ERROR << status.Message();
     }
     // Respond the request as cancelled.
-    InferenceRequest::RespondIfError(req, cancelled_status, true);
+    InferenceRequest::RespondIfError(
+        req, cancelled_status, true, FailureReason::CANCELED);
```
rmccorm4, line 96:
```suggestion
        req, cancelled_status, true /* release_requests */, FailureReason::CANCELED);
```

src/dynamic_batch_scheduler.cc
```diff
@@ -54,7 +54,8 @@ FinishSkippedRequests(
 {
   for (auto& queue : requests) {
     for (auto& request : queue) {
-      InferenceRequest::RespondIfError(request, response_status, true);
+      InferenceRequest::RespondIfError(
+          request, response_status, true, FailureReason::OTHER);
```
rmccorm4, line 58:
```suggestion
          request, response_status, true /* release_requests */, FailureReason::OTHER);
```

src/dynamic_batch_scheduler.cc
```diff
@@ -54,7 +54,8 @@ FinishSkippedRequests(
 {
   for (auto& queue : requests) {
     for (auto& request : queue) {
-      InferenceRequest::RespondIfError(request, response_status, true);
+      InferenceRequest::RespondIfError(
```
rmccorm4, line 57: It looks like `FinishSkippedRequests` is called for both `rejected` and `cancelled` requests. So this might do some double counting, and may not label them correctly as it currently is.

Can you elaborate on what this function is doing differently, and why it's needed if we call `RespondIfError` in the rejected/cancelled cases already? CC @kthui who may be able to help explain as well.

src/infer_request.cc
```diff
@@ -1375,8 +1383,10 @@ void
 InferenceRequest::ReportStatistics(
     MetricModelReporter* metric_reporter, bool success,
     const uint64_t compute_start_ns, const uint64_t compute_input_end_ns,
-    const uint64_t compute_output_start_ns, const uint64_t compute_end_ns)
+    const uint64_t compute_output_start_ns, const uint64_t compute_end_ns,
+    FailureReason reason)
 {
+  std::cerr << "===== InferenceRequest::ReportStatistics =====\n";
```
rmccorm4, line 1389: remove

src/infer_request.h
```diff
@@ -671,7 +673,13 @@ class InferenceRequest {
...
       MetricModelReporter* metric_reporter, bool success,
       const uint64_t compute_start_ns, const uint64_t compute_input_end_ns,
-      const uint64_t compute_output_start_ns, const uint64_t compute_end_ns);
+      const uint64_t compute_output_start_ns, const uint64_t compute_end_ns,
+      FailureReason reason);
+
+  // void ReportStatistics(
+  //     MetricModelReporter* metric_reporter, bool success,
+  //     const uint64_t compute_start_ns, const uint64_t compute_input_end_ns,
+  //     const uint64_t compute_output_start_ns, const uint64_t compute_end_ns);
```
rmccorm4, line 682: remove

src/infer_stats.cc
```diff
@@ -48,7 +68,11 @@ InferenceStatsAggregator::UpdateFailure(
 
 #ifdef TRITON_ENABLE_METRICS
   if (metric_reporter != nullptr) {
-    metric_reporter->IncrementCounter("inf_failure", 1);
+    std::string reason_str = failureReasonToString(reason);
+    std::cerr
+        << "===== InferenceStatsAggregator::UpdateFailure - Failure Reason: "
+        << reason_str << std::endl;
```
rmccorm4, line 74: remove

src/ensemble_scheduler/ensemble_scheduler.cc
```diff
@@ -1136,7 +1138,8 @@ EnsembleContext::FinishEnsemble(std::unique_ptr<InferenceResponse>&& response)
                 "more "
                 "ensemble steps can be made");
         InferenceRequest::RespondIfError(
-            request_tracker_->Request(), ensemble_status_);
+            request_tracker_->Request(), ensemble_status_, true,
```
rmccorm4, line 1141:
```suggestion
            request_tracker_->Request(), ensemble_status_, true /* release_requests */,
```

src/ensemble_scheduler/ensemble_scheduler.cc
```diff
@@ -1149,7 +1152,8 @@ EnsembleContext::FinishEnsemble(std::unique_ptr<InferenceResponse>&& response)
           ensemble_status_);
     } else {
       InferenceRequest::RespondIfError(
-          request_tracker_->Request(), ensemble_status_);
+          request_tracker_->Request(), ensemble_status_, true,
```
rmccorm4, line 1155:
```suggestion
          request_tracker_->Request(), ensemble_status_, true /* release_requests */,
```

src/infer_request.cc
```diff
@@ -443,6 +458,13 @@ InferenceRequest::RespondIfError(
           std::move(response), TRITONSERVER_RESPONSE_COMPLETE_FINAL, status),
       (request->LogRequest() + "failed to send error response").c_str());
 
+  request->ReportStatistics(
```
rmccorm4, line 461: While calling `ReportStatistics` may be easy, is it the right thing to call? Will it have any adverse side effects? Will it mess up the `request::trace_`? Please take a look over this part of the flow and see if it makes more sense to extract out a helper or something. It's possible this is correct, but just want a second look at it. Keep in mind the possibility of `ReportStatistics` also getting called twice now with the call from `TRITONBACKEND_ModelInstanceReportStatistics` as well, and that may be undesirable.

indrajit96: Separated out the helper to avoid interacting with request::trace_
Added new func 
void ReportErrorStatistics(

src/backend_model_instance.cc
```diff
@@ -687,6 +687,9 @@ TritonModelInstance::Execute(
     for (TRITONBACKEND_Request* tr : triton_requests) {
       std::unique_ptr<InferenceRequest> ur(
           reinterpret_cast<InferenceRequest*>(tr));
+      // CAVEAT: Any Backend calling TRITONBACKEND_ModelInstanceReportStatistics
+      // with an error can result in double counting of the failure metric,
+      // currently NO backend does this
```
rmccorm4, line 692: Tweak to the wording as the current wording isn't completely accurate. Reporting an error with the TRITONBACKEND_ModelInstanceReportStatistics is OK and is common. Returning an error from TRITONBACKEND_ModelInstanceExecute is less likely.
```suggestion
      // NOTE: If a backend both returns an error in TRITONBACKEND_ModelInstanceExecute
      // and reports an error with TRITONBACKEND_ModelInstanceReportStatistics,
      // this can result in double counting of the failure metric for the same request. However, 
      // it is currently not expected for this to be a common case, as the return value of
      // TRITONBACKEND_ModelInstanceExecute is used to express ownership of the request
      // rather than success of an inference request. See tritonbackend.h for more details on this.
```

rmccorm4: may need to fix spacing / linting of this after suggestion.

src/infer_stats.h
```diff
@@ -39,6 +39,9 @@
 
 namespace triton { namespace core {
 
+// Define the FailureReason enum within the triton::core namespace
+enum class FailureReason { SUCCESS, REJECTED, CANCELED, BACKEND, OTHER };
```
rmccorm4, line 43:
```suggestion
enum class FailureReason { REJECTED, CANCELED, BACKEND, OTHER };
```

src/dynamic_batch_scheduler.cc
rmccorm4: NVM, it turned out one usage was in dynamic batcher and the other was in sequence batcher.

src/backend_model_instance.cc
rmccorm4: tritonbackend.h docs and some reference examples like onnx and python helped confirm that returning an error from the execute function will be unlikely, and is typically used to reflect request ownership rather than expressing a failed request/response at the model level. A NOTE is still included in the code for possibility of double counting for future reference just in case a custom backend implement runs into this.

src/backend_model_instance.cc
```diff
@@ -558,7 +559,8 @@ TritonModelInstance::PrepareRequestsOrRespond(
   // If any errors occurred, respond with error for each request.
   if (!status.IsOk()) {
     for (auto& r : requests) {
-      InferenceRequest::RespondIfError(r, status, true /* release_requests */);
+      InferenceRequest::RespondIfError(
+          r, status, true /* release_requests */, FailureReason::OTHER);
```
rmccorm4, line 563: Can you think of any test cases to test some of the `reason=OTHER` scenarios (not counting ensembles)? Can be a follow-up, don't need to be in these 24.07 PRs, but would be good to have a test case for these.

src/ensemble_scheduler/ensemble_scheduler.cc
```diff
@@ -83,21 +83,23 @@ class RequestTracker {
...
       const auto& infer_stats = context_stats_aggregator_.ImmutableInferStats();
-      request_->ReportStatisticsWithDuration(
-          metric_reporter_, status_.IsOk(), compute_start_ns_,
-          infer_stats.compute_input_duration_ns_,
-          infer_stats.compute_infer_duration_ns_,
-          infer_stats.compute_output_duration_ns_);
-      if (status_.IsOk()) {
-        stats_aggregator_->UpdateInferBatchStatsWithDuration(
-            metric_reporter_, std::max(1U, request_->BatchSize()),
+      if (request_ != nullptr) {
```
rmccorm4, line 86: What was the reason for this change? Did you run into some newly introduced segfault case?

src/infer_request.cc
```diff
@@ -443,6 +458,11 @@ InferenceRequest::RespondIfError(
           std::move(response), TRITONSERVER_RESPONSE_COMPLETE_FINAL, status),
       (request->LogRequest() + "failed to send error response").c_str());
 
+  request->ReportErrorStatistics(
+      request->model_raw_->MetricReporter().get(),
+      reason  // reason: The specific reason for the failure
```
rmccorm4, line 463:
```suggestion
      reason
```

src/ensemble_scheduler/ensemble_scheduler.cc
indrajit96: Yes, I was facing a segfault here while testing the changes.
Used GDB to debug found calls were made when the request_ ptr was null, hence added the check

src/infer_request.h
```diff
@@ -673,6 +675,11 @@ class InferenceRequest {
       const uint64_t compute_start_ns, const uint64_t compute_input_end_ns,
       const uint64_t compute_output_start_ns, const uint64_t compute_end_ns);
 
+  // Report the error statistics to stats collectors associated with the
+  // request. reason provided
```
rmccorm4, line 679:
```suggestion
  // request.
  // FIXME: A separate function may not be necessary here, but is being used cautiously in case of
  // unforeseen issues such as possibly capturing a trace twice. This should be
  // revisited and better tested to see if the ReportStatistics function can be used as-is for the newly captured failure cases.
```

rmccorm4 (commented):
High level code LGTM since we're aiming for 24.07 cherry-pick, left some small comments and follow-ups.

src/ensemble_scheduler/ensemble_scheduler.cc
rmccorm4: Do you know what scenario or part of code newly introduced the segfault? It would be good to understand which code now makes requests possibly nullptr that wasn't happening before.

indrajit96: Fixed :)
Detailed RCA
this happens only in case of dynamic batcher when using an ensemble model
FinishSkippedRequests gets called that in turn calls
RespondIfError which releases the request. (So far soo good)
But
In https://github.com/triton-inference-server/core/blob/627eb5ae9f65a9c146a99272a2ba91c01e0ac576/src/ensemble_scheduler/ensemble_scheduler.cc#L987
We set a callback
RequestComplete this in turn calls
DecrementCounter this in turn calls
ReportStatisticsWithDuration which in turn causes the crash.
The main "culprit" is calling RespondIfError with
true /* release_requests */

src/infer_request.cc
rmccorm4: please remove the comment, it's explained in function docstring and variable name

src/infer_request.h
rmccorm4: please update with suggestion

src/infer_stats.h
rmccorm4: please update

src/ensemble_scheduler/ensemble_scheduler.cc
rmccorm4: Nice explanation, thanks for diving into it. So the `if (request_ == nullptr)` change is now essentially "unnecessary", right? But we can keep it in for safety purposes.

src/backend_model_instance.cc
rmccorm4: please update with suggestion

src/ensemble_scheduler/ensemble_scheduler.cc
indrajit96: Yes :)

rmccorm4 (approved):

PR #380, opened 2024-07-14
Fix shape and reformat free tensor handling in the input byte size check

pskiran1:
- Added support for the `is_non_linear_format_io` flag in the model configuration to identify if input/output tensors use a non-linear IO format.
- For shape tensors, use the original shape for byte size validation.
- For tensors using the non-linear IO format (reformat-free tensors), validate the input byte size in the backend.

### DIFF
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -1200,7 +1200,6 @@ InferenceRequest::Normalize()
     // make sure that all the normalization is before the check.
     {
       const auto& data_type = input.DType();
-      bool skip_byte_size_check = false;
 
       if (!input.IsReformatFreeTensor()) {
         TRITONSERVER_MemoryType input_memory_type;
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -1201,6 +1201,8 @@ InferenceRequest::Normalize()
     {
       const auto& data_type = input.DType();
 
+      // Non-linear IO format input byte size validation will be handled in the
+      // backend.
       if (!input.IsNonLinearFormatIo()) {
         TRITONSERVER_MemoryType input_memory_type;
         // Because Triton expects STRING type to be in special format
@@ -1213,6 +1215,9 @@ InferenceRequest::Normalize()
           // FIXME: Temporarily skips byte size checks for GPU tensors. See
           // DLIS-6820.
         } else {
+          // Shape tensor with dynamic batching does not introduce a new
+          // dimension to the tensor but adds an additional value to the 1-D
+          // array.
           const std::vector<int64_t>& input_dims =
               input.IsShapeTensor() ? input.OriginalShape()
                                     : input.ShapeWithBatchDim();
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -1563,7 +1563,7 @@ Status
 InferenceRequest::Input::SetIsNonLinearFormatIo(
     const bool is_non_linear_format_io)
 {
-  is_non_linear_  format_io_ = is_non_linear_format_io;
+  is_non_linear_format_io_ = is_non_linear_format_io;
   return Status::Success;
 }
 
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -1211,7 +1211,12 @@ InferenceRequest::Normalize()
           // FIXME: Temporarily skips byte size checks for GPU tensors. See
           // DLIS-6820.
         } else {
-          const std::vector<int64_t>& input_dims = input.ShapeWithBatchDim();
+          // Shape tensor with dynamic batching does not introduce a new
+          // dimension to the tensor but adds an additional value to the 1-D
+          // array.
+          const std::vector<int64_t>& input_dims =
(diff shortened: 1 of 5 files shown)


src/model_config_utils.cc
```diff
@@ -1732,6 +1732,17 @@ ValidateModelInput(
...
 
+  if ((platform != kTensorRTPlanPlatform) && io.is_non_linear_format_io()) {
+    return Status(
+        Status::Code::INVALID_ARG,
+        "Non-linear IO format is only supported for the TensorRT platform");
+  }
+
+  if (io.is_non_linear_format_io() && (io.dims_size() != 3)) {
+    return Status(
+        Status::Code::INVALID_ARG, "Non-linear IO format input require 3 dims");
```
tanmayv25, line 1743:
```suggestion
        Status::Code::INVALID_ARG, "Non-linear IO format input requires 3 dims");
```

src/model_config_utils.cc
```diff
@@ -1768,6 +1779,12 @@ ValidateModelOutput(
         "shape tensors are only supported for TensorRT platform");
   }
 
+  if ((platform != kTensorRTPlanPlatform) && io.is_non_linear_format_io()) {
```
tanmayv25, line 1782: We should also check for the number of dimension to be 3 or not for non_linear_format_io tensor.

I would advise to create a new function called `ValidateNonLinearFormatIO()` which could check the platflorm and number of dims. Then call to this function when a nonlinear tensor is encountered.

src/infer_request.cc
```diff
@@ -1182,28 +1201,26 @@ InferenceRequest::Normalize()
...
-      // shape-size assumption. See DLIS-6805 for proper fix for TRT backend
-      // reformat_free tensors.
-      bool skip_byte_size_check = false;
-      constexpr char trt_prefix[] = "tensorrt_";
-      const std::string& platform = model_raw_->Config().platform();
-      skip_byte_size_check |= (platform.rfind(trt_prefix) == 0);
-
-      if (!skip_byte_size_check) {
+      // Non-linear IO format input byte size validation will be handled in the
+      // backend.
```
tanmayv25, line 1205: TensorRT backend

src/infer_request.cc
```diff
@@ -1182,28 +1201,26 @@ InferenceRequest::Normalize()
...
           RETURN_IF_ERROR(
               ValidateBytesInputs(input_id, input, &input_memory_type));
+
           // FIXME: Temporarily skips byte size checks for GPU tensors. See
           // DLIS-6820.
-          skip_byte_size_check |=
-              (input_memory_type == TRITONSERVER_MEMORY_GPU);
         } else {
-          const auto& input_dims = input.ShapeWithBatchDim();
+          // Shape tensor with dynamic batching does not introduce a new
```
tanmayv25, line 1218: Can this be handled within ShapeWIthBatchDim call itself?
As in the input object already knows whether it is a shape tensor or not. How about not adding an extra batch dim in such a scenario.

src/infer_request.h
```diff
@@ -241,6 +247,7 @@ class InferenceRequest {
     std::vector<int64_t> shape_;
     std::vector<int64_t> shape_with_batch_dim_;
     bool is_shape_tensor_;
+    bool is_non_linear_format_io_;
```
GuanLuo, line 250: Can these two be combined as enum? i.e.
```
enum class {
  TENSOR;
  SHAPE_TENSOR;
  NON_LINEAR;
};

Comment (pskiran1) [reply src/infer_request.h]: Added a new enum class TensorType, encapsulated in Input class.
`enum class TensorType { TENSOR, SHAPE_TENSOR, NON_LINEAR };`

File: src/model_config_utils.cc [hunk 6]
```diff
@@ -1712,6 +1712,26 @@ ValidateInstanceGroup(
   return Status::Success;
 }
 
+Status
+ValidateNonLinearFormatIO(
+    const inference::ModelInput& io, const std::string& platform, bool is_input)
+{
+  if ((platform != kTensorRTPlanPlatform) && io.is_non_linear_format_io()) {
```
tanmayv25, line 1719: Add this in the beginning to drop repeated checks for whether io.is_non_linear_format_io() below.
```
if(!io.is_non_linear_format_io()) {
   // Nothing to validate as the tensor is not non-linear format. 
   return Status::Success;
}
```

tanmayv25 (dismissed):

src/infer_request.cc
pskiran1: @tanmayv25, I made the change above but had to undo it because we encountered an error when using the model config with batch size enabled ([shape mismatch due to batch size not being included](https://github.com/triton-inference-server/tensorrt_backend/blob/main/src/tensorrt_utils.cc#L374C3-L382C4)). We also may encounter issues in other areas while using `ShapeWithBatchDim()` for shape tensors. Can we proceed with the current change?
Thank you.

tanmayv25 (approved):

PR #382, opened 2024-07-23
refactor: Refactor core input size checks

yinggeh:
#### What does the PR do?
- Refactor core input size checks.
- More descriptive error message.
- Early stop in BYTES inputs check if more strings than expected.
- Fix bug for undesired error when INT_MAX < input_byte_size <= LLONG_MAX.

#### Checklist

#### Commit Type:
- [x] refactor

#### Related PRs:
https://github.com/triton-inference-server/server/pull/7592

n/a

#### Test plan:
L0_input_validation--base

- CI Pipeline ID:
18114270

#### Caveats:
n/a

#### Background
n/a

n/a

### DIFF
--- a/src/test/input_byte_size_test.cc
+++ b/src/test/input_byte_size_test.cc
@@ -261,7 +261,7 @@ TEST_F(InputByteSizeTest, ValidInputByteSize)
   // Create an inference request
   FAIL_TEST_IF_ERR(
       TRITONSERVER_InferenceRequestNew(
-          &irequest_, server_, "pt_identity", -1 /* model_version */),
+          &irequest_, server_, "simple", -1 /* model_version */),
       "creating inference request");
   FAIL_TEST_IF_ERR(
       TRITONSERVER_InferenceRequestSetReleaseCallback(
@@ -269,8 +269,8 @@ TEST_F(InputByteSizeTest, ValidInputByteSize)
       "setting request release callback");
 
   // Define input shape and data
-  std::vector<int64_t> shape{1, 8};
-  std::vector<float> input_data(8, 1);
+  std::vector<int64_t> shape{1, 16};
+  std::vector<float> input_data(16, 1);
   const auto input0_byte_size = sizeof(input_data[0]) * input_data.size();
 
   // Set input for the request
@@ -315,7 +315,7 @@ TEST_F(InputByteSizeTest, InputByteSizeMismatch)
   // Create an inference request
   FAIL_TEST_IF_ERR(
       TRITONSERVER_InferenceRequestNew(
-          &irequest_, server_, "pt_identity", -1 /* model_version */),
+          &irequest_, server_, "simple", -1 /* model_version */),
       "creating inference request");
   FAIL_TEST_IF_ERR(
       TRITONSERVER_InferenceRequestSetReleaseCallback(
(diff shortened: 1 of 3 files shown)


src/test/input_byte_size_test.cc
```diff
@@ -629,7 +643,8 @@ TEST_F(InputByteSizeTest, StringCountMismatchGPU)
   // Create an inference request
   FAIL_TEST_IF_ERR(
       TRITONSERVER_InferenceRequestNew(
-          &irequest_, server_, "simple_identity", -1 /* model_version */),
+          &irequest_, server_, "savedmodel_zero_1_object",
```
rmccorm4, line 646: What was the purpose of changing the model used here?

rmccorm4 (approved):
LGTM, but left a question out of curiosity

src/test/input_byte_size_test.cc
yinggeh: I changed non-string tests to use `qa_identity_model_repository/savedmodel_zero_1_float32` because it has fewer inputs than `docs/examples/model_repository/simple`. Updated `docs/examples/model_repository/simple_identity` to `qa_identity_model_repository/savedmodel_zero_1_object` just to unify their source location.

src/test/input_byte_size_test.cc
```diff
@@ -258,19 +258,20 @@ char InputByteSizeTest::input_data_string_
 
 TEST_F(InputByteSizeTest, ValidInputByteSize)
 {
+  const char* model_name = "savedmodel_zero_1_float32";
```
tanmayv25, line 261: I think we are now using the same model for all our tests below. Can we add model_name_ as class member in L245-252? This way we can use the model_name_ everywhere here and below.
It also shows our intention using the same model.

src/test/input_byte_size_test.cc
```diff
@@ -312,19 +313,20 @@ TEST_F(InputByteSizeTest, ValidInputByteSize)
...
       "creating inference request");
   FAIL_TEST_IF_ERR(
       TRITONSERVER_InferenceRequestSetReleaseCallback(
           irequest_, InferRequestComplete, nullptr /* request_release_userp */),
       "setting request release callback");
 
   // Define input shape and data
-  std::vector<int64_t> shape{1, 8};
-  std::vector<float> input_data(10, 1);
+  std::vector<int64_t> shape{1, 16};
```
tanmayv25, line 328: Why changing these values?

src/test/input_byte_size_test.cc
```diff
@@ -258,19 +258,20 @@ char InputByteSizeTest::input_data_string_
...
       "creating inference request");
   FAIL_TEST_IF_ERR(
       TRITONSERVER_InferenceRequestSetReleaseCallback(
           irequest_, InferRequestComplete, nullptr /* request_release_userp */),
       "setting request release callback");
 
   // Define input shape and data
-  std::vector<int64_t> shape{1, 8};
-  std::vector<float> input_data(8, 1);
+  std::vector<int64_t> shape{1, 16};
```
tanmayv25, line 273: Why changing these values?

tanmayv25 (commented):
This looks more like a refactor than addition of  client side byte size checks, Can you update the PR title to reflect the changes more accurately?

src/test/input_byte_size_test.cc
yinggeh: Not really. String input tests use savedmodel_zero_1_object instead.

src/test/input_byte_size_test.cc
yinggeh: I don't remember. Might just want to be consistent with some other tests I saw. Either 8 or 16 is fine.

src/test/input_byte_size_test.cc
yinggeh: Same as above.

rmccorm4 (approved):
Thanks for adding the `LLONG_MAX` fix as well :rocket:

PR #385, opened 2024-08-05
chore: Expose `TRITONSERVER_Server` instance in Python In-Process API 

KrishnanPrash:
Goal: To be able to start frontends in python, given that the server instance is being created with the Python In-Process API (`tritonserver`).

By exposing the c++ pointer to a running `TRITONSERVER_Server` instance, we can retrieve the memory address of the instance in python and pass it to the `tritonfrontend` python package. This package will then use `reinterepet_cast<...>` and recover the `TRITONSERVER_Server` object from the memory address given.


GuanLuo (dismissed):

python/tritonserver/_api/_server.py
```diff
@@ -621,6 +621,9 @@ def stop(self) -> None:
         self._server.stop()
         self._server = Server._UnstartedServer()
 
+    def get_c_ptr(self):
```
nnshah1, line 624: question: 

1) can we make this a private method: 
```_ptr(self):```

2) Is there a way to pass the pybind object as an opaque pointer to the other components instead of accessing the pointer from python?

File: python/tritonserver/_c/tritonserver_pybind.cc [hunk 2]
```diff
@@ -1434,6 +1434,11 @@ class PyServer : public PyWrapper<struct TRITONSERVER_Server> {
     owned_ = true;
   }
 
+  uintptr_t get_c_ptr()
```
nnshah1, line 1437:
```suggestion
  uintptr_t _ptr()
```
Would like to avoid ```get``` to keep it parallel with other APIs. Do we need this API or could the caller do the cast?

nnshah1:
@KrishnanPrash , @GuanLuo , @rmccorm4  - 

Two points:

1) Instead of passing the underlying TRITONSERVER object between modules - could we pass the PyServer object? Reasoning: Can avoid directly exposing the TRITONSERVER ptr and leverage PyBind / ctypes to access the pointer to PyServer. If this is not possible / advisable due to ABI compat - would just like confirmation to understand. 

2) If we need to expose the ptr, let's make it private with a leading underscore and also remove the `get_c` - as we don't use get / set generally.

KrishnanPrash:
@nnshah1 
1. To my understanding, passing the `PyServer` object would not be sufficient, because `tritonfrontend` (and by extension server repo) does not have access to the `PyServer` implementation and therefore cannot recover the underlying pointer to the `TRITONSERVER_SERVER` object.
2. Thank you for the suggestion, the new function name should be just `_ptr()`.

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -1434,6 +1434,11 @@ class PyServer : public PyWrapper<struct TRITONSERVER_Server> {
     owned_ = true;
   }
 
+  uintptr_t _ptr()
+  {  // Calls PyWrapper
+    return reinterpret_cast<uintptr_t>(this->Ptr());
```
GuanLuo, line 1439: Why define new function.. `Ptr()` is inherited..

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -2107,6 +2112,7 @@ PYBIND11_MODULE(triton_bindings, m)
       .export_values();
   py::class_<PyServer>(m, "TRITONSERVER_Server")
       .def(py::init<PyServerOptions&>())
+      .def("_ptr", &PyServer::_ptr)
```
GuanLuo, line 2115:
```suggestion
      .def("_ptr", &PyServer::Ptr)
```

.gitignore
```diff
@@ -1,3 +1,4 @@
 /build
 /.vscode
 *.so
+*__pycache__/
```
GuanLuo, line 4:
```suggestion
*__pycache__/

```

python/tritonserver/_c/tritonserver_pybind.cc
rmccorm4: This change breaks the build, because now `_ptr` returns an opaque `TRITONSERVER_Server*` through the `PyWrapper<>` instead of the `uintptr_t` it was returning before.

```
...
server/build/_deps/pybind11-src/include/pybind11/detail/descr.h:40:19: error: invalid use of incomplete type 'struct TRITONSERVER_Server'
```

nnshah1: is there a way to cast within the pybind definition itself?

GuanLuo: Does above work or not? This PR has this suggestion applied..

GuanLuo: If not, just add a lambda wrapper over it.

KrishnanPrash: It does not work, @rmccorm4 thank you for assisting in debugging. And thank you for suggestion @GuanLuo , will go ahead and add a lambda wrapper.

KrishnanPrash: Actually there are two approaches here:
1) If we write a lambda function, then, the binded function would be static require a parameter and would have to look like:
```.def("_ptr", &[](PyServer* obj){ reinterpret_cast<uintptr_t>(obj->Ptr();)})```. 
In `core/python/_api/_server.py`, we would have to do `Server._ptr(server)`.

2) Instead of this, would it make more sense to keep the original flow of 
```cpp
class PyServer : ... {
     ...
  uintptr_t _ptr() {  
    return reinterpret_cast<uintptr_t>(this->Ptr());
  } 
     ...
}
...
.def("_ptr", &PyServer::_ptr)
```
In `core/python/_api/_server.py`, we would have to do `self.server._ptr()`.

Both options will keep the same external-flow, but was wondering if you had a preference either way for 1 or 2? @GuanLuo

GuanLuo: Option 1 please, and you don't have to call `Server._ptr(server)` https://github.com/pybind/pybind11/issues/1763#issuecomment-752683926

GuanLuo (approved):

rmccorm4 (approved):

PR #386, opened 2024-08-07
feat: Add histogram metric type

yinggeh:
#### What does the PR do?
Support histogram metric type and add tests.

#### Checklist

#### Commit Type:
- [x] feat

#### Related PRs:
https://github.com/triton-inference-server/vllm_backend/pull/56
https://github.com/triton-inference-server/python_backend/pull/374
https://github.com/triton-inference-server/server/pull/7525

n/a

#### Test plan:
n/a

- CI Pipeline ID:
17487728

#### Caveats:
n/a

#### Background
Customer requested histogram metrics from vLLM.

n/a


include/triton/core/tritonserver.h
```diff
@@ -2655,11 +2656,14 @@ TRITONSERVER_MetricFamilyDelete(struct TRITONSERVER_MetricFamily* family);
...
 /// \param label_count The number of labels.
+/// \param buckets Monotonically increasing values representing the
+/// bucket boundaries. For histogram only.
 /// \return a TRITONSERVER_Error indicating success or failure.
 TRITONSERVER_DECLSPEC struct TRITONSERVER_Error* TRITONSERVER_MetricNew(
     struct TRITONSERVER_Metric** metric,
     struct TRITONSERVER_MetricFamily* family,
-    const struct TRITONSERVER_Parameter** labels, const uint64_t label_count);
+    const struct TRITONSERVER_Parameter** labels, const uint64_t label_count,
+    const void* buckets = nullptr);
```
rmccorm4, line 2666: We can not modify an existing C API function definition as it will break ABI compatibility. We need a new API instead, such as `TRITONSERVER_MetricNewV2` (name can be TBD).

Given this, the new C API should also aim to be more extensible to account for some of the shortcomings of the current `MetricNew` that necessitated this change.

For example, by introducing a configurable opaque object `TRITONSERVER_MetricSettings` that we can append new options as needed over time without breaking ABI compatibility:
```cpp
struct TRITONSERVER_MetricSettings;

// ...

TRITONSERVER_DECLSPEC struct TRITONSERVER_Error* TRITONSERVER_MetricNewV2(
    struct TRITONSERVER_Metric** metric,
    struct TRITONSERVER_MetricFamily* family,
    const struct TRITONSERVER_Parameter** labels, const uint64_t label_count,
    const struct TRITONSERVER_MetricSettings* settings);
```

`Histogram` will require `buckets` today.

`Summary` will require `quantiles`, and possibly some settings like `age` to account for update frequency, which we don't want to need yet another new C API for later on when we look at implementing this support.

So the `MetricSettings` object should be used to support adding new fields over time without breaking the API.

For easier maintenance, the C++ implementation of `TRITONSERVER_MetricNew` (for counter/gauge) internally can be updated to call `TRITONSERVER_MetricNewV2` with `settings = nullptr`.

rmccorm4: CC @GuanLuo @Tabrizian @oandreeva-nv for second opinion on new C API requirements

include/triton/core/tritonserver.h
```diff
@@ -2705,6 +2710,28 @@ TRITONSERVER_DECLSPEC struct TRITONSERVER_Error* TRITONSERVER_MetricIncrement(
...
+/// Collect metrics.
+/// Supports metrics of kind TRITONSERVER_METRIC_KIND_COUNTER,
+/// TRITONSERVER_METRIC_KIND_GAUGE, TRITONSERVER_METRIC_KIND_HISTOGRAM and
+/// returns TRITONSERVER_ERROR_UNSUPPORTED for unsupported
+/// TRITONSERVER_MetricKind.
+///
+/// \param metric The metric object to collect.
+/// \param value Returns the current value of the metric object.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC struct TRITONSERVER_Error* TRITONSERVER_MetricCollect(
```
GuanLuo, line 2732: Why this API is needed?

include/triton/core/tritonserver.h
```diff
@@ -2705,6 +2710,28 @@ TRITONSERVER_DECLSPEC struct TRITONSERVER_Error* TRITONSERVER_MetricIncrement(
...
     struct TRITONSERVER_Metric* metric, double value);
 
+/// Observe the current value of metric to value.
+/// Supports metrics of kind TRITONSERVER_METRIC_KIND_HISTOGRAM and returns
+/// TRITONSERVER_ERROR_UNSUPPORTED for unsupported TRITONSERVER_MetricKind.
+///
+/// \param metric The metric object to update.
+/// \param value The amount to observe metric's value to.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC struct TRITONSERVER_Error* TRITONSERVER_MetricObserve(
```
GuanLuo, line 2720: Reuse `TRITONSERVER_MetricSet`?

include/triton/core/tritonserver.h
GuanLuo: I agree

src/test/metrics_api_test.cc
```diff
@@ -232,6 +233,32 @@ MetricAPIHelper(TRITONSERVER_Metric* metric, TRITONSERVER_MetricKind kind)
...
+  std::vector<std::uint64_t> cumulative_counts = {1, 1, 2, 2, 3, 3};
+  double sum = 0.0;
+  for (auto datum : data) {
+    FAIL_TEST_IF_ERR(
+        TRITONSERVER_MetricObserve(metric, datum), "observe metric value");
+    sum += datum;
+  }
+
+  // Collect
+  prometheus::ClientMetric value;
```
GuanLuo, line 250: Leaking implementation..

include/triton/core/tritonserver.h
oandreeva-nv: agree

include/triton/core/tritonserver.h
yinggeh: Used in metrics_api_test.py to return values from a histogram metric.

include/triton/core/tritonserver.h
yinggeh: @rmccorm4 Any thought? Basically we need to merge two funcions below into one `Metric::Set(double value)`. It works but may add confusion.
https://github.com/triton-inference-server/core/blob/fd5c44bae3fb4eaef8b0ac277d68142a8eb11e03/src/metric_family.cc#L338-L404

src/test/metrics_api_test.cc
yinggeh: Suggestions?

GuanLuo: Remove the collect function and check the metrics by parsing from `TRITONSERVER_MetricsFormatted`

include/triton/core/tritonserver.h
rmccorm4: I can take a closer look tomorrow, but my gut reaction is that Guan is probably right and we can probably just reuse Get and Set which will call Collect and Observe internally when `kind == KIND_HISTOGRAM` if functionally equivalent

rmccorm4: `MetricValue` may not work if `Collect` returns multiple values (one per bucket?), but again will need to take a closer look. Let me know if you already know more details on this from your research.

yinggeh: `MetricValue` cannot be reused for histogram.

include/triton/core/tritonserver.h
Tabrizian: Agreed, one minor suggestion, can we name it `TRITONSERVER_MetricNewWithSettings` instead of v2? I think this is more consistent with what we've been doing with host policy etc.

include/triton/core/tritonserver.h
Tabrizian: Can we use `TRITONSERVER_MetricValue`?

yinggeh: No because histogram stores many values.

include/triton/core/tritonserver.h
yinggeh: Updated

src/test/metrics_api_test.cc
yinggeh: Updated

include/triton/core/tritonserver.h
yinggeh: I named them `TRITONSERVER_MetricNewWithArgs` and `TRITONSERVER_MetricArgs` respectively before reading your comment. We can revise the namings if needed.

include/triton/core/tritonserver.h
```diff
@@ -2644,6 +2646,42 @@ TRITONSERVER_DECLSPEC struct TRITONSERVER_Error* TRITONSERVER_MetricFamilyNew(
 TRITONSERVER_DECLSPEC struct TRITONSERVER_Error*
 TRITONSERVER_MetricFamilyDelete(struct TRITONSERVER_MetricFamily* family);
 
+/// Get the TRITONSERVER_MetricKind of the metric family.
+///
+/// \param metric The metric family object to query.
+/// \param kind Returns the TRITONSERVER_MetricKind of metric.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC struct TRITONSERVER_Error*
+TRITONSERVER_GetMetricFamilyKind(
```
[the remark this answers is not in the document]
yinggeh: There exists `TRITONSERVER_GetMetricKind` API. Shouldn't we be consistent with naming?

include/triton/core/tritonserver.h
```diff
@@ -2644,6 +2646,42 @@ TRITONSERVER_DECLSPEC struct TRITONSERVER_Error* TRITONSERVER_MetricFamilyNew(
...
+/// Create a new metric args object. The caller takes ownership of the
+/// TRITONSERVER_MetricArgs object and must call TRITONSERVER_MetricArgsDelete
+/// to release the object.
+///
+/// \param args Returns the new metric args object.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_DECLSPEC struct TRITONSERVER_Error* TRITONSERVER_MetricArgsNew(
+    struct TRITONSERVER_MetricArgs** args);
+
+/// Set metric args with prometheus histogram metric parameter.
```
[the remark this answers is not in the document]
yinggeh: What is the concern here?

include/triton/core/tritonserver.h
```diff
@@ -2655,12 +2693,35 @@ TRITONSERVER_MetricFamilyDelete(struct TRITONSERVER_MetricFamily* family);
 /// \param family The metric family to add this new metric to.
 /// \param labels The array of labels to associate with this new metric.
 /// \param label_count The number of labels.
+/// bucket boundaries. For histogram only.
```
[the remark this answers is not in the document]
yinggeh: Nice catch

include/triton/core/tritonserver.h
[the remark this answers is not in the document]
Tabrizian: Should be fine.

include/triton/core/tritonserver.h
Tabrizian: Should be fine.

include/triton/core/tritonserver.h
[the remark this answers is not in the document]
Tabrizian: Prometheus is an internal implementation and we should not be leaking this information to the external user.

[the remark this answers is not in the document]
yinggeh: Updated

include/triton/core/tritonserver.h
yinggeh: I would like to revisit this change. The consensus is to keep C API and python_backend API 1:1 matched. I am inclined to add a new C API `TRITONSERVER_MetricObserve` for histogram instead of reusing `TRITONSERVER_MetricSet` for three reasons.
1. Both `Histogram` and `Summary` types call `Observe` to record new value. We can reuse observe for `Summary` type if we add it in the future.
2. Histogram also has `ObserveMultiple` API which may be added in the future. I don't like the idea that `Histogram.Set` and `Histogram.ObserveMultiple` coexist.
3. Setting histogram to a value aka Histogram.set(val) is semantically wrong. It is confusing to users familiar with Prometheus APIs. The description of `TRITONSERVER_MetricSet` can be verbose as well in order to describe different behaviors for counter/gauge and histogram/summary.
cc @Tabrizian @rmccorm4 @GuanLuo

yinggeh: FYI
https://github.com/jupp0r/prometheus-cpp/blob/master/core/include/prometheus/histogram.h
https://github.com/jupp0r/prometheus-cpp/blob/master/core/include/prometheus/summary.h

GuanLuo: Well.. Triton metrics API is not supposed to be mirroring "Prometheus API", Prometheus is one of the "forms" we can exhibit the metrics as. So we should design the API to be using generic terms for statistics, the meaning of `gauge`/`counter`/`histogram` (and `summary`?) is not affected by the fact that Prometheus or other metrics libraries are used.

Thinking from this mindset, my question is if `observe` is the generic verb for recording the statistics of `histogram`. If that is the case, then I am fine to add `XXXObserve`, otherwise, we should use the proper verb

src/metric_family.h
```diff
@@ -37,6 +39,30 @@
...
+//
+class TritonServerMetricArgs {
+ public:
+  TritonServerMetricArgs() = default;
+
+  void* SetHistogramArgs(const double* buckets, uint64_t bucket_count)
+  {
+    kind_ = TRITONSERVER_METRIC_KIND_HISTOGRAM;
+    buckets_.resize(bucket_count);
+    std::memcpy(buckets_.data(), buckets, sizeof(double) * bucket_count);
```
GuanLuo, line 55:
```suggestion
    buckets_ = std::vector<double>(buckets, buckets + bucket_count);
```

src/test/metrics_api_test.cc
```diff
@@ -232,6 +233,64 @@ MetricAPIHelper(TRITONSERVER_Metric* metric, TRITONSERVER_MetricKind kind)
...
+  // Observe data
+  std::vector<double> data{0.05, 1.5, 6.0};
+  double sum = 0.0;
+  for (auto datum : data) {
+    FAIL_TEST_IF_ERR(
+        TRITONSERVER_MetricObserve(metric, datum), "observe metric value");
+    sum += datum;
+  }
+  std::vector<std::uint64_t> cumulative_counts = {1, 1, 2, 2, 3, 3};
+  ASSERT_EQ(buckets.size() + 1, cumulative_counts.size());
```
GuanLuo, line 251: The `cumulative_counts` is depending on the `buckets` you split for the histogram, you should initialize `cumulative_counts` according to the `buckets` and `data`

include/triton/core/tritonserver.h
yinggeh: I think either sample or observe. Voting for Observe for simplicity.
> A histogram samples observations (usually things like request durations or response sizes) and counts them in configurable buckets. It also provides a sum of all observed values.
> 
> Similar to a histogram, a summary samples observations (usually things like request durations and response sizes). While it also provides a total count of observations and a sum of all observed values, it calculates configurable quantiles over a sliding time window.

src/metric_family.h
yinggeh: Updated

src/test/metrics_api_test.cc
yinggeh: Updated

src/metric_family.h
yinggeh: Updated

include/triton/core/tritonserver.h
```diff
@@ -2695,8 +2757,9 @@ TRITONSERVER_DECLSPEC struct TRITONSERVER_Error* TRITONSERVER_MetricValue(
 TRITONSERVER_DECLSPEC struct TRITONSERVER_Error* TRITONSERVER_MetricIncrement(
     struct TRITONSERVER_Metric* metric, double value);
 
-/// Set the current value of metric to value.
-/// Supports metrics of kind TRITONSERVER_METRIC_KIND_GAUGE and returns
+/// Set the current value of metric to value or observe the value to metric.
+/// Supports metrics of kind TRITONSERVER_METRIC_KIND_GAUGE and
+/// TRITONSERVER_METRIC_KIND_HISTOGRAM. Returns
```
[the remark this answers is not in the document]
yinggeh: I updated the comment. https://github.com/triton-inference-server/core/pull/386/files#diff-ce70b44ec4760c05dc91fc32ae46c5c3d9fa27edb61b960bf299b5960b38fc9fR2770

GuanLuo (dismissed):

GuanLuo (approved):

PR #388, opened 2024-08-14
feat: Load new model version should not reload loaded existing model version(s)

kthui:
#### What does the PR do?

It is requested that any model version(s) already loaded should not be reloaded upon adding and loading new version(s), if the already loaded version(s) are not modified. This is a small optimization to the model load/unload logic to avoid reloading unchanged model version(s) unload a load request.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:

https://github.com/triton-inference-server/server/pull/7527

Start with the changes to model_lifecycle to see why the small twist to the logic enables the previously reloaded model version to be updated, and then move towards model_config_utils and model_repository_manager to see how those changes support the decision to update vs reload.

#### Test plan:

New tests on not reloading already loaded model version upon loading other model versions is added. See the tests on server PR.

- CI Pipeline ID: 17511909  17650341

#### Caveats:

N/A

#### Background

Previously, the model (all versions) will always be reloaded if there is a change in the model directory beyond the model config. This will cause unnecessary reload of unmodified model version(s), which this change makes model directory change detection more granular and decides if a model version should be reloaded or updated based on the model file(s) of the version.

N/A

### DIFF
--- a/src/model_config_utils.cc
+++ b/src/model_config_utils.cc
@@ -2432,13 +2432,15 @@ TritonToDataType(const TRITONSERVER_DataType dtype)
 }
 
 bool
-EquivalentInNonInstanceGroupConfig(
+EquivalentInNonInstanceGroupAndNonVersionPolicyConfig(
     const inference::ModelConfig& old_config,
     const inference::ModelConfig& new_config)
 {
   ::google::protobuf::util::MessageDifferencer pb_diff;
   pb_diff.IgnoreField(
       old_config.descriptor()->FindFieldByLowercaseName("instance_group"));
+  pb_diff.IgnoreField(
+      old_config.descriptor()->FindFieldByLowercaseName("version_policy"));
   return pb_diff.Compare(old_config, new_config);
 }
 
--- a/src/model_config_utils.cc
+++ b/src/model_config_utils.cc
@@ -2432,7 +2432,7 @@ TritonToDataType(const TRITONSERVER_DataType dtype)
 }
 
 bool
-EquivalentInNonInstanceGroupAndNonVersionPolicyConfig(
+ConfigChangeRequiresReload(
     const inference::ModelConfig& old_config,
     const inference::ModelConfig& new_config)
 {
@@ -2441,7 +2441,7 @@ EquivalentInNonInstanceGroupAndNonVersionPolicyConfig(
       old_config.descriptor()->FindFieldByLowercaseName("instance_group"));
   pb_diff.IgnoreField(
       old_config.descriptor()->FindFieldByLowercaseName("version_policy"));
-  return pb_diff.Compare(old_config, new_config);
+  return !pb_diff.Compare(old_config, new_config);
 }
 
 bool
--- a/src/model_repository_manager/model_repository_manager.cc
+++ b/src/model_repository_manager/model_repository_manager.cc
@@ -276,9 +276,9 @@ GetPathModifiedTime(const std::string& path)
   bool path_is_dir;
   Status status = IsDirectory(path, &path_is_dir);
   if (!status.IsOk()) {
-    LOG_ERROR << "Failed to determine modification time for '" << path
-              << "': " << status.AsString();
-    return 0;
+    throw std::runtime_error(
+        std::string("Failed to determine modification time for '") + path +
+        "': " + status.AsString());
   }
 
   // If 'path' is a file return its mtime. Otherwise, using the modification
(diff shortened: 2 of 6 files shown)


src/model_config_utils.cc
```diff
@@ -2432,13 +2432,15 @@ TritonToDataType(const TRITONSERVER_DataType dtype)
 }
 
 bool
-EquivalentInNonInstanceGroupConfig(
+EquivalentInNonInstanceGroupAndNonVersionPolicyConfig(
```
nnshah1, line 2435: do we need to be so precise in the name - could we make it more purpose driven:

```EquivalentConfig```

```Is_Reload_Required```

Comment (kthui) [reply src/model_config_utils.cc:2435]: yes, the second one sounds good. It is comparing the model config between the old and the new and decides if the model version needs to be reloaded to account for the change in model config.

Comment (kthui) [reply src/model_config_utils.cc]: updated: [Rename model config util config change require reload function](https://github.com/triton-inference-server/core/pull/388/commits/e6da04c1641636937a42ebbec71c25f81ae4876e)

File: src/model_repository_manager/model_repository_manager.cc [hunk 2]
```diff
@@ -326,64 +323,131 @@ GetDetailedModifiedTime(
...
-  for (const auto& child : contents) {
-    const auto full_path = JoinPath({model_dir_path, child});
-    if (full_path == model_config_path) {
-      // config.pbtxt or customized config file in configs folder
-      mtime.first = GetModifiedTime(full_path);
-    } else {
-      // model files
-      mtime.second = std::max(mtime.second, GetModifiedTime(full_path));
+    model_timestamps_.clear();
+    return;
```
GuanLuo, line 351: raise exception and try catch?

src/model_repository_manager/model_repository_manager.cc
```diff
@@ -326,64 +323,131 @@ GetDetailedModifiedTime(
...
+      if (!model_config_content_name_.empty()) {
+        LOG_ERROR << "Failed to determine modification time for '"
+                  << model_dir_path << "': Duplicate model config is detected";
+        model_timestamps_.clear();
+        model_config_content_name_.clear();
+        return;
+      }
+      model_config_content_name_ = content_name;
     }
+    model_timestamps_.emplace(content_name, GetPathModifiedTime(content_path));
```
GuanLuo, line 366: Break constructor content down into multiple helper functions?

GuanLuo (commented):
Overall LGTM, have a few comments.

src/model_repository_manager/model_repository_manager.cc
kthui: Did a quick test on L0_lifecycle, the test currently expects any error related to retrieving timestamp to remain hidden and simply set the failed timestamp to 0. To translate it into this `ModelTimestamp` object, setting timestamp to 0 is the same as initializing the object using the default constructor, which is leaving the `model_timestamps_` object empty. Thus, if an exception is raised here, the caller catching the exception will simply re-initialize `ModelTimestamp` object using the default constructor (to remain aligned with the current logic of setting the timestamp to 0). In this case, raising an exception is redundant and can be handled by the constructor in one or two lines, or with a helper function specifically for keeping everything empty.

Another approach is we can change the test cases to accept the behavior change that failure to determine timestamp will fail the model load.

src/model_repository_manager/model_repository_manager.cc
kthui: updated: [Break constructor into multiple functions](https://github.com/triton-inference-server/core/pull/388/commits/c35fc4d9f108094df9edb269f473d819134ca20d)

src/model_repository_manager/model_repository_manager.cc
kthui: I think to get this feature out quickly, we can retain the existing error handling on reading timestamps, and file another ticket for enhancing the error handling at a later time. @GuanLuo @rmccorm4 what do you think?

rmccorm4: Retaining existing error behavior and enhancing in a follow-up sounds fine to me.

kthui: Sure, added the comment: [Comment on should raise an exception when failed to create timestamp](https://github.com/triton-inference-server/core/pull/388/commits/f73ab61dadc5b1f89ca5a5c3f41cd9fbb5db7db9)

GuanLuo: Can be addressed as follow-up, but I do think an exception should be raised, and it is fine to have the caller catch the exception and then initialize a timestamp with default constructor (if not failing). This is from the stand point where if the user initializes a timestamp with a path, the user is expecting the returned timestamp to carry correct timestamp w.r.t. the given path, not 0. And thus an error should be raised programmatically and it is up to the user to decide how to handle the error.

GuanLuo (approved):

rmccorm4 (approved):

PR #390, opened 2024-08-20
feat: Load new model version should not reload loaded existing model …

kthui:
…version(s) (#388)

* Do not reload unmodified loaded model version

* Track model directory timestamps more granularly to detect updates to model version files

* Rename model config util config change require reload function

* Re-organize ModelTimestamp() and throw exception

* Revert "Re-organize ModelTimestamp() and throw exception"

This reverts commit 41e57e55cf55c3f0496fd95735e06e4764cf5659.

* Break constructor into multiple functions

* Comment on should raise an exception when failed to create timestamp

### DIFF
--- a/src/model_config_utils.cc
+++ b/src/model_config_utils.cc
@@ -2432,14 +2432,16 @@ TritonToDataType(const TRITONSERVER_DataType dtype)
 }
 
 bool
-EquivalentInNonInstanceGroupConfig(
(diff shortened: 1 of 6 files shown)


kthui:
Pick https://github.com/triton-inference-server/core/pull/388 into r24.08

nv-kmcgill53 (approved):

PR #395, opened 2024-09-04
Changes to fix data types mismatches for armv7 build

rvroge:
We are running the server and core in a armv7 device.
Build for platform linux/arm/v7 failed because the code was using 64 bit size variables and we have 32 bit.

### DIFF
--- a/src/cache_entry.cc
+++ b/src/cache_entry.cc
@@ -172,7 +172,7 @@ CacheEntry::SerializeResponse(InferenceResponse* response, Buffer& buffer)
   position += sizeof(uint32_t);
 
   for (const auto& output : response->Outputs()) {
-    uint64_t packed_output_byte_size = 0;
+    size_t packed_output_byte_size = 0;
     RETURN_IF_ERROR(SerializeResponseOutput(
         output, base + position, &packed_output_byte_size));
     // 2. Then the packed buffer will hold pairs of (output_size, output_bytes)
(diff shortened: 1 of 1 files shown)


src/cache_entry.cc
```diff
@@ -255,7 +255,7 @@ CacheEntry::SerializeResponseOutput(
   // Pack everything into provided buffer
   uint64_t position = 0;
```
oandreeva-nv, line 256: out of curiosity, why this line doesn't cause issues

src/cache_entry.cc
```diff
@@ -255,7 +255,7 @@ CacheEntry::SerializeResponseOutput(
...
   // Total serialized output size
-  memcpy(buffer + position, &total_byte_size, sizeof(uint64_t));
+  memcpy(buffer + position, &total_byte_size, sizeof(size_t));
   position += sizeof(uint64_t);
```
oandreeva-nv, line 259: ditto

src/cache_entry.cc
rvroge: @oandreeva-nv you have a good point. I'll elaborate a bit more on this one. The real issue is with the following lines. SerializeResponseOutput is expecting a size_t pointer:
`CacheEntry::SerializeResponseOutput(
    const InferenceResponse::Output& output, Byte* buffer, size_t* output_size)`

For our platform there is a mismatch between these two types:
error: cannot convert ‘uint64_t*’ {aka ‘long long unsigned int*’} to ‘size_t*’ {aka ‘unsigned int*’}

PR #396, opened 2024-10-12
feat: Add histogram support and new histogram metric

yinggeh:
#### What does the PR do?
The PR adds tests to histogram metrics and new `nv_inference_first_response_histogram`.
1. Verify this metric is only created in decoupled models.
2. Tests `--metrics-config histogram_latencies=<bool>`.

#### Checklist
- [ ] PR title reflects the change and is of format `<commit_type>: <Title>`
- [ ] Changes are described in the pull request.
- [ ] Related issues are referenced.
- [ ] Verified copyright is correct on all changed files.

#### Commit Type:
- [ ] feat

#### Test plan:
L0_metrics--base

- CI Pipeline ID:
19270121

#### Caveats:
1. Neelay suggested adding backend_name to metric label. During the implementation, I found it wasn't easy to only add the backend label to new metric without changing all existing ones. @nnshah1
2. Customer asked for TTFT in millisecond. However, it is not consistent with microsecond we use in other metrics. Please advise.

#### Background
Standardizing Large Model Server Metrics in Kubernetes


src/infer_response.cc
```diff
@@ -38,14 +38,20 @@ namespace triton { namespace core {
 //
 Status
 InferenceResponseFactory::CreateResponse(
-    std::unique_ptr<InferenceResponse>* response) const
+    std::unique_ptr<InferenceResponse>* response)
 {
   response->reset(new InferenceResponse(
       model_, id_, allocator_, alloc_userp_, response_fn_, response_userp_,
-      response_delegator_));
+      response_delegator_, response_cnt_
```
rmccorm4, line 45: Pick one of `response_cnt` or `seq_num` and use it in both places - it's harder to follow the flow of this variable with the name difference

src/model.h
```diff
@@ -148,6 +148,11 @@ class Model {
     return config_.response_cache().enable();
   }
 
+  bool isDecoupled() const
```
rmccorm4, line 151: Upper camel case functions like the rest `IsDecoupled()`
```suggestion
  bool IsDecoupled() const
```

src/infer_response.cc
```diff
@@ -282,6 +305,24 @@ InferenceResponse::TraceOutputTensors(
 }
 #endif  // TRITON_ENABLE_TRACING
 
+#ifdef TRITON_ENABLE_METRICS
+void
+InferenceResponse::UpdateResponseMetrics() const
+{
+  if (model_ != nullptr && seq_num_ == 0) {
```
rmccorm4, line 312: General question - the response factory already has this function and a notion of which response is being processed (response stats index):
```cpp
  uint64_t GetAndIncrementResponseIndex() { return response_stats_index_++; };
```

Can this be used instead of the new logic you have added for tracking the response count?

rmccorm4: General context on the feature: https://github.com/triton-inference-server/core/pull/325

src/model.h
yinggeh: Oops typo

src/infer_response.cc
yinggeh: Personally I think it makes sense to use different names. ResponseFactory keeps track of `response_cnt` which increases monotonically and each `Response` keeps its `seq_num`.

src/infer_response.cc
yinggeh: I couldn't find any places where`GetAndIncrementResponseIndex` is called except in square backend. @kthui should have some better ideas.
https://github.com/search?q=org%3Atriton-inference-server+TRITONBACKEND_ModelInstanceReportResponseStatistics&type=code

kthui: If I understand correctly
```c++
std::atomic<uint64_t> response_cnt_;
```
means the same as
```c++
std::atomic<uint64_t> response_stats_index_;
```
?
In this case, they should be one variable.

kthui: Then it makes sense to "reuse"
```c++
uint64_t GetAndIncrementResponseIndex() { return response_stats_index_++; };
```
for the same purpose, but some renaming may be needed.

src/infer_response.cc
yinggeh: Updated in the new commit -- response_index_ and index_ respectively.

src/infer_response.cc
yinggeh: Reused `response_stats_index_` and renamed to `response_index_` in the latest commit. @rmccorm4

yinggeh: I take that back. It might affect Square Backend.

src/infer_response.cc
yinggeh: Reverted the change due to potential breakage in Square Backend.

src/infer_response.cc
```diff
@@ -72,10 +78,20 @@ InferenceResponse::InferenceResponse(
...
+    uint64_t seq_idx
+#ifdef TRITON_ENABLE_METRICS
+    ,
+    uint64_t infer_start_ns
+#endif  // TRITON_ENABLE_METRICS
+    )
     : model_(model), id_(id), allocator_(allocator), alloc_userp_(alloc_userp),
       response_fn_(response_fn), response_userp_(response_userp),
-      response_delegator_(delegator), null_response_(false)
+      response_delegator_(delegator), seq_idx_(seq_idx),
```
kthui, line 90: What happen if the backend gets 10 response objects from the response factory, and the send them in the reverse order they are obtained from the factory?

yinggeh: Updated

src/model.h
```diff
@@ -148,6 +148,11 @@ class Model {
     return config_.response_cache().enable();
   }
 
+  bool IsDecoupled() const
```
kthui, line 151:
```suggestion

  // Get whether the model is decoupled.
  bool IsDecoupled() const
```

kthui (approved):
LGTM! @rmccorm4 should also take a look before merging.

src/backend_model_instance.cc
rmccorm4, line 1: I notice the table has no TTFT metrics reported for `ensemble`, after speaking offline it looks like the inference was called on `tensorrt_llm` model directly. Please try doing an inference on `ensemble` or `tensorrt_llm_bls` and showing the resulting metrics.

rmccorm4: ex: `curl -X POST localhost:8000/v2/models/ensemble/generate -d '{"text_input": "What is ML?", "max_tokens": 50, "bad_words": "", "stop_words": "", "pad_id": 2, "end_id": 2}'`

src/metric_model_reporter.h
```diff
@@ -46,25 +46,30 @@ struct ModelIdentifier;
...
   bool latency_counters_enabled_ = true;
+  // Create and use Histograms for per-model latency related metrics
+  bool latency_histograms_enabled_ = true;
   // Create and use Summaries for per-model latency related metrics
   bool latency_summaries_enabled_ = false;
   // Quantiles used for any summary metrics. Each pair of values represents
   // { quantile, error }. For example, {0.90, 0.01} means to compute the
   // 90th percentile with 1% error on either side, so the approximate 90th
   // percentile value will be between the 89th and 91st percentiles.
+  prometheus::Histogram::BucketBoundaries buckets_ = {10, 100, 500, 1000};
```
rmccorm4, line 64: Move buckets before the Quantiles comment, and add a comment describing bucket - ex: these are latencies in milliseconds

src/metric_model_reporter.h
```diff
@@ -46,25 +46,30 @@ struct ModelIdentifier;
...
-  void ParseConfig(bool response_cache_enabled);
+  void ParseConfig(bool response_cache_enabled, bool is_decoupled);
   // Parses pairs of quantiles "quantile1:error1, quantile2:error2, ..."
   // and overwrites quantiles_ field if successful.
   prometheus::Summary::Quantiles ParseQuantiles(std::string options);
 
   // Create and use Counters for per-model latency related metrics
   bool latency_counters_enabled_ = true;
+  // Create and use Histograms for per-model latency related metrics
+  bool latency_histograms_enabled_ = true;
```
rmccorm4, line 57: I think we should default to `false` until we've done some light benchmarking to verify any performance impact - especially since it looks like we're planning to cherry-pick this and planning to add multiple histogram metrics here.

After some further testing and perf validation, I think we can change the default to be enabled later if we want.

src/infer_response.cc
```diff
@@ -214,6 +232,10 @@ InferenceResponse::Send(
       TRITONSERVER_TRACE_TENSOR_BACKEND_OUTPUT, "InferenceResponse Send");
 #endif  // TRITON_ENABLE_TRACING
 
+#ifdef TRITON_ENABLE_METRICS
+  response->UpdateResponseMetrics();
```
rmccorm4, line 236: For an initial merge with a default of histograms being disabled - I think this is fine to go ahead with if we need to cherry-pick. However, please take note of the following:

I think this is relatively on the hot path, possibly impacting latency, compared to our other inference metrics (`TRITONBACKEND_ReportStatistics`) which are generally reported  **after** response sending in backends (impacting throughput but not response latency).

You can find some perf numbers of each prometheus-cpp metric type at the bottom of the README here: https://github.com/jupp0r/prometheus-cpp

One individual observation for a single metric and a small number of buckets may not be impactful for one request, but as we scale up high concurrency, more metrics, more buckets, etc - this could present a noticeable latency impact. 

It would be good to do some light validation of overall latency before/after the feature via `genai-perf`. Especially for high concurrency and streaming many responses/tokens - as there can be some synchronization in interaction with the prometheus registry with many concurrent responses as well.

It would probably be advantageous to do the actual prometheus registry interaction after sending the response if possible, such as by only doing the bare minimum of determining if we should report metrics (check if first response and record latency), then using that information to report the metric after initiating response send.

src/infer_response.h
```diff
@@ -246,8 +263,14 @@ class InferenceResponse {
       const ResponseAllocator* allocator, void* alloc_userp,
       TRITONSERVER_InferenceResponseCompleteFn_t response_fn,
       void* response_userp,
-      const std::function<void(
-          std::unique_ptr<InferenceResponse>&&, const uint32_t)>& delegator);
+      const std::function<
+          void(std::unique_ptr<InferenceResponse>&&, const uint32_t)>& delegator
+#ifdef TRITON_ENABLE_METRICS
+      ,
+      std::shared_ptr<std::atomic<uint64_t>> responses_sent_,
```
rmccorm4, line 270: function arg shouldn't have `_` suffix in our current style - only member variables

rmccorm4, line 270:
```suggestion
      std::shared_ptr<std::atomic<uint64_t>> responses_sent,
```
function arg here shouldn't have _ suffix in our current style - only member variables

src/metric_model_reporter.cc
```diff
@@ -53,6 +54,10 @@ MetricReporterConfig::ParseConfig(bool response_cache_enabled)
       latency_counters_enabled_ = false;
     }
 
+    if (pair.first == "histogram_latencies" && pair.second == "false") {
+      latency_histograms_enabled_ = false;
+    }
```
rmccorm4, line 59: same comment about defaulting to disabled histograms for now

src/metric_model_reporter.cc
```diff
@@ -53,6 +54,10 @@ MetricReporterConfig::ParseConfig(bool response_cache_enabled)
       latency_counters_enabled_ = false;
     }
 
+    if (pair.first == "histogram_latencies" && pair.second == "true") {
```
rmccorm4, line 57: Just an FYI for the future that it would probably make sense to expand histogram support to the existing latency metrics (queue, compute, etc) similar to how the summary metrics work. But it's not in scope for this PR.

For example, the `--metrics-config histogram_latencies=true` should enable this in the future similar to summaries. This would address this ask: https://github.com/triton-inference-server/server/issues/7672

src/metric_model_reporter.h
yinggeh: Fixed

src/metric_model_reporter.cc
yinggeh: Yes. In fact, I am planning on expaniding the following latency metrics to histograms in 24.11.
- nv_inference_request_duration_us
- nv_inference_queue_duration_us

src/infer_response.h
yinggeh: Fixed

src/metric_model_reporter.h
yinggeh: Updated

src/metric_model_reporter.h
yinggeh: Updated

src/model.h
yinggeh: Fixed

src/infer_response.cc
yinggeh: Fixed

src/backend_model_instance.cc
yinggeh: Updated the examples in PR description.

src/infer_response.cc
```diff
@@ -282,6 +305,24 @@ InferenceResponse::TraceOutputTensors(
...
+{
+  // Report inference to first response duration.
+  if (model_ != nullptr && responses_sent_ != nullptr &&
+      responses_sent_->fetch_add(1, std::memory_order_relaxed) == 0) {
+    auto now_ns = std::chrono::duration_cast<std::chrono::nanoseconds>(
+                      std::chrono::steady_clock::now().time_since_epoch())
+                      .count();
+    if (auto reporter = model_->MetricReporter()) {
+      reporter->ObserveHistogram(
+          "first_response_histogram", (now_ns - infer_start_ns_) / 1000000);
```
rmccorm4, line 320: Let's use the defined constants, ex: https://github.com/triton-inference-server/core/blob/571c5dcee325a78cd8f02e2ee8ebb17e1df2ac75/src/constants.h#L90

yinggeh: Updated

rmccorm4:
Deferring to Jacky as I'll be OOO, but let's make sure the test cases give us confidence in the feature and that we're addressing the spirit of the use case that motivates the feature

kthui (approved):
LGTM! The [test cases](https://github.com/triton-inference-server/server/pull/7694) are giving us confidence in the feature and the feature is addressing the use case.

Please make sure the CI passes with the all final changes before merging.

Arnold1:
what version of triton inference server is the histogram available and how to enable it when starting the server?

yinggeh:
@Arnold1 Staring from 24.11. Histogram support in Triton core is limited with `Request to First Response Time` metric for decoupled model. In python backend models, you can defined your own custom metrics with more freedom.
https://github.com/triton-inference-server/server/blob/r24.11/docs/user_guide/metrics.md#histograms
https://github.com/triton-inference-server/python_backend?tab=readme-ov-file#custom-metrics

Arnold1:
hi @yinggeh how to enable a histogram? by default i dont see any histogram. i need to add a custom metric?

yinggeh:
@Arnold1 Do you mean you attempted to add custom histogram metric but did not see it?

yinggeh:
Hi @Arnold1. Thank you for your patience. Below is an example where we use Python backend custom metrics to expose vLLm metrics.
https://github.com/triton-inference-server/vllm_backend/blob/b41f716d15100dc7bcbea27ebea20906452dadf5/src/utils/metrics.py#L39-L161
You can try modifying this starter code with histograms https://github.com/triton-inference-server/python_backend/tree/main/examples/custom_metrics

PR #398, opened 2024-10-14
Support for Windows Long Paths

ashishk98:
Add support for windows long paths

### DIFF
--- a/src/filesystem/implementations/local.h
+++ b/src/filesystem/implementations/local.h
@@ -31,7 +31,7 @@
 #include <fstream>
 
 #include "../../constants.h"
-#include "common.h"
+#include "common.h"W
 
 namespace triton { namespace core {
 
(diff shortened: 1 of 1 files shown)


src/filesystem/implementations/local.h
```diff
@@ -60,12 +60,45 @@ class LocalFileSystem : public FileSystem {
   Status MakeTemporaryDirectory(
       std::string dir_path, std::string* temp_dir) override;
   Status DeletePath(const std::string& path) override;
+
+ private:
+  inline std::string getOSValidPath(const std::string& _path);
+  const char* kWindowsLongPathPrefix = "\\\\?\\";
```
fpetrini15, line 66:
```suggestion
  const char* windows_long_path_prefix_ = "\\\\?\\";
```
To match Triton member variable convention.

src/filesystem/implementations/local.h
```diff
@@ -60,12 +60,45 @@ class LocalFileSystem : public FileSystem {
   Status MakeTemporaryDirectory(
       std::string dir_path, std::string* temp_dir) override;
   Status DeletePath(const std::string& path) override;
+
+ private:
+  inline std::string getOSValidPath(const std::string& _path);
```
fpetrini15, line 65:
```suggestion
  inline std::string GetOSValidPath(const std::string& path);
```
To match triton syntax conventions

src/filesystem/implementations/local.h
```diff
@@ -60,12 +60,45 @@ class LocalFileSystem : public FileSystem {
...
+//!
+//! On Linux there is not much to do but make sure correct slashes are used
+//! On Windows we need to take care of the long paths and handle them correctly
+//! to avoid legacy issues with MAX_PATH
+//!
+//! More details:
+//! https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry
+//!
+inline std::string
+LocalFileSystem::getOSValidPath(const std::string& _path)
```
fpetrini15, line 80:
```suggestion
LocalFileSystem::getOSValidPath(const std::string& path)
```
Cascading effect where the local `path` variable will need to change to something like `local_path` or `l_path`

fpetrini15:
Great work @ashishk98, just a few minor stylistic comments to be consistent with Triton. Two things:

1. Can you port these changes to be a direct branch from `main` -- our CI is not built around cloning forked branches
2. Can you update the copyright at the top of the file to extend to `2024`?

src/filesystem/implementations/local.h
nicolasnoble: But that's a constant, not a variable? AFAIK, constants should be named like the former, not the latter.

nicolasnoble: It *might* be more idiomatic to use `static constexpr char* kWindowsLongPathPrefix = "\\\\?\\";` to drive the fact this is, indeed, a constant.

fpetrini15: Had to verify with the google style [guide](https://google.github.io/styleguide/cppguide.html#Constant_Names) and you're both correct. I also found an instance in Triton code where we initialize to a constant [here](https://github.com/triton-inference-server/server/blob/main/src/command_line_parser.h#L85), however, we omit the leading "k", which is technically wrong according to the style standards.

We can defer to Nico's suggestion instead. I will update my comment above to that effect.

PR #400, opened 2024-10-15
feat: Incorporate "additional-dependency-dirs"

fpetrini15:
#### What does the PR do?
Adds core support for a new "additional-dependency-dirs" configuration option that will add user-specified directories to the search path during backend loading.

#### Checklist
- [ ] Related issues are referenced.

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [] test

#### Related PRs:

Server: https://github.com/triton-inference-server/server/pull/7707

- CI Pipeline ID:

[WIP]

### DIFF
--- a/src/shared_library.cc
+++ b/src/shared_library.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2021-2024, NVIDIA CORPORATION. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
--- a/src/shared_library.cc
+++ b/src/shared_library.cc
@@ -254,7 +254,7 @@ SharedLibrary::AddAdditionalDependencyDir(
     original_path.resize(len);
     GetEnvironmentVariableW(PATH.c_str(), &original_path[0], len);
   } else {
-    return Status(Status::Code::INTERNAL, "PATH variable is empty");
+    original_path = L"";
   }
 
   LOG_VERBOSE(1) << "Environment before extending PATH: "
@@ -298,7 +298,7 @@ SharedLibrary::AddAdditionalDependencyDir(
 }
 
 Status
-SharedLibrary::RemoveAdditionalDependencyDir(std::wstring& original_path)
+SharedLibrary::RemoveAdditionalDependencyDir(const std::wstring& original_path)
 {
 #ifdef _WIN32
   const std::wstring PATH(L"Path");
--- a/src/shared_library.cc
+++ b/src/shared_library.cc
@@ -249,6 +249,13 @@ SharedLibrary::AddAdditionalDependencyDir(
 #ifdef _WIN32
   const std::wstring PATH(L"Path");
 
+  if (additional_path.back() != ';') {
+    return Status(
+        Status::Code::INVALID_ARG,
+        "backend config parameter \"additional-dependency-dirs\" is malformed. "
+        "Each additional path provided should terminate with a ';'.");
+  }
+
   DWORD len = GetEnvironmentVariableW(PATH.c_str(), NULL, 0);
   if (len > 0) {
     original_path.resize(len);
--- a/src/shared_library.cc
+++ b/src/shared_library.cc
@@ -256,7 +256,9 @@ SharedLibrary::AddAdditionalDependencyDir(
   } else {
     return Status(Status::Code::INTERNAL, "PATH variable is empty");
   }
-  std::wcout << "Before Add: " << original_path << std::endl;
+
+  LOG_VERBOSE(1) << "Environment before extending PATH: "
+                 << std::string(original_path.begin(), original_path.end());
 
   std::wstring updated_path_value =
       std::wstring(additional_path.begin(), additional_path.end());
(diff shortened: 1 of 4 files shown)


src/shared_library.cc
```diff
@@ -242,4 +242,92 @@ SharedLibrary::GetEntrypoint(
...
+    if (len > 0) {
+      path_after.resize(len);
+      GetEnvironmentVariableW(PATH.c_str(), &path_after[0], len);
+    }
+    LOG_VERBOSE(1) << "Environment after extending PATH: "
+                   << std::string(path_after.begin(), path_after.end());
+  }
+#else
+  LOG_WARNING
+      << "The parameter \"additional-dependency-dirs\" has been specified but "
```
oandreeva-nv, line 293: oh wow, it's the first time I see Windows-only feature

fpetrini15: Ultimate goal is to have parity, however, the paradigm we use to load models is sufficiently different between Windows and Linux that we aren't sure what form it will take.

src/shared_library.h
```diff
@@ -65,6 +65,14 @@ class SharedLibrary {
   Status GetEntrypoint(
       void* handle, const std::string& name, const bool optional, void** befn);
 
+  // Add an additional dependency directory to PATH (Windows-only).
+  Status AddAdditionalDependencyDir(
+      const std::string& additional_path, std::wstring& original_path);
+
+  // Restore PATH to its original configuration. Should be used in
+  // conjunction with AddAdditionalDependencyDir (Windows-only).
+  Status RemoveAdditionalDependencyDir(std::wstring& original_path);
```
GuanLuo, line 74: `const &`

src/shared_library.cc
```diff
@@ -242,4 +242,92 @@ SharedLibrary::GetEntrypoint(
...
+{
+#ifdef _WIN32
+  const std::wstring PATH(L"Path");
+
+  DWORD len = GetEnvironmentVariableW(PATH.c_str(), NULL, 0);
+  if (len > 0) {
+    original_path.resize(len);
+    GetEnvironmentVariableW(PATH.c_str(), &original_path[0], len);
+  } else {
+    return Status(Status::Code::INTERNAL, "PATH variable is empty");
```
GuanLuo, line 257: Probably true that `PATH` will always has value, but should we just continue in this case?

src/shared_library.cc
```diff
@@ -242,4 +242,92 @@ SharedLibrary::GetEntrypoint(
...
+  }
+
+  LOG_VERBOSE(1) << "Environment before extending PATH: "
+                 << std::string(original_path.begin(), original_path.end());
+
+  std::wstring updated_path_value =
+      std::wstring(additional_path.begin(), additional_path.end());
+  updated_path_value += original_path;
```
GuanLuo, line 265: You don't need to add delimiter for concatenation?

fpetrini15: The assumption I've made is that users will add a `;` to each path they add at the command line--much like they would adding any other path. However, I believe we can be ultra-conservative here and add a `;` regardless. I believe any extra `;` will be ignored anyways.

src/shared_library.cc
```diff
@@ -242,4 +242,92 @@ SharedLibrary::GetEntrypoint(
...
+#endif
+  return Status::Success;
+}
+
+Status
+SharedLibrary::RemoveAdditionalDependencyDir(std::wstring& original_path)
+{
+#ifdef _WIN32
+  const std::wstring PATH(L"Path");
+  if (!SetEnvironmentVariableW(PATH.c_str(), original_path.c_str())) {
```
mc-nv, line 305: What happen if `original_path` value will be `NULL`?

mc-nv:
I'm against managing environment variables out of the source code.  
It must go through the code scanning tool before merge, NSPECT, VERACODE etc.

src/shared_library.cc
fpetrini15: That wouldn't happen. In the calling function we initialize `original_path` with the default constructor--which will initialize it to an empty string. We check whether `original_path` is empty before calling this function and we do not set it to NULL anywhere in between.

mc-nv: Have you tried it with an empty string?

fpetrini15: I'm saying it wouldn't make it in here if it were empty: https://github.com/triton-inference-server/core/pull/400/files#diff-c3cafbdf0ba28838ed1f0c9407e4cb16f33bc18f8274b326114fc3a6bbbbbf81R203

mc-nv: I would like to see this scenario tested because of the definition in the method that been chosen. 
https://learn.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-setenvironmentvariablew
> If this parameter is NULL, the variable is deleted from the current process's environment.

fpetrini15: As an update, based on Guan's suggestion that we continue when the PATH variable is empty instead of returning an error, there is now a valid (albeit very rare) scenario where `original_path` can be empty. This is an scenario I was able to test locally and can confirm that passing an empty string will result in `Path=` ({empty string} as it would have been before), whereas, passing `NULL` is distinct and will completely clear `Path` from the list of variables. Therefore, my argument about never being able to pass `NULL`, the real concern here, still applies.

src/shared_library.cc
fpetrini15: Reflecting on this more: We can technically add `;` to help the user (though this would only help if they specified a single path). However, I don't want to help/encourage the user to use the extension improperly. I think the best thing to do here would probably be to throw an error if their input string does not end with `;`.

src/shared_library.h
```diff
@@ -65,6 +65,14 @@ class SharedLibrary {
   Status GetEntrypoint(
       void* handle, const std::string& name, const bool optional, void** befn);
 
+  // Add an additional dependency directory to PATH (Windows-only).
+  Status AddAdditionalDependencyDir(
+      const std::string& additional_path, std::wstring& original_path);
+
+  // Restore PATH to its original configuration. Should be used in
+  // conjunction with AddAdditionalDependencyDir (Windows-only).
```
GuanLuo, line 73: nit: remove `(Windows-only)`

GuanLuo (dismissed):

src/shared_library.h
fpetrini15: Even if these new functions are Windows-only atm?

GuanLuo: It is just a matter of taste. This interface is abstracting away the platform-specific which is why I don't think you need to call out Windows here. You can keep it to highlight the behavior difference but then you should detail on what is the behavior for non-windows platform.

src/shared_library.cc
```diff
@@ -242,4 +242,99 @@ SharedLibrary::GetEntrypoint(
...
+        "failed to restore PATH to its original configuration " + errstr);
+  }
+
+  if (LOG_VERBOSE_IS_ON(1)) {
+    std::wstring path_after;
+    DWORD len = GetEnvironmentVariableW(PATH.c_str(), NULL, 0);
+    if (len > 0) {
+      path_after.resize(len);
+      GetEnvironmentVariableW(PATH.c_str(), &path_after[0], len);
+    }
```
rmccorm4, line 332: Don't need to change in this PR, but looks like this pattern of 
- check size of env var
- resize local string
- write env var to string

is used at least 3 times - probably good to be a helper function in the future

rmccorm4 (approved):

GuanLuo (approved):

PR #405, opened 2024-10-29
feat: Allow per-model metric configuration

yinggeh:
#### What does the PR do?
Tests new model_metrics message in pbtxt. Add example to override histogram buckets per-family.

#### Checklist

#### Commit Type:
- [x] feat

#### Related PRs:
https://github.com/triton-inference-server/common/pull/126
https://github.com/triton-inference-server/server/pull/7752

#### Test plan:
L0_metrics

- CI Pipeline ID:
20106057

#### Background
Default histogram buckets does not satisfy all use cases.

### DIFF
--- a/src/metric_model_reporter.cc
+++ b/src/metric_model_reporter.cc
@@ -80,12 +80,13 @@ MetricReporterConfig::ParseConfig(
   for (const auto& metric_control : model_metrics.metric_control()) {
     const std::string& family_name =
         metric_control.metric_identifier().family();
-    // Copy protobuf RepeatedField to std::vector
-    const auto& buckets_proto = metric_control.histogram_options().buckets();
-    const prometheus::Histogram::BucketBoundaries buckets(
-        buckets_proto.begin(), buckets_proto.end());
 
+    // If family name exists, override with new options.
     if (metric_map_.find(family_name) != metric_map_.end()) {
+      // Copy protobuf RepeatedField to std::vector
+      const auto& buckets_proto = metric_control.histogram_options().buckets();
+      const prometheus::Histogram::BucketBoundaries buckets(
+          buckets_proto.begin(), buckets_proto.end());
       histogram_options_[metric_map_.at(family_name)] = buckets;
     }
   }
--- a/src/metric_model_reporter.cc
+++ b/src/metric_model_reporter.cc
@@ -42,7 +42,8 @@ namespace triton { namespace core {
 //
 void
 MetricReporterConfig::ParseConfig(
-    bool response_cache_enabled, bool is_decoupled)
+    bool response_cache_enabled, bool is_decoupled,
+    const inference::ModelMetrics& model_metrics)
 {
   // Global config only for now in config map
   auto metrics_config_map = Metrics::ConfigMap();
@@ -74,6 +75,20 @@ MetricReporterConfig::ParseConfig(
   // Set flag to signal to stats aggregator if caching is enabled or not
   cache_enabled_ = response_cache_enabled;
   is_decoupled_ = is_decoupled;
+
+  // Override default histogram options if set in model_metrics.
+  for (const auto& metric_control : model_metrics.metric_control()) {
+    const std::string& family_name =
+        metric_control.metric_identifier().family();
+    // Copy protobuf RepeatedField to std::vector
+    const auto& buckets_proto = metric_control.histogram_options().buckets();
+    const prometheus::Histogram::BucketBoundaries buckets(
+        buckets_proto.begin(), buckets_proto.end());
+
+    if (metric_map_.find(family_name) != metric_map_.end()) {
+      histogram_options_[metric_map_.at(family_name)] = buckets;
+    }
+  }
 }
 
 prometheus::Summary::Quantiles
@@ -120,6 +135,7 @@ MetricModelReporter::Create(
     const ModelIdentifier& model_id, const int64_t model_version,
     const int device, bool response_cache_enabled, bool is_decoupled,
     const triton::common::MetricTagsMap& model_tags,
+    const inference::ModelMetrics& model_metrics,
     std::shared_ptr<MetricModelReporter>* metric_model_reporter)
 {
   static std::mutex mtx;
@@ -148,7 +164,7 @@ MetricModelReporter::Create(
 
   metric_model_reporter->reset(new MetricModelReporter(
       model_id, model_version, device, response_cache_enabled, is_decoupled,
-      model_tags));
+      model_tags, model_metrics));
   reporter_map.insert({hash_labels, *metric_model_reporter});
   return Status::Success;
 }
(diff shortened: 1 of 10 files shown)


src/metric_model_reporter.cc
```diff
@@ -74,6 +75,20 @@ MetricReporterConfig::ParseConfig(
...
   is_decoupled_ = is_decoupled;
+
+  // Override default histogram options if set in model_metrics.
+  for (const auto& metric_control : model_metrics.metric_control()) {
+    const std::string& family_name =
+        metric_control.metric_identifier().family();
+    // Copy protobuf RepeatedField to std::vector
+    const auto& buckets_proto = metric_control.histogram_options().buckets();
+    const prometheus::Histogram::BucketBoundaries buckets(
+        buckets_proto.begin(), buckets_proto.end());
```
GuanLuo, line 86: move into if clause

src/metric_model_reporter.h
```diff
@@ -73,6 +77,12 @@ struct MetricReporterConfig {
   bool cache_enabled_ = false;
 
   bool is_decoupled_ = false;
+
+ private:
+  // Maps the metric fullname to its lookup key.
+  // All new histograms must update the map.
```
GuanLuo, line 83: Add comment on why you need to have a mapping

src/model_config_utils.cc
```diff
@@ -1620,6 +1620,32 @@ ValidateModelConfig(
...
+              Status::Code::INVALID_ARG,
+              "metric_identifier must specify 'family'");
+        }
+      } else {
+        return Status(
+            Status::Code::INVALID_ARG,
+            "model_control must specify 'metric_identifier'");
+      }
+
+      if (!metric_control.has_histogram_options()) {
```
GuanLuo, line 1639: Shouldn't you check if the histogram is empty?

src/model_config_utils.cc
```diff
@@ -1620,6 +1620,32 @@ ValidateModelConfig(
...
+        if (metric_control.metric_identifier().family().empty()) {
+          return Status(
+              Status::Code::INVALID_ARG,
+              "metric_identifier must specify 'family'");
+        }
+      } else {
+        return Status(
+            Status::Code::INVALID_ARG,
+            "model_control must specify 'metric_identifier'");
+      }
```
GuanLuo, line 1637: avoid nested if clause
```
if (!has()) {
    return Error;
}
if (...empty()) {
    return Error;
}
```

src/model_config_utils.cc
yinggeh: If you are referring to buckets, I don't see a way to check if buckets field is empty or not. The following two protobuf objects are identical examined in GDB. Since buckets `[]` is legal, I think we just have to allow buckets to be omitted.
```
histogram_options: {
    buckets: []
}

histogram_options: {
}
```

src/metric_model_reporter.cc
yinggeh: Updated

src/metric_model_reporter.h
yinggeh: Updated

src/model_config_utils.cc
yinggeh: Updated

src/model_config_utils.cc
yinggeh: Discussioned offline. Will disallow users to set empty buckets.

src/metric_model_reporter.cc
```diff
@@ -74,6 +75,25 @@ MetricReporterConfig::ParseConfig(
...
+    if (metric_map_.find(family_name) != metric_map_.end()) {
+      // Copy protobuf RepeatedField to std::vector
+      const auto& buckets_proto = metric_control.histogram_options().buckets();
+      const prometheus::Histogram::BucketBoundaries buckets(
+          buckets_proto.begin(), buckets_proto.end());
+      histogram_options_[metric_map_.at(family_name)] = buckets;
+    } else {
+      // metric_control config may be extended to support backend metrics.
+      LOG_WARNING << "Metric family '" << family_name
+                  << "' in 'model_identifier' is not a Triton core metric.";
```
rmccorm4, line 94: Is `'model_identifier' a placeholder? Should it be the real model name?

src/metric_model_reporter.cc
```diff
@@ -74,6 +75,25 @@ MetricReporterConfig::ParseConfig(
...
+    const std::string& family_name =
+        metric_control.metric_identifier().family();
+
+    // If family name exists, override with new options.
+    if (metric_map_.find(family_name) != metric_map_.end()) {
+      // Copy protobuf RepeatedField to std::vector
+      const auto& buckets_proto = metric_control.histogram_options().buckets();
+      const prometheus::Histogram::BucketBoundaries buckets(
+          buckets_proto.begin(), buckets_proto.end());
+      histogram_options_[metric_map_.at(family_name)] = buckets;
```
rmccorm4, line 90: What happens if a model is dynamically reloaded/updated (--model-control-mode=poll/explicit) with different buckets after server intial load?

ex:
```
# assume modelA with some config.pbtxt that specifies some bucket values
tritonserver --model-store models --model-control-mode poll

# Change bucket values
vim models/modelA/config.pbtxt

# Triton reloads/updates model after detecting change (or manually updated in 'explicit' mode)
load_model(modelA)
```

What's the behavior? Will the buckets get updated?

I think it's OK if we don't support dynamic update in this PR - but want to know the limitations up front and make sure nothing crashes.

src/metric_model_reporter.cc
yinggeh: Typo. Should be 'metric_identifier' in `model_config.proto`.

src/metric_model_reporter.cc
yinggeh: Explicit mode: explicit reloading the model with new buckets will update the histogram metric.
Poll mode: dynamically updating buckets does not update the histogram metric.

I suppose two modes use different mechanisms to reload the model?

src/metric_model_reporter.cc
GuanLuo: Fixed?

src/metric_model_reporter.h
```diff
@@ -79,7 +79,9 @@ struct MetricReporterConfig {
   bool is_decoupled_ = false;
 
  private:
-  // Maps the metric fullname to its lookup key.
+  // Maps the metric family fullname to its lookup key. This field is required
+  // to find the lookup key given its fullname when users specify in the custom
+  // metric configuration "ModelMetrics".
```
GuanLuo, line 84:
```suggestion
  // Maps the metric family fullname to its lookup key. This field is required
  // because the users are expected to configure metric configuration "ModelMetrics"
  // with the full name displayed from metrics reporting while a different name is used internally .
```

src/metric_model_reporter.cc
yinggeh: Updated

src/model_config_utils.cc
```diff
@@ -1620,6 +1620,38 @@ ValidateModelConfig(
             " cache.");
   }
 
+  // If model_metric is specified, validate it.
+  if (config.has_model_metrics()) {
+#ifdef TRITON_ENABLE_METRICS
+    for (const auto& metric_control : config.model_metrics().metric_control()) {
```
rmccorm4, line 1626: Assuming this section will continue to grow, would be good to break out to a helper function ValidateModelMetricsConfig - but this is fine inline for now.

rmccorm4 (approved):

src/model_config_utils.cc
yinggeh: Refactored to a helper function.

GuanLuo (dismissed):

src/model_config_utils.cc
```diff
@@ -1620,6 +1653,17 @@ ValidateModelConfig(
             " cache.");
   }
 
+  // If model_metric is specified, validate it.
+  if (config.has_model_metrics()) {
+#ifdef TRITON_ENABLE_METRICS
+    Status status = ValidateModelMetrics(config.model_metrics());
```
rmccorm4, line 1659: There should be some macro for this like RETURN_IF_ERR

yinggeh: Fixed

rmccorm4 (approved):

PR #406, opened 2024-10-31
Support Windows Long Paths

fpetrini15:
Contribution from internal team to enable Triton to support Windows long paths.

### DIFF
--- a/src/filesystem/implementations/local.h
+++ b/src/filesystem/implementations/local.h
@@ -84,9 +84,9 @@ LocalFileSystem::GetOSValidPath(const std::string& path)
   // compatibility, all paths are limited to MAX_PATH length
   if (l_path.size() >= MAX_PATH) {
     // Must be prefixed with "\\?\" to be considered long path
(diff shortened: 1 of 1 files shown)


src/filesystem/implementations/local.h
```diff
@@ -60,12 +60,44 @@ class LocalFileSystem : public FileSystem {
   Status MakeTemporaryDirectory(
       std::string dir_path, std::string* temp_dir) override;
   Status DeletePath(const std::string& path) override;
+
+ private:
+  inline std::string GetOSValidPath(const std::string& path);
+  static constexpr const char* kWindowsLongPathPrefix = "\\\\?\\";
```
GuanLuo, line 66: Why not move this into GetOSValidPath as static local variable

GuanLuo (approved):

src/filesystem/implementations/local.h
fpetrini15: I can do that.

GuanLuo (approved):

PR #414, opened 2024-11-22
build: Improves python packaging infrastructure

pranavm-nvidia:
Updates the Python package to use `build` to package instead of the legacy `bdist_wheel` style. `pyproject.toml` is the single source of truth for the package.

Additionally, this change also move stub generation to `setup.py` so that it happens automatically when the wheels are generated. This way the stub files don't need to be maintained in the repo and manually updated.


python/setup.py
```diff
@@ -25,89 +25,20 @@
...
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-import os
-import sys
-from itertools import chain
 
-from setuptools import find_packages, setup
+import subprocess
 
-if "--plat-name" in sys.argv:
```
pranavm-nvidia, line 34: Still need to figure out how to get the platform flags working

python/build_wheel.py
```diff
@@ -108,17 +104,18 @@ def sed(pattern, replace, source, dest=None):
 
     shutil.copyfile("LICENSE.txt", os.path.join(FLAGS.whl_dir, "LICENSE.txt"))
     shutil.copyfile("setup.py", os.path.join(FLAGS.whl_dir, "setup.py"))
+    shutil.copyfile("pyproject.toml", os.path.join(FLAGS.whl_dir, "pyproject.toml"))
 
     os.chdir(FLAGS.whl_dir)
     print("=== Building wheel")
-    args = ["python3", "setup.py", "bdist_wheel"]
+    args = ["python3", "-m", "build"]
```
pranavm-nvidia, line 111: Not sure how to make sure the `build` package is installed in the containers.

rmccorm4:
Need to run CI

python/build_wheel.py
rmccorm4, line 1: Can we remove TRITON_PYBIND from the github action in this PR now after rebasing on main?

pranavm-nvidia: I think so, I will need to test it out

python/setup.py
```diff
@@ -25,90 +25,23 @@
...
-    raise Exception("envvar VERSION must be specified")
 
-VERSION = os.environ["VERSION"]
+class BuildPyCommand(build_py):
+    def run(self):
+        build_py.run(self)
+        # Generate stub files:
+        package_name = self.distribution.metadata.name
+        subprocess.run(
+            ["stubgen", "-p", f"{package_name}._c", "-o", f"{self.build_lib}"],
```
rmccorm4, line 41: How was the stub generated before? Not seeing it in the diff

pranavm-nvidia: @rmccorm4 the same way but it was done as a manual offline step and the stub files were committed - that's why I'm able to remove those files now since they'll be generated automatically during the wheel build.

python/build_wheel.py
pranavm-nvidia: Removed this and looks like CI is still passing.

python/build_wheel.py
rmccorm4: TODO: I need to double check this

rmccorm4: Started pipeline: 21638660

pyproject.toml
```diff
@@ -24,6 +24,53 @@
...
+    "Intended Audience :: Science/Research",
+    "Intended Audience :: Information Technology",
+    "Topic :: Scientific/Engineering",
+    "Topic :: Scientific/Engineering :: Image Recognition",
+    "Topic :: Scientific/Engineering :: Artificial Intelligence",
+    "Topic :: Software Development :: Libraries",
+    "Topic :: Utilities",
+    "License :: OSI Approved :: BSD License",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.10",
```
rmccorm4, line 46:
```suggestion
    "Programming Language :: Python :: 3.10",
    "Programming Language :: Python :: 3.12",
```

rmccorm4: Add 3.12 per latest 24.11+ containers based on ubuntu 24.04

python/build_wheel.py
rmccorm4: Yep will need to install this in build container too

> /usr/bin/python3: No module named build
> error: Building wheel failed failed

I believe that would go here for RHEL and Ubuntu builds:
- https://github.com/triton-inference-server/server/blob/0194c3da9ddeeff07547f46aa058cf88acb51893/build.py#L987-L992
- https://github.com/triton-inference-server/server/blob/0194c3da9ddeeff07547f46aa058cf88acb51893/build.py#L1107-L1109

rmccorm4: I will try this out.

rmccorm4 (approved):
LGTM but needs https://github.com/triton-inference-server/server/pull/7898 to pass CI

Pipeline: 21672308

PR #415, opened 2024-11-22
Update `main` branch post 24.11

mc-nv:
This PR is subset of changes which need to be carried over to the `main` branch after the release

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -257,7 +257,7 @@ else()
   target_compile_options(
     triton-core
     PRIVATE
-      -Wall -Wextra -Wno-unused-parameter -Wno-deprecated-declarations -Werror
+      -Wall -Wextra -Wno-unused-parameter -Wno-deprecated-declarations -Wno-error=maybe-uninitialized -Werror
   )
   set_target_properties(
     triton-core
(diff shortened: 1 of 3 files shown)


lkomali (approved):

krishung5 (approved):

yinggeh (approved):

src/CMakeLists.txt
```diff
@@ -257,7 +257,7 @@ else()
   target_compile_options(
     triton-core
     PRIVATE
-      -Wall -Wextra -Wno-unused-parameter -Wno-deprecated-declarations -Werror
+      -Wall -Wextra -Wno-unused-parameter -Wno-deprecated-declarations -Wno-error=maybe-uninitialized -Werror
```
whoisj, line 260: would this be done better as a `#pragma` in the files which need this, this leaving the warning->error for everything else?

uninitialized memory can be difficult to debug when it causes unexpected behavior.

whoisj: example:
```C
    #pragma GCC diagnostic push
    #pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
    /* code which uses a maybe uninitialized variable here. */
    #pragma GCC diagnostic pop
```

I would be OK with entire files, types, function, etc. being marked as such. Still better than the entire source code base IMO.

mc-nv: It was easiest way to pass the compilation issues. 
This change is no more then cherry-pick after the release. 

In my opinion we should write code in the way that the warning as such won't come up.

PR #416, opened 2024-11-22
ci: Add GitHub action for core build and python testing in pull requests

pranavm-nvidia:


python/test/test_api.py
```diff
@@ -359,8 +359,9 @@ def test_stop(self):
                         "parameters": {"decoupled": {"string_value": "False"}},
                         # Keep instance count low for fast startup/cleanup.
                         # Alternatively can use KIND_CPU here, but keeping gpus/count explicit.
+                        # TODO: Check above
                         "instance_group": [
-                            {"kind": "KIND_GPU", "gpus": [0], "count": 1}
+                            {"kind": "KIND_CPU"}
```
pranavm-nvidia, line 364: This allows us to run this test on machines without GPUs, but the comment suggests it was intentionally run on GPU. If that's case, we will need to skip it in CI.

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,40 @@
...
+        apt install -y --no-install-recommends clang-format-15 cmake libb64-dev rapidjson-dev libre2-dev
+        wget -O /tmp/boost.tar.gz https://archives.boost.io/release/1.80.0/source/boost_1_80_0.tar.gz && (cd /tmp && tar xzf boost.tar.gz) && mv /tmp/boost_1_80_0/boost /usr/include/boost
+        pip install build pytest
+
+    - name: Build
+      run: |
+        mkdir -p /core/build
+        cd /core/build
+        cmake -DCMAKE_INSTALL_PREFIX:PATH=`pwd`/install -DTRITON_CORE_HEADERS_ONLY=OFF ..
+        export TRITON_PYBIND="_c/triton_bindings.cpython-310-x86_64-linux-gnu.so"
```
pranavm-nvidia, line 33: Need to remove this once my packaging tests are in

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,40 @@
+name: Build And Test
+
+on:
+    pull_request:
+      branches:
+        - main
+      types: [synchronize, opened, reopened, ready_for_review]
```
rmccorm4, line 7: NOTE: Essentially to skip running action when PR is a `draft`

python/test/test_api.py
rmccorm4: This test isn't doing anything GPU specific 👍

rmccorm4 (approved):

PR #422, opened 2024-12-31
UTF-16 Path Support

ashishk98:
Triton core server does not currently support paths that contain special unicode characters. This PR aims to fix this by decoding the paths just before use to a "std::wstring" and pass it to the relevant UTF16 supporting variants of functions being used in the core.

### DIFF
--- a/src/filesystem/api.h
+++ b/src/filesystem/api.h
@@ -25,6 +25,8 @@
 // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #pragma once
 
+#include <codecvt>
+#include <locale>
 #include <string>
 
 #include "../status.h"
@@ -67,6 +69,20 @@ class LocalizedPath {
   // tmp directory for the lifetime of this object
   // FIXME: Remove when no longer required
   std::vector<std::shared_ptr<LocalizedPath>> other_localized_path;
+
+  static inline std::wstring string2wstring(const std::string& str)
+  {
+    std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
+    std::wstring wide = converter.from_bytes(str);
+    return wide;
+  }
+
+  static inline std::string wstring2string(const std::wstring& str)
+  {
+    std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
+    std::string narrow = converter.to_bytes(str);
+    return narrow;
+  }
 
  private:
   std::string original_path_;
--- a/src/filesystem/api.h
+++ b/src/filesystem/api.h
@@ -32,6 +32,12 @@
 #include "../status.h"
 #include "google/protobuf/message.h"
 
+#ifdef _WIN32
+// suppress the min and max definitions in Windef.h.
+#define NOMINMAX
+#include <Windows.h>
+#endif
+
 namespace triton { namespace core {
 
 enum class FileSystemType { LOCAL, GCS, S3, AS };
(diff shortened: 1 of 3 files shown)


ashishk98:
@nnshah1 @nv-kmcgill53 Can you please take a look at this PR?

src/filesystem/api.h
```diff
@@ -25,6 +25,8 @@
 // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #pragma once
 
+#include <codecvt>
```
fpetrini15, line 28: Since this and the below only apply to Windows, should we encapsulate them within a #ifdef _WIN32?

src/filesystem/api.h
```diff
@@ -68,6 +70,20 @@ class LocalizedPath {
   // FIXME: Remove when no longer required
   std::vector<std::shared_ptr<LocalizedPath>> other_localized_path;
 
+  static inline std::wstring string2wstring(const std::string& str)
```
fpetrini15, line 73:
```suggestion
  static inline std::wstring String2Wstring(const std::string& str)
```
According to the style sheet, I suppose these functions should follow this naming convention, unless there's an exception I'm unaware of: https://google.github.io/styleguide/cppguide.html#Function_Names

src/filesystem/implementations/local.h
```diff
@@ -106,11 +112,18 @@ LocalFileSystem::IsDirectory(const std::string& path, bool* is_dir)
...
 
+#ifdef _WIN32
+  std::wstring wpath = LocalizedPath::string2wstring(GetOSValidPath(path));
+  struct _stat64i32 st;
+  if (_wstat64i32(wpath.c_str(), &st) != 0) {
+    return Status(Status::Code::INTERNAL, "failed to stat file " + path);
+  }
+#else
   struct stat st;
-  if (stat(GetOSValidPath(path).c_str(), &st) != 0) {
```
fpetrini15, line 110: @ashishk98, why do we no longer need your `GetOSValidPath` helper function here for Windows long paths?

fpetrini15:
These changes seem to move our Linux handling away from using `GetOSValidPath` which is built to handle both the Windows and Linux use-cases. Personally, I'd prefer it to be consistent, either we remove the Linux's build's use of `GetOSValidPath` and change the name to something more specific like `ConvertToWindowsLongPath` or have the Linux flow continue to use `GetOSValidPath` for parity's sake. However, I do understand that the latter has needless function calling overhead.

Thoughts from @nv-kmcgill53 or @mc-nv?

fpetrini15:
Thanks for the contribution, Ashish! When all threads have been resolved, I can migrate your changes to a dev branch and run in through CI!

src/filesystem/api.h
mc-nv: Unless same issue present for Linux. 

But in fact `codecvt` is part of `std` namespace same as `locale` and shouldn't impact

ashishk98: ^^ Exactly why I did not encapsulate these. The headers and functions used are a part of std lib and could be used on linux as well. Let me know if you want me to encapsulate these in _WIN32 ifdef

ashishk98:
> These changes seem to move our Linux handling away from using `GetOSValidPath` which is built to handle both the Windows and Linux use-cases. Personally, I'd prefer it to be consistent, either we remove the Linux's build's use of `GetOSValidPath` and change the name to something more specific like `ConvertToWindowsLongPath` or have the Linux flow continue to use `GetOSValidPath` for parity's sake. However, I do understand that the latter has needless function calling overhead.

Fixed this in the latest commit by changing GetOSValidPath() -> GetWindowsValidPath()

nv-kmcgill53:
I will rely on other's review of this code. I don't see any major things on my end but I don't feel as though I can put my approval stamp on this PR with the amount I looked at this.

src/filesystem/api.h
fpetrini15: Right, but in the present design, the functions and headers are exclusively used for Windows. It's stylistic and not a blocker.

fpetrini15:
LGTM! @mc-nv can you notify me when Windows CI is passing? We'll have to have this in for 24.01

mc-nv:
Closed based on merged changes

PR #426, opened 2025-01-11
perf: Python binding inference performance improvement

kthui:
#### What does the PR do?
Refactor `infer()` and `async_infer()` APIs to handle memory allocation and callbacks internally in C++, and only expose the basic interface for the Python iterator to fetch responses.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [x] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
https://github.com/triton-inference-server/server/pull/7949

Start with the tritonserver_pybind.cc for the interface change, then move on to _model.py on how the Python iterator interacts with the interface. Finally, move on to the _request.py and _response.py on how they support the Python iterator.

For testing, start with test_binding.py and test_api.py, and then _tensor.py on the DLPack limitation regarding bytes.

#### Test plan:
Existing L0_python_api is sufficient for catching any regression from this performance improvement. It is modified to test from the new interface.

- CI Pipeline ID: 23514700

#### Caveats:
User are no longer able to specify custom:
* request release callback
* response allocator
* response callback

Currently, only CPU memory output is supported at the binding level, so GPU memory output will involve an extra D2H copy at the backend and a H2D copy at the frontend. This will be resolved as a follow-up.

The [test_stop ](https://github.com/triton-inference-server/core/pull/426/files#diff-f4adade0afd240d379f3f6400b2a57103329f0866b8e2c7ac2fd626997725ef0R423-R426) failure will have to be triaged and fixed as a follow-up.

#### Background
N/A

N/A


python/test/test_api.py
```diff
@@ -137,6 +137,7 @@ def test_memory_fallback_to_cpu(self, server_options):
 
         tritonserver.default_memory_allocators[tritonserver.MemoryType.GPU] = allocator
```
nnshah1, line 138: should this be removed - or changed in some way to indicate the allocator is internal ....

python/test/test_api.py
```diff
@@ -164,6 +165,7 @@ def test_memory_allocator_exception(self, server_options):
             ):
                 pass
 
+    @pytest.mark.skip(reason="Skipping test, infer no longer use allocator")
```
nnshah1, line 168: we should keep this but refactor - if user requests output memory type gpu and that is not supported by the internal allocator - we would still want to raise an exception during inference

python/test/test_api.py
```diff
@@ -418,6 +420,9 @@ def test_ready(self, server_options):
         server = tritonserver.Server(server_options).start()
         assert server.ready()
 
+    @pytest.mark.skip(
+        reason="Skipping test, some request/response object may not be released which may cause server stop to fail"
```
nnshah1, line 424: Can we use xfail instead of skip?

@rmccorm4 - what do you think - as we had spent time on fixing this - how much an issue is this?

python/test/test_binding.py
```diff
@@ -666,18 +594,17 @@ def test_trace(self):
...
-            request_counter,
-        ) = self._prepare_inference_request(server)
+        request, input_ = self._prepare_inference_request(server)
         server.infer_async(request, trace)
 
         # [FIXME] WAR due to trace lifecycle is tied to response in Triton core,
         # trace reference should drop on response send..
-        res = response_queue.get(block=True, timeout=10)
+        future = concurrent.futures.Future()
+        request.get_next_response(future)
```
nnshah1, line 603: question - why does get_next_response not return a future, instead of taking one in?

python/test/test_binding.py
```diff
@@ -1030,12 +945,13 @@ def test_infer_async(self):
                 byte_size,
                 memory_type,
                 memory_type_id,
-                numpy_buffer,
             ) = out
+            ctypes_buffer = ctypes.create_string_buffer(byte_size)
+            ctypes.memmove(ctypes_buffer, out_buffer, byte_size)
+            numpy_buffer = numpy.frombuffer(ctypes_buffer, dtype=numpy.byte)
```
nnshah1, line 951: who owns the memory in this case - 

does it become owned by numpy or ctypes_buffer ... just for my understanding

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -31,6 +31,7 @@
 #include <triton/core/tritonserver.h>
 
 #include <iostream>
+#include <mutex>
```
nnshah1, line 34: oh no, one of my least favorite things to think about :)

python/test/test_binding.py
kthui: This workflow is mainly centered around [asyncio future](https://docs.python.org/3.12/library/asyncio-future.html#asyncio.Future.get_loop), which requires the presence of a running loop when it is created, for example
```python
loop = asyncio.get_running_loop()
future = loop.create_future()
```
where calling `asyncio.get_running_loop()` outside the async method that it will be awaited may ends up not finding the loop or finding the wrong loop, so I think it is more robust to have the `AsyncIterator` creates the future and then pass it into the binding.

The `Iterator` future simply follows the Async routine for a more alike interface, but in the future I think we can improve it  by having the async `get_next_response()` to return an awaitable, and the normal one to simply block until there is a response.

python/test/test_binding.py
kthui: The `out_buffer` is owned by the `res` object. The `ctypes_buffer` is a copy of the `out_buffer`, which the `ctypes_buffer` owns itself.

The `numpy_buffer` is not entirely clear from [the docs](https://numpy.org/doc/stable/reference/generated/numpy.frombuffer.html#numpy-frombuffer), but from my experiment it is referencing the `ctypes_buffer` as the `numpy_buffer` is marked as read-only and cannot interface with DLPack without a copy.

python/tritonserver/_c/tritonserver_pybind.cc
kthui: 😅

python/test/test_api.py
kthui: yes, updated the test to indicate the allocator is internal, and it will always use CPU memory regardless of the backend memory preference.

python/test/test_api.py
kthui: yes, we should raise an exception if the output memory type specified on the request is not supported, but currently the bindings does not accept a requested output memory type, so I think we can skip this test for now and add proper testing after adding support for allocating GPU memory.

python/test/test_api.py
kthui: yes, switched to xfail.

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -896,106 +871,128 @@ class PyInferenceRequest
...
+        allocator_.get(), ResponseAllocatorQueryFn));
+    ThrowIfError(TRITONSERVER_ResponseAllocatorSetBufferAttributesFunction(
+        allocator_.get(), ResponseAllocatorBufferAttributesFn));
+  }
+  static TRITONSERVER_Error* ResponseAllocatorQueryFn(
+      struct TRITONSERVER_ResponseAllocator* allocator, void* userp,
+      const char* tensor_name, size_t* byte_size,
+      TRITONSERVER_MemoryType* memory_type, int64_t* memory_type_id)
+  {
+    *memory_type = TRITONSERVER_MEMORY_CPU;
```
nnshah1, line 921: can we add a todo to support GPU?

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -896,106 +871,128 @@ class PyInferenceRequest
...
+    return nullptr;
+  }
+  static TRITONSERVER_Error* ResponseAllocatorAllocFn(
+      struct TRITONSERVER_ResponseAllocator* allocator, const char* tensor_name,
+      size_t byte_size, TRITONSERVER_MemoryType memory_type,
+      int64_t memory_type_id, void* userp, void** buffer, void** buffer_userp,
+      TRITONSERVER_MemoryType* actual_memory_type,
+      int64_t* actual_memory_type_id)
+  {
+    *buffer = malloc(byte_size * sizeof(uint8_t));
```
nnshah1, line 943: are there tritonserver apis we should be using for memory allocation instead of directly calling malloc?

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -896,106 +871,128 @@ class PyInferenceRequest
...
+    *actual_memory_type = TRITONSERVER_MEMORY_CPU;
+    *actual_memory_type_id = 0;
+    // The response allocator needs to be kept alive until the allocated memory
+    // is released via the release function.
```
nnshah1, line 947: is this true? - question: why not an allocator singleton for cpu - and reuse it for all requests?

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -896,106 +871,128 @@ class PyInferenceRequest
...
+    *buffer_userp = reinterpret_cast<void*>(allocator_ptr.release());
+    return nullptr;
+  }
+  static TRITONSERVER_Error* ResponseAllocatorReleaseFn(
+      struct TRITONSERVER_ResponseAllocator* allocator, void* buffer,
+      void* buffer_userp, size_t byte_size, TRITONSERVER_MemoryType memory_type,
+      int64_t memory_type_id)
+  {
+    free(buffer);
+    // Release ownership of the response allocator.
```
nnshah1, line 963: thinking a singleton could be simpler and avoid alloc / dealloc of the allocator object itself - if possible.

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -896,106 +871,128 @@ class PyInferenceRequest
...
+    std::unique_ptr<std::shared_ptr<struct TRITONSERVER_ResponseAllocator>>
+        allocator_ptr(reinterpret_cast<
+                      std::shared_ptr<struct TRITONSERVER_ResponseAllocator>*>(
```
nnshah1, line 966: actually - what do we use the buffer_userp for?

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -1175,20 +1171,77 @@ class PyInferenceRequest
...
+  {
+    std::lock_guard lock(response_mu_);
 
- public:
-  std::unique_ptr<CallbackResource> request_callback_resource_{nullptr};
+    if (responses_.empty()) {
+      if (response_future_.get() != nullptr) {
+        throw AlreadyExistsError("cannot call GetNextResponse concurrently");
+      }
+      response_future_.reset(new py::object(py_future));
```
nnshah1, line 1189: not sure I follow the logic here - why do we create a new object based on the passed in object? why not a reference on the original object?

second: how does the swap below work?

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -1175,20 +1171,77 @@ class PyInferenceRequest
...
+      std::lock_guard lock(response_mu_);
+      if (response_future_.get() == nullptr) {
+        responses_.push_back(std::move(py_response));
+        return;
+      }
+    }
+    {
+      py::gil_scoped_acquire gil;
+      std::unique_ptr<py::object> response_future_local(nullptr);
+      response_future_.swap(response_future_local);
```
nnshah1, line 1218: not sure I follow this  - we create a new py object from null, swap it with the stored one (releasing it?) - then set result on the nullptr based object?...

Or ... I guess swap works the opposite way :) 

we swap the null ptr for the one stored in the future ....

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -1175,20 +1171,77 @@ class PyInferenceRequest
...
+      const py::object& py_future,
+      std::pair<std::shared_ptr<PyInferenceResponse>, const uint32_t>&
+          py_response)
+  {
+    py::tuple py_res = py::make_tuple(
+        (py_response.first.get() ? py::cast(py_response.first) : py::none()),
+        py_response.second);
+    if (py::hasattr(py_future, "get_loop")) {
+      py_future.attr("get_loop")().attr("call_soon_threadsafe")(
+          py_future.attr("set_result"), std::move(py_res));
```
nnshah1, line 1233: neat! So this support async and non async?

python/tritonserver/_api/_model.py
```diff
@@ -180,40 +181,52 @@ def async_infer(
...
+        # alive.
         request = inference_request._create_tritonserver_inference_request()
 
-        response_iterator = AsyncResponseIterator(
-            self,
-            request,
-            inference_request.response_queue,
-            raise_on_error,
-        )
+        self._server.infer_async(request)
```
nnshah1, line 189: how do we pass in the requested memory type to the C++ side?

python/tritonserver/_api/_model.py
```diff
@@ -180,40 +181,52 @@ def async_infer(
...
+
+        return AsyncResponseIterator(self, request, inference_request)
 
     def infer(
         self,
         inference_request: Optional[InferenceRequest] = None,
         raise_on_error: bool = True,
         **kwargs: Unpack[InferenceRequest],
-    ) -> ResponseIterator:
+    ) -> Iterable:
```
nnshah1, line 229: why not use the explicit types?

Is it because of forward declaration?

If so we can use quotes, or import from typing future to allow for forward declaration,

```
"ResponseIterator"
```

python/tritonserver/_api/_model.py
```diff
@@ -180,40 +181,52 @@ def async_infer(
...
-            inference_request.response_queue,
-            raise_on_error,
-        )
+        self._server.infer_async(request)
 
-        response_allocator = ResponseAllocator(
-            inference_request.output_memory_allocator,
-            inference_request.output_memory_type,
-        ).create_tritonserver_response_allocator()
+        class AsyncResponseIterator:
```
nnshah1, line 191: I see this is a nested class - is that needed?

python/tritonserver/_api/_response.py
```diff
@@ -34,288 +34,27 @@
 from dataclasses import asdict, dataclass, field
 from typing import TYPE_CHECKING, Optional
 
+import numpy
 from tritonserver._api import _model
 
 if TYPE_CHECKING:
     from tritonserver._api._model import Model
 
+from tritonserver._api._allocators import MemoryBuffer
```
nnshah1, line 43: we could rename this as MemoryBuffer and /or put it into Tensor.py as its used there as well and no longer needed in allocators ....

python/tritonserver/_api/_response.py
```diff
@@ -395,14 +133,21 @@ def _from_tritonserver_inference_response(
...
+                    data_ptr=data_ptr,
+                    memory_type=memory_type,
+                    memory_type_id=memory_type_id,
+                    size=byte_size,
+                    owner=response,
+                )
                 tensor = Tensor(data_type, shape, memory_buffer)
-
+                if output_memory_type is not None:
+                    tensor = tensor.to_device(output_memory_type)
```
nnshah1, line 150: do we need to check for error here - or will propagate?

Should check original behavior in terms of fallback behavior -

python/tritonserver/_api/_tensor.py
```diff
@@ -509,16 +509,32 @@ def to_device(self, device: DeviceOrMemoryType) -> Tensor:
         if self.memory_type == MemoryType.CPU_PINNED and memory_type == MemoryType.CPU:
             return self
         if cupy is not None:
+            # DLPack does not support bytes type.
+            original_data_type = self.data_type
+            original_shape = self.shape
+            if self.data_type == DataType.BYTES:
```
nnshah1, line 515: question - would this remain after we enable device memory on c++ side?

or is temporary - if temporary - we can add a note on when it can be removed?

nnshah1 (commented):
Nice work! 

Will want secondary review as well. 

For some of the questions may be easier to sync offline.

python/test/test_api.py
nnshah1: hmm, but if we fail because say cupy is not available? Can we still make sure the right error gets propagated?

kthui: yes, the "test_unsupported_memory_type" is repurposed for testing moving outputs to unsupported memory type

python/tritonserver/_c/tritonserver_pybind.cc
kthui: yes, it is added

python/tritonserver/_c/tritonserver_pybind.cc
kthui: I don't think there is one, because there are multiple calls to malloc in the core code base for different purpose, i.e.
https://github.com/triton-inference-server/core/blob/r24.12/src/backend_memory_manager.cc#L98
https://github.com/triton-inference-server/core/blob/r24.12/src/backend_model_instance.cc#L73
https://github.com/triton-inference-server/core/blob/r24.12/src/cache_manager.cc#L95

There is one on the server repo, but it is our of reach from the core binding:
https://github.com/triton-inference-server/server/blob/r24.12/src/memory_alloc.cc#L96

python/tritonserver/_c/tritonserver_pybind.cc
kthui: yes, the core will need the allocator opaque object for finding the set release function. The allocator opaque object is basically [this class](https://github.com/triton-inference-server/core/blob/r24.12/src/response_allocator.h#L35) that stores the [release function pointer](https://github.com/triton-inference-server/core/blob/r24.12/src/response_allocator.h#L73).

python/tritonserver/_c/tritonserver_pybind.cc
kthui: yes, updated to use a singleton response allocator.

python/tritonserver/_c/tritonserver_pybind.cc
kthui: I tried holding a raw pointer to the original object, but the pointer can no longer be dereferenced (segmentation fault) after the `GetNextResponse()` returned, so the only safe way of holding on to the future object after `GetNextResponse()` returns is to increment the reference count on the future object, which is achieved by "copying" it to a new object.

python/tritonserver/_c/tritonserver_pybind.cc
kthui: the guarantee here that the `GetNextResponse()` will not be called and receive a new future object (vs this future object) is this future object is not yet set with a result, so as soon as this future object is set with a result, we need to be ready for the next future object from `GetNextResponse()` that may take the place of this future object. thus, the solution here is to move the current future object from the class to local variable, making the class variable ready for the next future object before the current future object is set with a result.

python/tritonserver/_c/tritonserver_pybind.cc
kthui: yes

python/tritonserver/_api/_model.py
kthui: I think we can add a new binding function to the request wrapper that allows setting a requested output memory type to the request wrapper object, then the response allocation function can access the requested memory type from the request wrapper object and allocate the correct memory type.

python/tritonserver/_api/_model.py
kthui: it is reverted to use the explicit type `ResponseIterator`.

python/tritonserver/_api/_model.py
kthui: it is not needed, the iterators are moved to the _response.py.

python/tritonserver/_api/_response.py
kthui: yes, the file is renamed to _memorybuffer.py, and the allocator apis are removed as custom allocators are no longer used with the infer api.

python/tritonserver/_api/_response.py
kthui: the error will be set in the response.

it is updated to raise `InvalidArgumentError` similar to the original behavior.

python/tritonserver/_api/_tensor.py
kthui: I think this can be permanent, because we (may?) want to support moving bytes tensors between host and device(s), which this adds the functionality.

python/test/test_api.py
rmccorm4, line 1: Not a blocker for this PR if you want to separate it, but I think it should be relatively easy - can we add a perf benchmark test case on the infer/async_infer APIs to start having some data or easier way to measure python in-process API performance for future reference?

Some examples:
- https://github.com/triton-inference-server/triton_distributed/blob/a94373d5c1c779c485175cab5dc4445597cbf5f7/worker/tests/python/integration/test_perf_benchmark.py#L225
- https://github.com/ionelmc/pytest-benchmark

rmccorm4: For example, if you picked a dummy model that calls `sleep(1)` and we can expect every inference to take ~1 second with some allowance threshold, we can probably automate identifying some bottlenecks or regressions from the results: https://pytest-benchmark.readthedocs.io/en/latest/usage.html

python/tritonserver/_c/tritonserver_pybind.cc
kthui: Experiment: Increment the reference count on the py_future object, and see if the object can be remained valid after the function return.

rmccorm4: Can we have some kind of small memory leak/growth pytest test that checks memory before, does some inferences, and checks memory after?

kthui: The implementation is updated to remove the pointer wrapper around `response_future_`, and more comments clarifying the flow among `GetNextResponse()` and `AddNextResponse()`: [Avoid extra malloc when tracking response_future_](https://github.com/triton-inference-server/core/pull/426/commits/c5ee14009b2c10434e3eefece50db40432da7314)
cc @nnshah1 @rmccorm4

kthui: > Can we have some kind of small memory leak/growth pytest test that checks memory before, does some inferences, and checks memory after?

Do you think something like [this ](https://pytest-memray.readthedocs.io/en/latest/usage.html) makes sense? I think there will always be some small growth, but we can set the acceptable growth on the test case and have the test fail if it grows beyond what we set.

python/test/test_api.py
kthui: How about we file a new ticket that does

> > Can we have some kind of small memory leak/growth pytest test that checks memory before, does some inferences, and checks memory after?

> Do you think something like [this ](https://pytest-memray.readthedocs.io/en/latest/usage.html)makes sense? I think there will always be some small growth, but we can set the acceptable growth on the test case and have the test fail if it grows beyond what we set.

and the mini perf benchmark together?

The mini perf test ensures there are many inference calls in the test which magnifies leaks if any, making the leak test more accurate. If the leak test is impacting performance, we can always run the benchmark twice, one without the leak test and another one with it.

python/tritonserver/_c/tritonserver_pybind.cc
rmccorm4: Looks worth a try to me - I see it's used by quite a few big projects - https://github.com/search?q=pytest-memray&type=code

- pydantic
- datadog
- huggingface
- ...

I'm curious how much time overhead it adds to run.

rmccorm4: > but we can set the acceptable growth on the test case and have the test fail if it grows beyond what we set.

Let's make sure whatever threshold we set isn't tailored to the specific test case or number of iterations, and doesn't get hit if we increased the iterations on the test case or something

python/test/test_api.py
```diff
@@ -135,36 +129,10 @@ def test_memory_fallback_to_cpu(self, server_options):
...
-                inputs={
-                    "string_input": tritonserver.Tensor.from_string_array([["hello"]])
-                },
-                output_memory_type="gpu",
-                output_memory_allocator=TestAllocators.MockMemoryAllocator(),
-            ):
-                pass
-
     def test_unsupported_memory_type(self, server_options):
+        # TODO: Revisit this test when GPU memory support is added, i.e. the request
```
rmccorm4, line 133: Do we have a ticket for GPU follow-up? Can add `DLIS-XXXX` here if so

rmccorm4 (commented):
Awesome work on this PR @kthui ! 

Overall changes and concepts to reduce thrashing on python GIL acquisition make sense to me - this was an awesome find.

Left a couple comments on benchmarking, memory checks, tests, etc

python/test/test_api.py
rmccorm4: If you have a manual verification that this change doesn't have or introduce a memory leak - then follow-up sounds good to me. For example, tracking memory usage of OpenAI frontend after a few runs of genai-perf or something.

rmccorm4: Just want to double check since this PR deals directly with memory allocators, which can be contingent on lifecycle management of requests/responses/etc - as well as the python future object lifetime management/ref_count changes

python/test/test_api.py
```diff
@@ -418,6 +310,10 @@ def test_ready(self, server_options):
         server = tritonserver.Server(server_options).start()
         assert server.ready()
 
+    @pytest.mark.xfail(
+        run=False,
+        reason="Some request/response object may not be released which may cause server stop to fail",
+    )
```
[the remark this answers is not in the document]
kthui: > What happens when server.stop() is called?

According to `test_api_logs/test_stop.server.log`, the server tried to shutdown, but it timeout at the end. There is no crash on the server log.

The pytest itself crashed afterward, in a segmentation fault.

> Will it always happen, or only intermittently?

It happens 100% of the time.

> Does letting the server go out of scope naturally without a call to .stop() work nicely?

There will be no crash on the pytest, but the server shutdown still timeout, according to the `test_stop.server.log`.

python/test/test_api.py
kthui: Yes, added the number to the codes: [[comment] Add ticket number to GPU follow-up TODOs](https://github.com/triton-inference-server/core/pull/426/commits/5a307b8d9b43e60773878255a544855df804d129)

python/test/test_api.py
kthui: Ok, will add the leak test.

python/tritonserver/_api/_model.py
```diff
@@ -128,9 +127,7 @@ def create_request(self, **kwargs: Unpack[InferenceRequest]) -> InferenceRequest
         _server=<tritonserver._c.triton_bindings.TRITONSERVER_Server
         object at 0x7f5827156bf0>, request_id=None, flags=0,
         correlation_id=None, priority=0, timeout=0, inputs={},
-        parameters={}, output_memory_type=None,
-        output_memory_allocator=None, response_queue=None,
-        _serialized_inputs={})
+        parameters={}, output_memory_type=None, _serialized_inputs={})
```
nnshah1, line 130: Todo: for the future - we should find a way to document how to add a custom allocator - memory allocation is an area where customization is probably needed

python/tritonserver/_api/_model.py
nnshah1: let's make sure the stories to follow up on customizing allocation are in place and scheduled. Priority would be to specify the memory type (device id, etc.)

python/tritonserver/_api/_model.py
```diff
@@ -180,33 +177,12 @@ def async_infer(
...
             inference_request = InferenceRequest(model=self, **kwargs)
 
-        if (inference_request.response_queue is not None) and (
-            not isinstance(inference_request.response_queue, asyncio.Queue)
-        ):
-            raise InvalidArgumentError(
-                "asyncio.Queue must be used for async response iterator"
-            )
-
+        # The 'request' is referencing pointers managed by the 'inference_request', so
```
nnshah1, line 180: this note seems unnecessary here - would we add it instead to the asyncresponseiterator object? That's where we maintain that relationship is that correct?

python/tritonserver/_api/_model.py
```diff
@@ -275,31 +251,12 @@ def infer(
...
             inference_request = InferenceRequest(model=self, **kwargs)
 
-        if (inference_request.response_queue is not None) and (
-            not isinstance(inference_request.response_queue, queue.SimpleQueue)
-        ):
-            raise InvalidArgumentError(
-                "queue.SimpleQueue must be used for response iterator"
-            )
-
+        # The 'request' is referencing pointers managed by the 'inference_request', so
```
nnshah1, line 254: see above - I think this more logically belongs with the comments for the Response Iterator - or could be omitted

python/tritonserver/_api/_response.py
```diff
@@ -29,293 +29,37 @@
...
-                self._user_queue.put(response)
-            if flags == TRITONSERVER_ResponseCompleteFlag.FINAL:
-                del self._request
-                self._request = None
-        except Exception as e:
-            message = f"Catastrophic failure in response callback: {e}"
-            LogMessage(LogLevel.ERROR, message)
-            # catastrophic failure
-            raise e from None
+DeviceOrMemoryType = (
```
nnshah1, line 60: nit: this may be better defined in _memorybuffer.py and reused

python/tritonserver/_api/_response.py
```diff
@@ -360,13 +104,12 @@ class InferenceResponse:
...
     def _from_tritonserver_inference_response(
         model: _model.Model,
-        request: TRITONSERVER_InferenceRequest,
         response,
         flags: TRITONSERVER_ResponseCompleteFlag,
+        output_memory_type: Optional[DeviceOrMemoryType] = None,
     ):
         result = InferenceResponse(
             model,
-            request.id,
```
nnshah1, line 369: question: why omit this / change this - the idea here is that in case there is a fundamental error (i.e. an empty response, or some other error) - we get at least the request.id as stored in the original request to tie this response to the error - as in line 118 we aren't guranteed a full response ...

python/tritonserver/_api/_response.py
```diff
@@ -360,13 +104,12 @@ class InferenceResponse:
     @staticmethod
     def _from_tritonserver_inference_response(
         model: _model.Model,
-        request: TRITONSERVER_InferenceRequest,
         response,
         flags: TRITONSERVER_ResponseCompleteFlag,
+        output_memory_type: Optional[DeviceOrMemoryType] = None,
```
nnshah1, line 109: maybe instead of passing output_memory_type alone - we could pass the entire original request object and also use it to set the response id - incase the response is none?

python/tritonserver/_api/_response.py
```diff
@@ -395,22 +138,103 @@ def _from_tritonserver_inference_response(
...
+                    f"Memory type {output_memory_type} not supported: {e}"
+                )
+
         # TODO: support classification
         # values["classification_label"] = response.output_classification_label()
 
         return result
+
+
+class AsyncResponseIterator:
```
nnshah1, line 179: If I understand correctly - does this remove the ability for someone to set a queue / async queue in addition to the iterators?

python/tritonserver/_api/_response.py
```diff
@@ -29,293 +29,37 @@
...
-
-        if self._complete:
-            raise StopAsyncIteration
-        response = await self._queue.get()
-        self._complete = response.final
-        if response.error is not None and self._raise_on_error:
-            raise response.error
-        return response
-
-    def cancel(self) -> None:
```
nnshah1, line 151: aren't we missing cancel in the new iterator?

python/tritonserver/_api/_response.py
```diff
@@ -29,293 +29,37 @@
...
-
-        if loop is None:
-            loop = asyncio.get_running_loop()
-        self._loop = loop
-        self._queue = asyncio.Queue()
-        self._user_queue = user_queue
-        self._complete = False
-        self._request = request
-        self._model = model
-        self._raise_on_error = raise_on_error
```
nnshah1, line 104: are we missing raise_on_error with new iterator?

python/tritonserver/_api/_response.py
```diff
@@ -29,293 +29,37 @@
...
-            Final responses may or may not include outputs and must be
-            checked.
-
-        """
-
-        if self._complete:
-            raise StopIteration
-        response = self._queue.get()
-        self._complete = response.final
-        if response.error is not None and self._raise_on_error:
```
nnshah1, line 276: this seems missing from the new code?

python/tritonserver/_api/_tensor.py
nnshah1: would like to see some examples - to understand a little - we also have the to_bytes_array / from_bytes_array - maybe those should also take a target device? currently assume to be numpy ... but not a blocker now

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -868,16 +844,15 @@ class PyInferenceResponse
     ThrowIfError(TRITONSERVER_InferenceResponseOutput(
         triton_object_, index, &name, &datatype, &shape, &dim_count, &base,
         &byte_size, &memory_type, &memory_type_id, &userp));
+    // The base pointer is deallocated when the response is deallocated.
```
nnshah1, line 847: comment needed? are all pointers deallocated when response is deallocated?

python/tritonserver/_c/tritonserver_pybind.cc
nnshah1: @GuanLuo to double check - should we be using the backend_memory_manager ? Reason I ask is that these / and the server already do support the different types of memory gpu / cpu / cpu pinned - and we would want that kind of support here if we can get it - and we should reuse it ...

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -896,107 +871,120 @@ class PyInferenceRequest
...
   {
-    // Note that Release() doesn't change ownership as the
-    // same PyInferenceRequest will be passed along the life cycle.
-    allocator_callback_resource_.release();
-    response_callback_resource_.release();
-    return triton_object_;
+    // The request wrapper needs to be kept alive until the release callback is
+    // invoked, so the Triton request can be deleted after the core/backend is
+    // done with it.
+    std::unique_ptr<std::shared_ptr<PyInferenceRequest>> request_ptr(
```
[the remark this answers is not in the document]
nnshah1: or should we be using ```make_unique``` instead of new? maybe not supported for our c++ version.

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -896,107 +871,120 @@ class PyInferenceRequest
...
+    ThrowIfError(TRITONSERVER_InferenceRequestSetReleaseCallback(
+        triton_object_, ReleaseCallback,
+        reinterpret_cast<void*>(
+            request_ptr.release()) /* request_release_userp */));
+  }
+  static void ReleaseCallback(
+      struct TRITONSERVER_InferenceRequest* request, const uint32_t flags,
+      void* userp)
+  {
+    std::unique_ptr<std::shared_ptr<PyInferenceRequest>> request_ptr(
```
nnshah1, line 900: I almost prefer to be more verbose here, 
```
std::shared_ptr<PyInferenceRequest> * request_ptr = reinterpret_cast<std::shared_ptr<PyInferenceRequest>*>(userp));
delete request_ptr;
```

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -896,107 +871,120 @@ class PyInferenceRequest
...
-    // goes through the request release cycle. This is due to
-    // a 'keep_alive' relationship is built between 'PyInferenceRequest'
-    // and 'PyServer': a request is associated with a server and the server
-    // should be kept alive until all associated requests is properly released.
-    // And here we exploit the 'keep_alive' utility in PyBind to guarantee so.
-    // See PyServer::InferAsync on how this field is set to avoid potential
-    // circular inclusion.
-    std::shared_ptr<PyInferenceRequest> request;
+      if (allocator != nullptr) {
+        ThrowIfError(TRITONSERVER_ResponseAllocatorDelete(allocator));
```
nnshah1, line 908: question - is it safe to call TRITONSERVER_ResponseAllocatorDelete(nullptr) ?

Generally for free / delete that is safe - just question if that is true here as well

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -896,107 +871,120 @@ class PyInferenceRequest
...
+    };
   };
-
-
-  void SetReleaseCallback(ReleaseFn release, const py::object& user_object)
+  void SetResponseAllocator()
+  {
+    // The response allocator is shared among all instances of this class, so it
+    // needs to be initialized once and once only.
+    // TODO: Why 'numpy.ones(2**27)' gets stuck when the response allocator is
```
nnshah1, line 916: seems very odd like something is interacting between numpy and this allocator - and that doesn't seem to make sense to me ....

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -896,107 +871,120 @@ class PyInferenceRequest
...
+  static TRITONSERVER_Error* ResponseAllocatorBufferAttributesFn(
+      struct TRITONSERVER_ResponseAllocator* allocator, const char* tensor_name,
+      struct TRITONSERVER_BufferAttributes* buffer_attributes, void* userp,
+      void* buffer_userp)
   {
-    py::gil_scoped_acquire gil;
-    auto cr = reinterpret_cast<CallbackResource*>(userp);
-    cr->release_fn(cr->request, flags, cr->user_object);
-    delete cr;
+    ThrowIfError(TRITONSERVER_BufferAttributesSetMemoryType(
```
nnshah1, line 946: do we want to throw if error or return the error?

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -896,107 +871,120 @@ class PyInferenceRequest
...
+    *actual_memory_type = TRITONSERVER_MEMORY_CPU;
+    *actual_memory_type_id = 0;
+    return nullptr;
+  }
+  static TRITONSERVER_Error* ResponseAllocatorReleaseFn(
+      struct TRITONSERVER_ResponseAllocator* allocator, void* buffer,
+      void* buffer_userp, size_t byte_size, TRITONSERVER_MemoryType memory_type,
+      int64_t memory_type_id)
+  {
+    free(buffer);
```
nnshah1, line 969: can we add assert here to ensure memory type and memory type id are cpu and 0?

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -1175,21 +1162,109 @@ class PyInferenceRequest
...
+  // Response management
+  void GetNextResponse(const py::object& py_future)
+  {
+    // Called exclusively from Python, so GIL is always held.
+    std::lock_guard lock(response_mu_);
 
- public:
-  std::unique_ptr<CallbackResource> request_callback_resource_{nullptr};
+    // There are two cases:
+    // 1. The backend is faster than the frontend:
```
nnshah1, line 1178:
```suggestion
    // 1. Responses have been generated and queued before GetNextResponse() has been called
    // 2. GetNextResponse() is called before responses have been generated and queued. 
```

I would remove the notion of "faster" and "slower" - I think that can be confusing here - it's not necessarily about speed of execution (i.e. applicaiton may be purposefully not reading, handling something else, etc.) - faster and slower makes it seem like a rate matching issue - but's really just whether or not there are items available when GetNextResponse is called ....

nnshah1: also change the description to match the if / else order, so:

1. GetNextResponse is called before / when no responses have been queued.

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -1175,21 +1162,109 @@ class PyInferenceRequest
...
+    //      There will be response(s) queued at 'responses_' deque, so pop the
+    //      response from the deque and set it to the 'py_future' object.
+    // 2. The backend is slower than the frontend:
+    //      The 'responses_' deque will be empty, so track the 'py_future' in
+    //      this request object until the 'AddNextResponse()' is called.
+    if (responses_.empty()) {
+      if (response_future_.ptr() != nullptr) {
+        throw AlreadyExistsError("cannot call GetNextResponse concurrently");
+      }
+      response_future_ = py_future;
```
nnshah1, line 1188: question - does this increment the py_future ref count?

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -1175,21 +1162,109 @@ class PyInferenceRequest
...
+  {
+    // Pair the 'response' and 'flags' so they can be easily moved around.
+    std::shared_ptr<PyInferenceResponse> managed_ptr(nullptr);
+    if (response != nullptr) {
+      managed_ptr.reset(new PyInferenceResponse(response, true /* owned */));
+    }
+    std::pair<std::shared_ptr<PyInferenceResponse>, const uint32_t> py_response(
+        std::move(managed_ptr), std::move(flags));
+
+    // There are two cases:
```
nnshah1, line 1209:
```suggestion
    // There are two cases:
    // 1. GetNextResponse() hasn't been called and the future is not set.
    // 2. GetNextResponse() has been called and future is set.
```

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -1175,21 +1162,109 @@ class PyInferenceRequest
...
+        return;
+      }
+    }
+    // There is no need to hold the 'response_mu_' after knowing the frontend is
+    // waiting, given there should be no concurrent calls into
+    // 'GetNextResponse', the frontend will wait until a result is set for the
+    // 'response_future_'.
+    {
+      py::gil_scoped_acquire gil;
+      py::object response_future_local(std::move(response_future_));
```
nnshah1, line 1231: still a little unsure why we need this is GetNextResponse holds the gil - but I think we can explore that seperately (the explicit move and release of the response_future_)

python/tritonserver/_c/tritonserver_pybind.cc
nnshah1: @rmccorm4, @GuanLuo for style - I see the general goodness of unique_ptr - but it kind of obscures what's happening here ....

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -896,107 +871,120 @@ class PyInferenceRequest
...
-      PyInferenceResponse::CompleteFn response,
-      const py::object& response_user_object)
+  void SetResponseCallback()
   {
-    allocator_callback_resource_.reset(
-        new PyResponseAllocator::CallbackResource(
-            allocator, allocater_user_object));
-    response_callback_resource_.reset(new PyInferenceResponse::CallbackResource(
-        response, allocator_callback_resource_.get(), response_user_object));
+    // The caller is responsible for keeping the request wrapper alive until
```
nnshah1, line 975: this is maintained by the _api ? is that correc t- so user doesn't have to worry about it explicitly -double checking

nnshah1 (commented):
generally looks good to me - had a few questions about removing the queues, cancel and request id from the response iterators / creation -

python/tritonserver/_c/tritonserver_pybind.cc
GuanLuo: What is in server certainly won't work, it is tightly related to the server frontend for optimization which is why allocator is up for the user to define.

You probably can reuse the one for backend but that is kind of strange. Plus the one in backend may not be as well crafted as you expected.. Please evaluate whether it may sense to proceed in this way.

nnshah1: Let's make that a todo as we bring in support for more memory types - we should look to see if we can reuse a common memory allocator in the core - of the ones backend_memeory_manager seems the closest - but agree is odd to reuse here directly.

python/tritonserver/_api/_model.py
kthui: yes, DLIS-7824.

python/tritonserver/_api/_tensor.py
kthui: I think the to_bytes_array underneath does the same hack and then call _to_numpy_on_host which underneath calls from_dlpack. Probably it is a good idea to unify the logic in the future, filed a ticket DLIS-7936 to keep track of this issue.

python/tritonserver/_api/_model.py
kthui: There is a story DLIS-7824 for supporting GPU memory allocation, but not a custom allocator that can be supplied by the user.

do you think that will be enough? or there will be use-case where there are memory types that we cannot implement in advance and it has to be supplied by the user at runtime?

python/tritonserver/_api/_model.py
kthui: sure, updated.

python/tritonserver/_api/_model.py
kthui: sure, updated.

python/tritonserver/_api/_response.py
kthui: sure, unified the definition across apis. [Unify definition of DeviceOrMemoryType](https://github.com/triton-inference-server/core/pull/426/commits/bc6e7acac7fc021d128de5951f33ee920362da2e)

python/tritonserver/_api/_response.py
kthui: there were some issue with the request lifecycle, but it is fixed later, so it is safe to add by the request.id setting. updated: [Update inference response conversion interface](https://github.com/triton-inference-server/core/pull/426/commits/023a1459d84d26b42a80b659de2a5882cb62a4f8)

python/tritonserver/_api/_response.py
kthui: yes, updated: [Update inference response conversion interface](https://github.com/triton-inference-server/core/pull/426/commits/023a1459d84d26b42a80b659de2a5882cb62a4f8)

python/tritonserver/_api/_response.py
kthui: yes

python/tritonserver/_api/_response.py
kthui: yes, added the ability to call cancel on the request. [Add ability to cancel a request from response iterator](https://github.com/triton-inference-server/core/pull/426/commits/dbbd68e3a10397ea0f164ac9db9255b3b4ef895b)

python/tritonserver/_api/_response.py
kthui: yes, added raise on error: [Add raise on error to response iterators](https://github.com/triton-inference-server/core/pull/426/commits/898ef02f33e74822a6b175068f4ae7cdf13c62d4)

python/tritonserver/_api/_response.py
kthui: yes, this is added [Add raise on error to response iterators](https://github.com/triton-inference-server/core/pull/426/commits/898ef02f33e74822a6b175068f4ae7cdf13c62d4)

python/tritonserver/_c/tritonserver_pybind.cc
kthui: yes, the pointers are deallocated when the response is deallocated. cleaned up the comment: [Clean up base pointer deallocation comment](https://github.com/triton-inference-server/core/pull/426/commits/9575a0054b4510f161710aa60355a5d29c20c509)

python/tritonserver/_c/tritonserver_pybind.cc
kthui: yes, removed the unique_ptr wrapper: [Remove unique_ptr wrapper on release callback](https://github.com/triton-inference-server/core/pull/426/commits/141483fa946cf5aa2610b6ea08facc3cc89bc705)

python/tritonserver/_c/tritonserver_pybind.cc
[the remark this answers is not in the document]
kthui: I'm not sure if I follow the idea behind
```c++
1  std::shared_ptr<PyInferenceRequest> request_ptr = request;  // increment reference count
2  void * user_p = reinterpret_cast<void*>(new std::shared_ptr<T>(request_ptr)); // convert to void * 
3  TRITONSERVER_InferenceRequestSetReleaseCallback(triton_object_, ReleaseCallback, user_p); 
```
correctly. It seems like #1 increment the ref count by creating a new request on the stack, and then it is passed to #2 which creates another new request on the heap which increment the ref count again. Then, the address of #2 on the heap is passed to the C-API, which the second ref count increment will be decremented when the release callback is invoked later, and the first increment is decremented as soon as this set function returns. If this is correct, then we can remove the first ref count increment:
```c++
1  
2  void* user_p = reinterpret_cast<void*>(new std::shared_ptr<PyInferenceRequest>(request));
3  TRITONSERVER_InferenceRequestSetReleaseCallback(triton_object_, ReleaseCallback, user_p /* request_release_userp */);
```
The purpose of the unique_ptr is to protect the `user_p` raw pointer in case there are extra logics added in between #2 and #3 and return earlier making #2 a memory leak, but I think this is unlikely, then maybe it is better to remove the unique_ptr for readability?

updated: [Remove unique_ptr wrapper on release callback](https://github.com/triton-inference-server/core/pull/426/commits/141483fa946cf5aa2610b6ea08facc3cc89bc705)

python/tritonserver/_c/tritonserver_pybind.cc
kthui: No, the C-API does not check for `nullptr`.
https://github.com/triton-inference-server/core/blob/r25.01/src/tritonserver.cc#L831-L838

python/tritonserver/_c/tritonserver_pybind.cc
kthui: my observation: the issue does not happen as long as we are not allocating memory during the static initialization phase

cc @GuanLuo if you have any thought on this?

python/tritonserver/_c/tritonserver_pybind.cc
kthui: sure, we can return the error. [Return if error when setting buffer attributes](https://github.com/triton-inference-server/core/pull/426/commits/c534d3a8a9e2b594b5e584f7c37b151a381fdcbf)

python/tritonserver/_c/tritonserver_pybind.cc
kthui: yes, we can assert the memory type and id. [Assert memory type and id on release function](https://github.com/triton-inference-server/core/pull/426/commits/a8f7bda312521a7d3da57bf1d0e3483f87ea5d87)

python/tritonserver/_c/tritonserver_pybind.cc
```diff
@@ -1175,21 +1162,109 @@ class PyInferenceRequest
...
+
   void Cancel()
   {
     ThrowIfError(TRITONSERVER_InferenceRequestCancel(triton_object_));
   }
 
+  // Response management
+  void GetNextResponse(const py::object& py_future)
+  {
+    // Called exclusively from Python, so GIL is always held.
```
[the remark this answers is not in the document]
kthui: it is ok to hold the GIL explicitly, but there will be extra overhead for pybind to manage the GIL lifecycle with the extra scope lock.

python/tritonserver/_c/tritonserver_pybind.cc
kthui: sure, the comment is updated. [Update comments related to GetNextResponse() and AddNextResponse()](https://github.com/triton-inference-server/core/pull/426/commits/1d4f98df4530d1e0b75ee426f33ff80d035840c0)

python/tritonserver/_c/tritonserver_pybind.cc
kthui: yes, the comments are updated. [Update comments related to GetNextResponse() and AddNextResponse()](https://github.com/triton-inference-server/core/pull/426/commits/1d4f98df4530d1e0b75ee426f33ff80d035840c0)

python/tritonserver/_c/tritonserver_pybind.cc
kthui: the GIL is required when the `response_future_local` object goes out of scope, which decrement the ref count to the future object.

python/tritonserver/_c/tritonserver_pybind.cc
kthui: yes, the ref count of `response_future_` is decremented, if it is managing an object, and the `py_future` ref count is incremented and the additional count is managed by the `response_future_`.
https://github.com/pybind/pybind11/blob/d2e7e8c68711d1ebfb02e2f20bd1cb3bfc5647c0/include/pybind11/pytypes.h#L405-L416

python/tritonserver/_c/tritonserver_pybind.cc
kthui: yes, as long as the user is using the response iterators on the _api.

kthui:
> What is in server certainly won't work, it is tightly related to the server frontend for optimization which is why allocator is up for the user to define.

> You probably can reuse the one for backend but that is kind of strange. Plus the one in backend may not be as well crafted as you expected.. Please evaluate whether it may sense to proceed in this way.

I think we will proceed with a static class in the binding that is singleton and exclusive for the binding, which as support for both CPU and GPU allocation and is optimized for the binding use-case. DLIS-7824

python/test/test_api.py
```diff
@@ -499,3 +645,54 @@ def test_parameters(self, server_options):
...
+        mem_inc = new_current - current
+        assert mem_inc >= min_mem_inc
+        assert mem_inc < max_mem_inc
+        tracemalloc.stop()
+
+    # max_infer_repeat / min_infer_repeat >> max_mem_inc / min_mem_inc, so the memory
+    # increase is not proportional to the increase in inference requests.
+    @pytest.mark.parametrize("infer_repeat", [200, 500, 1000, 8000])
+    def test_inference_leak(self, server_options, infer_repeat):
+        min_mem_inc, max_mem_inc = 20000, 120000
```
rmccorm4, line 668: are these values in bytes? 20KB and 120KB?

python/test/test_api.py
```diff
@@ -499,3 +645,54 @@ def test_parameters(self, server_options):
...
+        current, peak = tracemalloc.get_traced_memory()
+        yield
+        new_current, new_peak = tracemalloc.get_traced_memory()
+        mem_inc = new_current - current
+        assert mem_inc >= min_mem_inc
+        assert mem_inc < max_mem_inc
+        tracemalloc.stop()
+
+    # max_infer_repeat / min_infer_repeat >> max_mem_inc / min_mem_inc, so the memory
+    # increase is not proportional to the increase in inference requests.
```
rmccorm4, line 665: If there is no per-request/response memory leaks, shouldn't sending ~200 requests and ~8000 requests with the same concurrency/batching see roughly the same amount of memory usage before/after?

Do you have more detailed results/findings from local experiments that led you to these numbers and methodology?

python/tritonserver/_api/_response.py
```diff
@@ -231,10 +278,30 @@ def __next__(self):
...
+        return self
+
+    def __next__(self):
+        if self._complete:
+            raise StopIteration
+
+        if self._iterator_queue is None:
+            response = self._get_next_response()
+        else:
+            response = self._iterator_queue.get()
```
rmccorm4, line 298: Why is iterator_queue only used when a user_queue is provided? Would it make sense to always use iterator_queue  for no if/else here? 

One major difference currentlly seems to be only getting next response on demand from iteration calls to `__next__` vs. pre-fetching the responses to iterator queue in a background thread, even if the response iterator isn't iterated on.

Can you elaborate on your thoughts/decision here?

rmccorm4 (commented):
Majority of changes LGTM - but left a few questions to wrap up tomorrow

python/test/test_api.py
kthui: yes, those are in bytes, so 20KB and 120KB

python/test/test_api.py
kthui: I found the memory usage increment will start higher and converge to around 40KB locally on my machine as the test progresses, regardless of the requests count, i.e.
1. requests = 200; memory increment = 71193
2. requests = 500; memory increment = 60405
3. requests = 1000; memory increment = 40133
4. requests = 8000; memory increment = 38519

and in the opposite order:
1. requests = 8000; memory increment = 83592
2. requests = 1000; memory increment = 47596
3. requests = 500; memory increment = 41406
4. requests = 200; memory increment = 40853

requests = 1000 repeat 6x:
1. memory increment = 84436
2. memory increment = 41967
3. memory increment = 40166
4. memory increment = 45353
5. memory increment = 41281
6. memory increment = 34740

python/tritonserver/_api/_response.py
kthui: > Why is iterator_queue only used when a user_queue is provided? Would it make sense to always use iterator_queue for no if/else here?

The code would be a bit simpler if the iterator_queue is always used, but it is not needed if the user did not provide an additional response queue to be feed, so the iterator can pull directly from the binding without the intermediate iterator queue, which in theory will allow for a higher maximum output throughput / lower latency compared to w/ the intermediate iterator queue (and its thread/coroutine).

In other words, we can bypass the intermediate queue when there is only a single consumer for the producer at the binding. If there is more than one consumer for the producer, an additional queue for each consumer is needed to allow those consumer to consume at a different rate without blocking each other. A [key part](https://github.com/triton-inference-server/core/pull/426/files#diff-f4adade0afd240d379f3f6400b2a57103329f0866b8e2c7ac2fd626997725ef0R516-R517) of the test is to make sure the user provided queue is feed with responses without having to iterate over the iterator.

python/test/test_api.py
rmccorm4: Can you add `_bytes` suffix to the var names?

kthui: Sure, added the `_bytes` suffix. [Add _bytes suffix to mem size vars](https://github.com/triton-inference-server/core/pull/426/commits/536cf986c528229345406908f5b520b343061993)

rmccorm4 (approved):

rmccorm4:
Let's prioritize the follow-ups for consistent tests (grpc fork errors), gpu memory support, etc

rmccorm4:
Nice work Jacky! Thanks for pushing this through!

PR #434, opened 2025-04-23
fix: Input batch size overflow vulnerability

yinggeh:
#### What does the PR do?
1. Cast uint32_t (batch_size_) to int32_t is not safe. Cast to int64_t instead. Before the change, request with overflowed batch size will bypass core check and propagate to backend. Although backend will catch the error, it's still better to stop the request at early stage. 
2. Remove input_memory_type from `ValidateBytesInputs` parameters as it's not used.

#### Checklist

#### Commit Type:
- [x] fix

#### Related PRs:
https://github.com/triton-inference-server/server/pull/8165

#### Test plan:
L0_input_validation--base

- CI Pipeline ID:
27387023

- closes GitHub issue: #xxx

### DIFF
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -1076,7 +1076,7 @@ InferenceRequest::Normalize()
 
   // Make sure request batch-size doesn't exceed what is supported by
   // the model.
-  if ((int)batch_size_ > model_config.max_batch_size()) {
+  if (static_cast<int64_t>(batch_size_) > model_config.max_batch_size()) {
     return Status(
         Status::Code::INVALID_ARG,
         LogRequest() + "inference request batch-size must be <= " +
@@ -1201,10 +1201,8 @@ InferenceRequest::Normalize()
       // Non-linear IO format input byte size validation will be handled in the
       // TensorRT backend.
       if (!input.IsNonLinearFormatIo()) {
-        TRITONSERVER_MemoryType input_memory_type;
         if (data_type == inference::DataType::TYPE_STRING) {
-          RETURN_IF_ERROR(ValidateBytesInputs(
-              input_name, input, model_name, &input_memory_type));
+          RETURN_IF_ERROR(ValidateBytesInputs(input_name, input, model_name));
         } else {
           // Shape tensor with dynamic batching does not introduce a new
           // dimension to the tensor but adds an additional value to the 1-D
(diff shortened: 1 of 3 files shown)


src/infer_request.cc
```diff
@@ -1076,7 +1076,7 @@ InferenceRequest::Normalize()
 
   // Make sure request batch-size doesn't exceed what is supported by
   // the model.
-  if ((int)batch_size_ > model_config.max_batch_size()) {
```
yinggeh, line 1079: `batch_size_` is `uint32_t` type.

src/test/input_byte_size_test.cc
```diff
@@ -757,6 +757,59 @@ TEST_F(InputByteSizeTest, StringCountMismatchGPU)
...
+
+  // Set response callback
+  FAIL_TEST_IF_ERR(
+      TRITONSERVER_InferenceRequestSetResponseCallback(
+          irequest_, allocator_, nullptr /* response_allocator_userp */,
+          InferResponseComplete, reinterpret_cast<void*>(&p)),
+      "setting response callback");
+
+  // Run inference
+  FAIL_TEST_IF_SUCCESS(
```
yinggeh, line 802: Before the change, request with overflowed batch size will bypass core check and propagate to backend.
Although backend returns error "error: response status: Internal - batch size 2147483648 for 'onnx_zero_1_bool_0', max allowed is 8", it's still better to stop the request at early stage.

yinggeh:
No HTTP or gRPC client test because input size is gigantic. The minimum input byte size to validate, assuming bool input with shape [-1,1], is 2^31 bytes or 2 GB.

In gRPC client, the maximum byte size allowed is exactly 2^31-1 bytes.
In http client, request stops at server with error
`tritonclient.utils.InferenceServerException: [400] inference header size should be in range (0, -2147483525), got: 155`.
In shared memory, returns error when shared memory size is 1 GB.
```
Bus error (core dumped)
UserWarning: resource_tracker: There appear to be 1 leaked shared_memory objects to clean up at shutdown
```

src/test/input_byte_size_test.cc
ziqif-nv: which part of your change would `stop the request at early stage` as you mentioned?

or, it will just prevent cast overflow

yinggeh: Before the fix, error msg is `"error: response status: Internal - batch size 2147483648 for 'onnx_zero_1_bool_0', max allowed is 8"` returning from onnxruntime_backend check.

After the fix, error msg becomes `"inference request batch-size must be <=  8  for 'onnx_zero_1_bool_0'"` returning from core check.

ziqif-nv: ah, got it now. w/o your change, after the cast to int32, it would slip through the check

ziqif-nv (approved):
LGTM

src/test/input_byte_size_test.cc
yinggeh: Correct!

PR #429, opened 2025-03-12
feat: Ensemble async callback execution

yinggeh:
#### What does the PR do?
Reduce e2e latency in ensemble model by executing callbacks asynchronously at the end of each ensemble step.

#### Checklist
- [x] PR title reflects the change and is of format `<commit_type>: <Title>`
- [x] Changes are described in the pull request.
- [x] Related issues are referenced.
- [x] Populated [github labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels) field
- [ ] Added [test plan](#test-plan) and verified test passes.
- [ ] Verified that the PR passes existing CI.
- [x] Verified copyright is correct on all changed files.
- [ ] Added _succinct_ git squash message before merging [ref](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
- [ ] All template sections are filled out.
- [ ] Optional: Additional screenshots for behavior/output changes with before/after.

#### Commit Type:
Check the [conventional commit type](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type)
box here and add the label to the github PR.
- [x] feat

#### Related PRs:
https://github.com/triton-inference-server/common/pull/133

#### Test plan:
L0_simple_ensemble

- CI Pipeline ID:
25280555

#### Related Issues: (use one of the action keywords Closes / Fixes / Resolves / Relates to)
- closes GitHub issue: #7650

PR #435, opened 2025-04-29
fix: Revert async execution of ensemble model ResponseComplete callback

yinggeh:
#### What does the PR do?
Fixes `L0_backend_python--base` and `L0_sequence_batcher--base`.

Initially ensemble asynchronous callback feature was introduced in https://github.com/triton-inference-server/core/pull/429 in order to reduce latency in overheads added in the ensemble pipeline.

Asynchrous `ResponseComplete` didn't work in decoupled model because the order of responses cannot be guarenteed. In rare case, `TRITONSERVER_RESPONSE_COMPLETE_FINAL` is sent before other responses and causes segfault.

Even though I disable async call of `ResponseComplete` in decoupled model, it's still failing one test case in `L0_sequence_batcher`. So I ended up reverting async ResponseComplete callback feature. But async RequestComplete callback is working fine.

#### Checklist
- [ ] Verified copyright is correct on all changed files.

#### Commit Type:
- [x] fix

#### Test plan:

- CI Pipeline ID:
27706423

- closes GitHub issue: #xxx

### DIFF
--- a/src/constants.h
+++ b/src/constants.h
@@ -62,7 +62,7 @@ constexpr char kPythonBackend[] = "python";
 
 #ifdef TRITON_ENABLE_ENSEMBLE
 constexpr char kEnsemblePlatform[] = "ensemble";
-constexpr uint64_t ENSEMBLE_CB_POOL_SIZE = 8u;
+constexpr uint64_t ENSEMBLE_CB_POOL_SIZE = 4u;
 #endif  // TRITON_ENABLE_ENSEMBLE
 
 constexpr char kTensorRTExecutionAccelerator[] = "tensorrt";
--- a/src/ensemble_scheduler/ensemble_scheduler.cc
+++ b/src/ensemble_scheduler/ensemble_scheduler.cc
@@ -331,8 +331,6 @@ class EnsembleContext {
   void CacheEnsembleTopLevelRequest(
       std::unique_ptr<InferenceResponse>& response);
 
-  triton::common::ThreadPool* CallbackPool() const { return callback_pool_; }
-
   InferenceServer* is_;
 
   EnsembleInfo* info_;
@@ -382,10 +380,6 @@ class EnsembleContext {
       TRITONSERVER_ResponseAllocator,
       decltype(&TRITONSERVER_ResponseAllocatorDelete)>
       allocator_;
-
-  // The thread pool used to execute ensemble callbacks and reduce e2e latency.
-  // The thread pool is managed by InferenceServer.
-  triton::common::ThreadPool* const callback_pool_;
 };
 
 EnsembleContext::EnsembleContext(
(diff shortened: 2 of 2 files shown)


src/constants.h
```diff
@@ -62,7 +62,7 @@ constexpr char kPythonBackend[] = "python";
 
 #ifdef TRITON_ENABLE_ENSEMBLE
 constexpr char kEnsemblePlatform[] = "ensemble";
-constexpr uint64_t ENSEMBLE_CB_POOL_SIZE = 8u;
```
yinggeh, line 65: Decrease thread pool size because there will be fewer callbacks running asynchronusly.

oandreeva-nv: why?

GuanLuo: Do you even need this constant?

GuanLuo:
> So I ended up reverting async ResponseComplete callback feature. But async RequestComplete callback is working fine.

Why `RequestComplete` is also made async at the first place?

src/constants.h
yinggeh: Yes. Because RequestComplete callback is still running asynchrously, which still benefits the model throughput.

yinggeh: https://github.com/triton-inference-server/core/blob/729422e4a9feafdd80a8c7c5c0e08dfbe5aa391b/src/ensemble_scheduler/ensemble_scheduler.cc#L608-L632

yinggeh:
> > So I ended up reverting async ResponseComplete callback feature. But async RequestComplete callback is working fine.
> 
> Why `RequestComplete` is also made async at the first place?

Explanation with visualization:
Screenshot from 2025-04-29 15-42-15
From the chart, this composing model `preprocessing` backend thread takes 62 usec to complete, including 24 usec `EnsembleContext::ResponseComplete` and 7 usec `EnsembleContext::RequestComplete`.

Maximum throughput of sample ensemble_model
Both synchronous (no optimization): 39794.5 infer/sec
Synchronous `EnsembleContext::ResponseComplete` and asynchronous `EnsembleContext::ResponseComplete` (this PR): 47017.2 infer/sec

src/constants.h
oandreeva-nv: how was this number picked and why this number?

yinggeh: The previous pool size 8 was explained here https://github.com/triton-inference-server/core/pull/429#discussion_r2023453242.
In a non-decoupled ensemble model, this PR will reduce half of async callbacks. Thus I reduce pool size by half to 4.

oandreeva-nv: My concern is how general those experiments were?

yinggeh: Could you elaborate? If thread pool queue is full, new task will execute synchronously like normal to avoid delay. See https://github.com/triton-inference-server/core/blob/729422e4a9feafdd80a8c7c5c0e08dfbe5aa391b/src/ensemble_scheduler/ensemble_scheduler.cc#L625-L632

PR #483, opened 2026-03-30
fix: RequestTracker counter mismatch

mattwittwer:
#### What does the PR do?

Fixes a `RequestTracker` counter and lifetime mismatch in `ScheduleSteps()` that could prematurely release the top-level ensemble request when a parallel step failed to enqueue, typically surfacing after `FAILED_ENQUEUE` / `Exceeds maximum queue size` as a SIGSEGV during ensemble error finalization.

Issue
When multiple steps were prepared in a single `ScheduleSteps()` pass, the per-step cleanup path always called `RequestTracker::DecrementCounter()`, even for steps that never called `IncrementCounter()` because scheduling had already failed or been skipped. That could corrupt `inflight_request_counter_` and release the top-level request while `FinishEnsemble()` was still using it, leaving later error-handling and cleanup paths operating on invalid request-tracker state.

Fix
Preserve the shared per-step `max_inflight_requests` limiter flow from `main` while correcting the failure path in `ScheduleSteps()`.

Keep the top-level `RequestTracker` alive across async release and finalization using `std::shared_ptr`, and route cancellation, logging, and error response paths through synchronized helper methods.

Pass a heap-allocated `RequestTrackerReference` through the C release callback and explicitly clean it up for prepared steps that never reach the callback-owned release path.

Only call `RequestTracker::DecrementCounter()` for steps that actually called `IncrementCounter()`, while still decrementing `inflight_step_counter_` for unscheduled steps and releasing any acquired limiter slot.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:

https://github.com/triton-inference-server/server/pull/8722

The request processing follows this order:
PrepareSteps() in src/ensemble_scheduler/ensemble_scheduler.cc: 1011-1015
ScheduleSteps() decision to schedule and take the lifetime ref: 1467-1474
ScheduleSteps() handoff to InferAsync() and failure capture: 1496-1506
ScheduleSteps() local cleanup for unscheduled / failed-dispatch steps: 1512-1528
FinishEnsemble() final release of the top-level tracker: 1292-1297

#### Test plan:

Test case here:
https://github.com/triton-inference-server/server/pull/8722

- CI Pipeline ID:

47868062

- closes GitHub issue: #xxx

### DIFF
--- a/src/ensemble_scheduler/ensemble_scheduler.cc
+++ b/src/ensemble_scheduler/ensemble_scheduler.cc
@@ -87,7 +87,24 @@ class RequestTracker {
   {
   }
 
+  // Unsynchronized access -- only safe under EnsembleContext::mutex_ during
+  // normal (non-finalization) paths where no concurrent Release can occur.
   std::unique_ptr<InferenceRequest>& Request() { return request_; }
+
+  // Thread-safe accessors for the error-finalization path where request_
+  // may be concurrently moved by DecrementCounter -> Release.
+  bool IsCancelled()
+  {
+    std::lock_guard<std::mutex> lk(mtx_);
+    return (request_ == nullptr) || request_->IsCancelled();
+  }
+
+  std::string LogRequest()
+  {
+    std::lock_guard<std::mutex> lk(mtx_);
+    return (request_ != nullptr) ? request_->LogRequest()
+                                 : std::string("[request released] ");
+  }
 
   InferenceStatsAggregator* StatsAggregator() { return stats_aggregator_; }
 
@@ -141,6 +158,23 @@ class RequestTracker {
     status_ = status;
   }
 
+  void RespondIfError(const Status& status, FailureReason reason)
+  {
+    std::lock_guard<std::mutex> lk(mtx_);
+    if (request_ != nullptr) {
+      InferenceRequest::RespondIfError(
+          request_, status, false /* release_request */, reason);
+    }
+  }
+
+  void SendFlags(const uint32_t flags)
+  {
+    std::lock_guard<std::mutex> lk(mtx_);
+    if (request_ != nullptr) {
+      request_->ResponseFactory()->SendFlags(flags);
+    }
+  }
+
  private:
   std::mutex mtx_;
   uint32_t inflight_request_counter_;
@@ -153,6 +187,7 @@ class RequestTracker {
   triton::common::ThreadPool* const callback_pool_;
 };
 
+using RequestTrackerReference = std::shared_ptr<RequestTracker>;
 // Step is used as 'userp' and keeps ensemble context alive
 // until no more internal requests are inflight.
 // Step contains metadata, and status for the
@@ -188,6 +223,12 @@ struct Step {
   const bool preserve_responses_order_;
 
   size_t step_idx_;
+
+  // Raw pointer to the heap-allocated RequestTrackerReference passed as
+  // release callback userp.  Normally freed by RequestComplete when the
+  // request is released; stored here so ScheduleSteps can clean it up if
+  // InferAsync fails and the release callback will never fire.
+  RequestTrackerReference* tracker_ref_{nullptr};
 };
 
 struct TensorData {
@@ -396,7 +437,7 @@ class EnsembleContext {
 
   // Objects related to the ensemble infer request
   Status ensemble_status_;
-  RequestTracker* request_tracker_;
+  std::shared_ptr<RequestTracker> request_tracker_;
   // Use in conjunction with 'is_decoupled_' in EnsembleInfo to
   // better distinguish ensemble ending behavior (see annotation in
   // FinishEnsemble for details).
@@ -429,7 +470,7 @@ EnsembleContext::EnsembleContext(
 {
   uint64_t compute_start_ns = 0;
   INFER_STATS_SET_TIMESTAMP(compute_start_ns);
-  request_tracker_ = new RequestTracker(
+  request_tracker_ = std::make_shared<RequestTracker>(
       std::move(request), compute_start_ns, metric_reporter, stats_aggregator,
       callback_pool);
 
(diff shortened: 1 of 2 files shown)


src/ensemble_scheduler/ensemble_scheduler.cc
```diff
@@ -1070,12 +1109,18 @@ EnsembleContext::InitStep(
...
 #endif
+  // Heap-allocate a shared_ptr to the tracker so it can be passed through the
+  // C callback API (void* userp).  Ownership transfers to the callback on
+  // successful PrepareForInference; cleaned up by unique_ptr on early return.
+  auto request_tracker_ref =
+      std::make_unique<RequestTrackerReference>(request_tracker_);
   irequest->SetResponseCallback(
       reinterpret_cast<ResponseAllocator*>(allocator_.get()), step->get(),
       ResponseComplete, step->get());
-  irequest->SetReleaseCallback(RequestComplete, request_tracker_);
```
whoisj, line 1076: what does the release callback do? if it calls `free()` on `request_tracker_` we have an ownership problem because the `std::shared_ptr<T>` will do the same.

mattwittwer: `RequestComplete()` does not directly free the `RequestTracker` object.

In [`InitStep()`](https://github.com/triton-inference-server/core/blob/5002e706b7c433295f7da578a6e24df428bf5813/src/ensemble_scheduler/ensemble_scheduler.cc#L1112-L1123), the release callback `userp` is a heap-allocated [`RequestTrackerReference`](https://github.com/triton-inference-server/core/blob/5002e706b7c433295f7da578a6e24df428bf5813/src/ensemble_scheduler/ensemble_scheduler.cc#L189) (`std::shared_ptr<RequestTracker>`), not the raw `RequestTracker*`. Then in [`RequestComplete()`](https://github.com/triton-inference-server/core/blob/5002e706b7c433295f7da578a6e24df428bf5813/src/ensemble_scheduler/ensemble_scheduler.cc#L684-L698), we destroy that heap-allocated shared_ptr wrapper, which only drops one shared reference.

The callback also deletes the internal step `TRITONSERVER_InferenceRequest*` and calls [`DecrementCounter()`](https://github.com/triton-inference-server/core/blob/5002e706b7c433295f7da578a6e24df428bf5813/src/ensemble_scheduler/ensemble_scheduler.cc#L125-L151), but that does not `delete` the `RequestTracker` object. The tracker lifetime is owned by [`EnsembleContext::request_tracker_`](https://github.com/triton-inference-server/core/blob/5002e706b7c433295f7da578a6e24df428bf5813/src/ensemble_scheduler/ensemble_scheduler.cc#L435-L437) plus any outstanding callback-held references, and the context drops its own reference later in [`FinishEnsemble()`](https://github.com/triton-inference-server/core/blob/5002e706b7c433295f7da578a6e24df428bf5813/src/ensemble_scheduler/ensemble_scheduler.cc#L1292-L1297). There is also an early-failure cleanup in [`ScheduleSteps()`](https://github.com/triton-inference-server/core/blob/5002e706b7c433295f7da578a6e24df428bf5813/src/ensemble_scheduler/ensemble_scheduler.cc#L1518-L1528) if `InferAsync()` fails before the release callback runs.

So there should not be a double-free here: the callback no longer does `delete request_tracker`; it only drops its shared reference.

src/ensemble_scheduler/ensemble_scheduler.cc
```diff
@@ -1070,12 +1109,18 @@ EnsembleContext::InitStep(
...
+  irequest->SetReleaseCallback(RequestComplete, request_tracker_ref.get());
 
   RETURN_IF_ERROR(irequest->PrepareForInference());
+  (*step)->tracker_ref_ = request_tracker_ref.release();
```
whoisj: what is happening here?

are you releasing ownership of an allocation managed by `request_tracker_ref` and assigning ownership to `tracker_ref_` at the same time?

Why is `request_tracking_ref` not a `std::shared_ptr` to begin with? Seems to me that it would make this whole dance less confusing and more obvious as to what is happening here.

Is there a reason to use `std::unique_ptr`?

whoisj (changes requested):

src/ensemble_scheduler/ensemble_scheduler.cc
mattwittwer: Good question. `RequestTrackerReference` is already a `std::shared_ptr<RequestTracker>` ([here](https://github.com/triton-inference-server/core/blob/mwittwer/ensemble-schedule-steps-counter-cleanup/src/ensemble_scheduler/ensemble_scheduler.cc#L189)), so the `std::unique_ptr` in [`InitStep()`](https://github.com/triton-inference-server/core/blob/mwittwer/ensemble-schedule-steps-counter-cleanup/src/ensemble_scheduler/ensemble_scheduler.cc#L1113-L1125) is not owning the `RequestTracker` itself. It is only owning the heap allocation that we pass through the C `void* userp` release-callback API.

So there are two layers:

- the inner `shared_ptr<RequestTracker>` keeps the actual `RequestTracker` alive
- the outer heap allocation is just the callback payload that gives `SetReleaseCallback()` a stable address to store in `userp`

The flow is:

1. [`InitStep()`](https://github.com/triton-inference-server/core/blob/mwittwer/ensemble-schedule-steps-counter-cleanup/src/ensemble_scheduler/ensemble_scheduler.cc#L1113-L1125) allocates a heap `RequestTrackerReference` from `request_tracker_`, which adds one shared reference to the tracker.
2. Before `PrepareForInference()` succeeds, that heap allocation is owned by the local `unique_ptr`, so an early return cleans it up automatically.
3. After `PrepareForInference()` succeeds, ownership of the heap allocation is transferred to [`step->callback_tracker_ref_`](https://github.com/triton-inference-server/core/blob/mwittwer/ensemble-schedule-steps-counter-cleanup/src/ensemble_scheduler/ensemble_scheduler.cc#L1124-L1125).
4. On the normal path, [`RequestComplete()`](https://github.com/triton-inference-server/core/blob/mwittwer/ensemble-schedule-steps-counter-cleanup/src/ensemble_scheduler/ensemble_scheduler.cc#L685-L698) adopts that raw pointer into a local `unique_ptr`, which deletes the heap allocation when the callback exits.
5. On the failure path where the request never reaches the callback-owned release path, [`ScheduleSteps()`](https://github.com/triton-inference-server/core/blob/mwittwer/ensemble-schedule-steps-counter-cleanup/src/ensemble_scheduler/ensemble_scheduler.cc#L1520-L1529) deletes `step->callback_tracker_ref_` directly.
6. Later, [`FinishEnsemble()`](https://github.com/triton-inference-server/core/blob/mwittwer/ensemble-schedule-steps-counter-cleanup/src/ensemble_scheduler/ensemble_scheduler.cc#L1294-L1299) drops the context's own `shared_ptr` with `request_tracker_.reset()`.

So this is not a double-free of `RequestTracker`: the single-owner heap callback payload is destroyed in exactly one place, and destroying that payload only drops one `shared_ptr<RequestTracker>` reference.

whoisj (approved):
LGTM

PR #489, opened 2026-04-12
fix: Respect cc_model_filenames in Python backend auto-fill

nightflight-dk:
### Problem

When `backend` is set to `"python"` and `default_model_filename` is left empty, `AutoCompleteBackendFields` unconditionally sets `default_model_filename` to `"model.py"`. This ignores any `cc_model_filenames` mappings the user may have configured to use a custom model filename (e.g., `custom_model.py` per device).

This causes the Python backend stub to always look for `model.py`, even when `cc_model_filenames` explicitly maps to a different file.

### Fix

In `model_config_utils.cc`, the auto-fill of `default_model_filename` now checks that `cc_model_filenames` is also empty before defaulting to `"model.py"`:

```cpp
if (config->default_model_filename().empty() &&
    config->cc_model_filenames().empty()) {
  config->set_default_model_filename(kPythonFilename);
}

### DIFF
--- a/src/test/model_config_utils_test.cc
+++ b/src/test/model_config_utils_test.cc
@@ -1,0 +1,178 @@
+// Copyright 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions
+// are met:
+//  * Redistributions of source code must retain the above copyright
+//    notice, this list of conditions and the following disclaimer.
+//  * Redistributions in binary form must reproduce the above copyright
+//    notice, this list of conditions and the following disclaimer in the
+//    documentation and/or other materials provided with the distribution.
+//  * Neither the name of NVIDIA CORPORATION nor the names of its
+//    contributors may be used to endorse or promote products derived
+//    from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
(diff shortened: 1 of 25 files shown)

### CODE REVIEW

**Comment by nightflight-dk** [comment]
@whoisj this helps provide cc dependent implementation part of the same repo

**Comment by whoisj** [comment]
LGTM. I'll run it through our internal CI lab.

@nightflight-dk, could you please address the pre-commit issues? Thanks.

CI Pipeline ID: 49121709

you can always run pre-commit locally via

```pre-commit run --files <file you've changed>```

File: src/test/model_config_utils_test.cc [hunk 1]
```diff
@@ -0,0 +1,153 @@
...
+
+  auto status = tc::AutoCompleteBackendFields("test_model", dir.Path(), &config);
+  ASSERT_TRUE(status.IsOk()) << status.AsString();
+  EXPECT_EQ(config.default_model_filename(), "")
+      << "default_model_filename should remain empty when cc_model_filenames "
+         "is set";
+}
+
+// When backend is "python" and default_model_filename is already set,
+// it should be preserved regardless of cc_model_filenames.
```
yinggeh, line 122: Add a test case where both `default_model_filename` and  `cc_model_filenames` are set.

nightflight-dk: done

whoisj (approved):
LGTM. @yinggeh are you satisfied with the requested test case?

src/test/model_config_utils_test.cc
```diff
@@ -0,0 +1,178 @@
...
+  config.set_backend("python");
+  (*config.mutable_cc_model_filenames())["gpu"] = "custom_model.py";
+  // default_model_filename is empty, cc_model_filenames is set
+
+  auto status =
+      tc::AutoCompleteBackendFields("test_model", dir.Path(), &config);
+  ASSERT_TRUE(status.IsOk()) << status.AsString();
+  EXPECT_EQ(config.default_model_filename(), "")
+      << "default_model_filename should remain empty when cc_model_filenames "
+         "is set";
```
yinggeh, line 118: In this case, which python file backend should load if not "gpu"?

PR #497, opened 2026-05-12
fix: Modernize Child Path Detection

whoisj:
This change updates the algorithm used to detect if a 'child path' escapes its parent.

### DIFF
--- a/src/filesystem/api.cc
+++ b/src/filesystem/api.cc
@@ -64,6 +64,8 @@ LocalizedPath::~LocalizedPath()
 
 namespace {
 
+namespace fs = std::filesystem;
+
 class FileSystemManager {
  public:
   Status GetFileSystem(
@@ -413,15 +415,20 @@ IsChildPathEscapingParentPath(
   std::string canonical_child_path;
   std::string canonical_parent_path;
   try {
-    canonical_child_path =
-        std::filesystem::weakly_canonical(child_path).string();
-    canonical_parent_path = std::filesystem::canonical(parent_path).string();
+    canonical_child_path = fs::weakly_canonical(child_path).string();
+  }
+  catch (const std::exception& ex) {
+    return Status(
+        Status::Code::INVALID_ARG,
+        "Failed to canonicalize child path '" + child_path + "': " + ex.what());
+  }
+  try {
+    canonical_parent_path = fs::canonical(parent_path).string();
   }
   catch (const std::exception& e) {
     return Status(
-        Status::Code::INVALID_ARG,
-        "Invalid path comparison (child:" + child_path +
-            ", parent:" + parent_path + "): " + e.what());
+        Status::Code::INVALID_ARG, "Failed to canonicalize parent path '" +
+                                       parent_path + "': " + e.what());
   }
   // Determine if the parent path is the prefix of the child path.
   // Ensure that the paths are identical or the first mismatch character in the
@@ -429,13 +436,23 @@ IsChildPathEscapingParentPath(
   // (i.e. "/parent" and "/parent_dir").
   if (canonical_parent_path.find(canonical_child_path, 0) == 0 &&
       ((canonical_child_path.size() > canonical_parent_path.size() &&
-        canonical_child_path[canonical_parent_path.size()] == '/') ||
+        IsPathSeparator(canonical_child_path[canonical_parent_path.size()])) ||
        (canonical_child_path.size() == canonical_parent_path.size()))) {
     *is_escaped = false;
     return Status::Success;
   }
   *is_escaped = true;
   return Status::Success;
+}
+
+bool
+IsPathSeparator(char c)
+{
+#if defined(_WIN32)
+  return c == '/' || c == '\\';
+#else
+  return c == '/';
+#endif
 }
 
 std::string
--- a/src/filesystem/api.cc
+++ b/src/filesystem/api.cc
@@ -436,7 +436,7 @@ IsChildPathEscapingParentPath(
   // (i.e. "/parent" and "/parent_dir").
   if (canonical_parent_path.find(canonical_child_path, 0) == 0 &&
       ((canonical_child_path.size() > canonical_parent_path.size() &&
-        IsPathSeparator(canonical_child_path[canonical_parent_path.size()])) ||
+        canonical_child_path[canonical_parent_path.size()] == '/') ||
        (canonical_child_path.size() == canonical_parent_path.size()))) {
     *is_escaped = false;
     return Status::Success;
--- a/src/filesystem/api.cc
+++ b/src/filesystem/api.cc
@@ -445,16 +445,6 @@ IsChildPathEscapingParentPath(
   return Status::Success;
 }
 
-bool
-IsPathSeparator(char c)
-{
-#if defined(_WIN32)
-  return c == '/' || c == '\\';
-#else
-  return c == '/';
-#endif
-}
-
 std::string
 JoinPath(std::initializer_list<std::string> segments)
 {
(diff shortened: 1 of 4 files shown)


src/filesystem/api.cc
```diff
@@ -404,27 +404,35 @@ IsChildPathEscapingParentPath(
...
+    absolute_parent_path = std::filesystem::canonical(parent_path);
   }
   catch (const std::exception& e) {
     return Status(
         Status::Code::INVALID_ARG,
-        "Nonexistent path '" + parent_path + "': " + e.what());
+        "Nonexistent path '" + parent_path.string() + "': " + e.what());
+  }
+  static const std::filesystem::path empty_path;
+  while (absolute_child_path != empty_path) {
```
mudit-eng: !absolute_child_path.empty()

src/filesystem/api.cc
```diff
@@ -404,27 +404,35 @@ IsChildPathEscapingParentPath(
...
+    if (std::filesystem::equivalent(
+            absolute_child_path, absolute_parent_path)) {
```
mudit-eng: Do we need to validate absolute_parent_path is not empty?

> Unlike weakly_canonical(), equivalent() requires both paths to exist.
> 
> If either path does not exist, it throws std::filesystem::filesystem_error.

src/filesystem/api.cc
```diff
@@ -404,27 +404,35 @@ IsChildPathEscapingParentPath(
...
+      *is_escaped = false;
+      return Status::Success;
+    }
+    absolute_child_path = absolute_child_path.parent_path();
```
mudit-eng: How long can a path be? Would this loop get really expensive if path is really long like /a/b/c/d/e/f/g/h/i/j/k.......?

mudit-eng: May be put an example in the comment to show why this is needed.

mudit-eng:
How do we validate this code works?

src/filesystem/api.cc
whoisj: a path can be infinite length, but most distros limit it to 4096 characters with 256 characters maximum per segment.

src/filesystem/api.cc
```diff
@@ -404,27 +405,35 @@ IsChildPathEscapingParentPath(
...
   }
   catch (const std::exception& e) {
     return Status(
         Status::Code::INVALID_ARG,
         "Nonexistent path '" + parent_path + "': " + e.what());
   }
-  // Can use starts_with() over rfind() in C++20.
-  *is_escaped = absolute_child_path.rfind(absolute_parent_path, 0) != 0;
+  static const std::filesystem::path empty_path;
+  while (canonical_child_path != empty_path) {
```
yinggeh: This recursive approach becomes inefficient for long file paths. Why not subtract canonical_child_path from canonical_parent_path at the beginning and check if the remainder substring is empty or starts with `/`?

src/model_repository_manager/model_repository_manager.cc
```diff
@@ -162,7 +162,10 @@ class LocalizeRepoAgent : public TritonRepoAgent {
 
               // Resolve any relative paths or symlinks, and enforce that target
               // directory stays within model directory for security.
-              if (file_path.rfind(temp_dir, 0) != 0) {
+              bool escapes{false};
```
yinggeh: `escapes` should be `true` by default unless set by `IsChildPathEscapingParentPath`.

src/filesystem/api.cc
whoisj: this isn't recursive, it's iterative. but I get your point.

src/model_repository_manager/model_repository_manager.cc
whoisj: fixed up.

src/filesystem/api.cc
whoisj: fixed up.

src/filesystem/api.cc
whoisj: done.

whoisj:
> How do we validate this code works?

One would assume our CI tests cover most if not all cases. I'll see to it that a new one with the case that this PR fixes is added.

src/filesystem/api.cc
```diff
@@ -405,33 +405,34 @@ IsChildPathEscapingParentPath(
...
+        "Invalid path comparison (child:" + child_path +
+            ", parent:" + parent_path + "): " + e.what());
+  }
+  // Determine if the parent path is the prefix of the child path.
+  // Ensure that the paths are identical or the first mismatch character in the
+  // child path is a path separator ('/') to avoid partial matching
+  // (i.e. "/parent" and "/parent_dir").
+  if (canonical_parent_path.find(canonical_child_path, 0) == 0 &&
+      ((canonical_child_path.size() > canonical_parent_path.size() &&
+        canonical_child_path[canonical_parent_path.size()] == '/') ||
```
yinggeh: Note that with this change, we will drop support on Windows completely.

src/filesystem/api.cc
```diff
@@ -405,33 +405,34 @@ IsChildPathEscapingParentPath(
...
-  }
-  static const std::filesystem::path empty_path;
-  while (canonical_child_path != empty_path) {
-    if (std::filesystem::equivalent(
-            canonical_child_path, canonical_parent_path)) {
-      *is_escaped = false;
-      return Status::Success;
-    }
-    canonical_child_path = canonical_child_path.parent_path();
+        "Invalid path comparison (child:" + child_path +
```
yinggeh: > "Invalid path comparison"

I don't get it. One of the exception is when the file name is greater than 256 characters.

src/model_repository_manager/model_repository_manager.cc
yinggeh: There is one more place need to fix up

src/filesystem/api.cc
whoisj: That should be fine as we officially dropped support in January.

whoisj: Fixed up regardless.

src/filesystem/api.cc
whoisj: I've separated the `try { ... } catch` blocks, hopefully the error message is more helpful now.

src/model_repository_manager/model_repository_manager.cc
whoisj: fixed up.

src/filesystem/api.h
```diff
@@ -132,12 +132,17 @@ bool IsAbsolutePath(const std::string& path);
...
+/// path, false if the child path is within its parent path.
+/// \return Error status
 Status IsChildPathEscapingParentPath(
     const std::string& child_path, const std::string& parent_path,
     bool* is_escaped);
 
+/// Check if a character is a path separator.
+/// \param c The character to check.
+/// \return `true` when the character is a path separator, otherwise `false`.
+bool IsPathSeparator(char c);
```
yinggeh:
```suggestion
inline bool IsPathSeparator(char c);
```

whoisj: why? it's not like the value could be modified by the function. 😕

src/filesystem/api.h
```diff
@@ -141,7 +141,15 @@ Status IsChildPathEscapingParentPath(
 /// Check if a character is a path separator.
 /// \param c The character to check.
 /// \return `true` when the character is a path separator, otherwise `false`.
-bool IsPathSeparator(char c);
+inline bool
```
yinggeh: I believe the function should be in anonymous namespace in `api.cc`  instead of making a public interface of `class LocalizedPath`.

mudit-eng: I wonder if there is a need of this function given that Triton does not have Windows support.

whoisj: I just removed it.

yinggeh (approved):

PR #502, opened 2026-06-06
fix: Model readiness for namespaced models

yinggeh:
#### What does the PR do?
Fixes the `L0_model_namespacing--base` and `L0_metrics--base` tests. Strict server readiness (`InferenceServer::IsReady`) and `InferenceServer::ModelIsReady` resolved models by name only, which is ambiguous when model namespacing is enabled and the same model name exists in multiple namespaces — the lookup fails and the server is reported not-ready. Readiness now resolves models by the full `ModelIdentifier`.

This regression was introduced by #473 (`fix: /v2/health/ready returns 200 when Python backend stub is dead (#8604)`, commit 53fc26e), which added the runtime backend readiness check in the strict-readiness path using name-only model lookup.

#### Checklist
- [x] Populated github labels field
- [x] Added test plan and verified test passes.
- [ ] Added succinct git squash message before merging.

#### Commit Type:
- [x] fix

#### Related PRs:
- #473 — introduced the regression this PR fixes.

- `src/server.cc` / `src/server.h`: `IsReady` strict-readiness loop now resolves each model via the full `ModelIdentifier` (`GetModel(mv.first, ...)`); `ModelIsReady` takes an already-resolved `Model&`.
- `src/tritonserver.cc`: `TRITONSERVER_ServerModelIsReady` resolves the model (name-based, matching the public C-API contract) and calls the new `ModelIsReady` overload.

#### Test plan:
- `L0_model_namespacing--base`
- `L0_metrics--base`

- CI Pipeline ID: 53863861

#### Caveats:
The public C-API `TRITONSERVER_ServerModelIsReady` still resolves by name (no namespace parameter in the API contract); namespaced readiness resolution applies to the internal strict-readiness path.

#### Related Issues:
N/A

### DIFF
--- a/src/server.cc
+++ b/src/server.cc
@@ -1,4 +1,4 @@
-// Copyright 2018-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2018-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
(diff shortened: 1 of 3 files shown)


whoisj (approved):
LGTM

PR #506, opened 2026-06-27
Fix implicit sequence state output mapping

### LINKED ISSUE #8586: [bug] Implicit sequence state mapping swaps states when output_name lexicographic order differs from input_name order
When using **sequence batching + implicit state** with multiple state tensors, Triton can **swap cached states** across requests if the **lexicographic order of** `output_name`s differs from the lexicographic order of `input_name`s. This manifests as Triton injecting the wrong state tensor into the wrong model input on the 2nd request of a sequence.

A concrete failing pair is:

* `cache_last_channel` (FP16, 4D)
* `cache_last_channel_len` (INT64 or INT32, 2D)

With outputs:

* `cache_last_channel_next`
* `cache_last_channel_len_next`

On request number 2, Triton injects the **len** state into `cache_last_channel` and the **channel** state into `cache_last_channel_len`, producing a TensorRT binding dimension error and a client-side “invalid shape for input” failure.

**Workaround:** Renaming the “\_len” state to avoid being a prefix-extension of the other state name (e.g., `cache_last_chan_len`) fixes the issue in both Triton 25.04 and 25.12.

**Strong suspected root cause (core):** `SequenceStates::OutputState()` pairs output-state to input-state [by index in two different `std::map`s](<https://github.com/triton-inference-server/core/blob/r25.12/src/sequence_state.cc#L309>) (distance/advance), rather than by the configured `(input_name, output_name)` mapping. That index-based mapping is incorrect when the two maps’ key ordering differs - exactly what happens with names like `cache_last_channel_next` vs `cache_last_channel_len_next`.

---

### Host / HW

* GPU: NVIDIA L40S
* Driver: 535.274.02
* CUDA: 12.2

### Triton containers tested

* `nvcr.io/nvidia/tritonserver:25.04-py3`
  * Triton Server: 2.57.0
  * TensorRT: 10.9.0
* `nvcr.io/nvidia/tritonserver:25.12-py3`
  * Triton Server: 2.64.0
  * TensorRT: 10.14.1

### Client

* Python `tritonclient` gRPC
* Uses `sequence_id` with `sequence_start` / `sequence_end`

### Model

* TensorRT plan built from ONNX export
* Implicit state via `sequence_batching { state [...] }`

---

### State inputs (implicit)

* `cache_last_time`: FP16, `(B, 18, 384, 30)`
* `cache_last_channel`: FP16, `(B, 32, 8, 50)`
* `cache_last_channel_len`: INT32 or INT64, `(B, 1)`

### State outputs (implicit)

* `cache_last_time_next`
* `cache_last_channel_next`
* `cache_last_channel_len_next`

---

## Triton model config snippet (relevant part)

```protobuf
sequence_batching {
  max_sequence_idle_microseconds: 15000000
  oldest { max_queue_delay_microseconds: 10000 max_candidate_sequences: 4096 }

  state: [
    {
      input_name: "cache_last_channel_len"
      output_name: "cache_last_channel_len_next"
      data_type: TYPE_INT64  # also reproduced with INT32
      dims: [ 1 ]
      initial_state: { name: "cache_last_channel_len_initial" data_type: TYPE_INT64 dims: [ 1 ] zero_data: true }
    },
    {
      input_name: "cache_last_channel"
      output_name: "cache_last_channel_next"
      data_type: TYPE_FP16
      dims: [ 32, 8, 50 ]
      initial_state: { name: "cache_last_channel_initial" data_type: TYPE_FP16 dims: [ 32, 8, 50 ] zero_data: true }
    },
    {
      input_name: "cache_last_time"
      output_name: "cache_last_time_next"
      data_type: TYPE_FP16
      dims: [ 18, 384, 30 ]
      initial_state: { name: "cache_last_time_initial" data_type: TYPE_FP16 dims: [ 18, 384, 30 ] zero_data: true }
    }
  ]
}
```

---

## Steps to reproduce

1. Export ONNX that includes the three states (time/channel/channel_len).
2. Build TensorRT engine from ONNX (`trtexec`).

**I could provide resources if needed**

3. Start Triton with verbose logs enabled, e.g.:

   ```bash
   tritonserver --model-repository=/models --log-verbose=4
   ```
4. Run a minimal client that:
   * sends only the non-state inputs (`audio_signal`, `length`)
   * uses `sequence_id`
   * sets `sequence_start=True` on step 0
   * requests output state tensors (`cache_last_*_next`) to confirm correctness
5. Observe:
   * step 0 succeeds and returns correct shapes/dtypes for all outputs
   * step 1 fails with an invalid shape error for `cache_last_channel` (4D expected, 2D provided)

---

## Expected behavior

Across requests in the same sequence, Triton should:

* feed `cache_last_channel_next` from request *i* into `cache_last_channel` for request *i+1*
* feed `cache_last_channel_len_next` from request *i* into `cache_last_channel_len` for request *i+1*
* never swap or alias state buffers between different state entries

---

## Actual behavior

* Request 0 (step 0) succeeds, state outputs look correct.
* Request 1 (step 1) fails.
* Client error:

  ```
  [StatusCode.INTERNAL] request specifies invalid shape for input 'cache_last_channel' ...
  model expected 4 dimensions but received 2 dimensions
  ```
* Server error (25.12):

  ```
  instance_state.cc:767] "error setting the binding dimension"
  ```
* High-verbosity logs show that on the failing request, the injected override inputs are swapped:
  * `cache_last_channel` becomes INT64/INT32 with shape `[1,1]`
  * `cache_last_channel_len` becomes FP16 with shape `[1,32,8,50]`

---

## Evidence (log excerpt showing the swap on the 2nd request)

On request 2 (step 1), Triton logs show:

```text
added input override for cache_last_channel:
  input: cache_last_channel, type: INT64, original shape: [1,1], ...

added input override for cache_last_channel_len:
  input: cache_last_channel_len, type: FP16, original shape: [1,32,8,50], ...
...
E... instance_state.cc:767] "error setting the binding dimension"
```

This is consistent with the model expecting `cache_last_channel` to be 4D FP16, but receiving the 2D integer length state instead.

---

## Key observation / workaround

Renaming the length state to avoid a prefix extension of another state name resolves the issue:

* **Fails:** `cache_last_channel` + `cache_last_channel_len`
* **Works:** `cache_last_channel` + `cache_last_chan_len` (or any non-prefix-colliding name)

This fixes the issue in **both**:

* `25.04-py3` (Triton 2.57.0)
* `25.12-py3` (Triton 2.64.0)

Notably, changing the length dtype (INT32 ↔ INT64) does **not** fix the issue. The bug appears related to **name-based ordering/mapping**, not dtype.

---

## Suspected root cause (core r25.12): index-based pairing across `std::map`

[https://github.com/triton-inference-server/core/blob/r25.12/src/sequence_state.cc#L309](<https://github.com/triton-inference-server/core/blob/r25.12/src/sequence_state.cc#L309>)<br>In `triton-inference-server/core` (branch `r25.12`), the suspected problem is in `sequence_state.cc`:

* `SequenceStates::OutputState()` finds an output state by name in `output_states_` (`std::map` keyed by `output_name`).
* It then computes an index via `std::distance(output_states_.begin(), it)` and uses that index to select the “corresponding” input state from `input_states_` by advancing an iterator by the same amount.
* But `input_states_` is a **different** `std::map`, keyed by `input_name`.

This implicitly assumes that `input_states_` and `output_states_` have identical key ordering. That is not generally true and fails deterministically when the sorted key order differs.

### Why the prefix-collision example deterministically swaps

Given:

Input keys (sorted by `std::map`):

1. `cache_last_channel`
2. `cache_last_channel_len`
3. `cache_last_time`

Output keys (sorted by `std::map`):

1. `cache_last_channel_len_next`
2. `cache_last_channel_next`
3. `cache_last_time_next`

Because `"cache_last_channel_len_next" < "cache_last_channel_next"` lexicographically, indices don’t align.

So:

* `OutputState("cache_last_channel_next")` is index 1 in `output_states_`<br>→ picks index 1 in `input_states_`<br>→ **maps to** `cache_last_channel_len` (wrong)
* `OutputState("cache_last_channel_len_next")` is index 0 in `output_states_`<br>→ picks index 0 in `input_states_`<br>→ **maps to** `cache_last_channel` (wrong)

This exactly matches the observed swap in the logs.

---

## Additional notes

* Repro is independent of integer dtype (INT32 and INT64 both reproduce).
* Repro is present in both `25.04` and `25.12` containers.
* Renaming the state to avoid prefix/lexicographic collision is an effective workaround but not acceptable as a general requirement.

notandruu:
#### What this PR does
Fixes implicit sequence state updates when configured input and output state names sort differently.

`SequenceStates::OutputState()` previously found the output state by `output_name`, computed its index in `output_states_`, and advanced to that same index in `input_states_`. Both containers are `std::map`s keyed by different names, so the index only works when input and output names happen to have the same lexicographic order.

This PR uses the configured `SequenceState::OtherState()` pairing established during initialization instead, and preserves that pairing when null sequence states are copied.

#### Why
For names such as:

- input: `cache_last_channel`, output: `cache_last_channel_next`
- input: `cache_last_channel_len`, output: `cache_last_channel_len_next`

`cache_last_channel_len_next` sorts before `cache_last_channel_next`, while the input names sort in the opposite order. The old index-based lookup can update the wrong cached input state and inject the wrong tensor on the next sequence request.

#### Tests
Added `sequence_state_test` covering mismatched input/output lexicographic ordering.

Local validation:

- `clang-format -i src/sequence_state.cc src/test/sequence_state_test.cc`
- `git diff --check`
- `cmake -S . -B build-codex -DTRITON_CORE_HEADERS_ONLY=OFF -DTRITON_ENABLE_GPU=OFF -DTRITON_ENABLE_METRICS_GPU=OFF -DTRITON_ENABLE_S3=OFF -DTRITON_ENABLE_GCS=OFF -DTRITON_ENABLE_AZURE_STORAGE=OFF`
- `cmake --build build-codex --target triton-core -j4` could not complete on this Apple Silicon host because third-party Abseil is built with `-msse4.1`, which AppleClang rejects for `arm64-apple-darwin`. The failure happens before compiling Triton core sources.

Fixes triton-inference-server/server#8586.

### DIFF
--- a/src/sequence_state.cc
+++ b/src/sequence_state.cc
@@ -35,14 +35,18 @@
 
 namespace triton { namespace core {
 
-SequenceState::SequenceState() : data_(new MemoryReference) {}
+SequenceState::SequenceState()
+    : data_(new MemoryReference), other_state_(nullptr)
+{
+}
 
 SequenceState::SequenceState(
     const std::string& name, const inference::DataType datatype,
     const int64_t* shape, const uint64_t dim_count, bool use_single_buffer,
     bool use_growable_memory)
     : name_(name), datatype_(datatype), shape_(shape, shape + dim_count),
-      data_(new MemoryReference), use_single_buffer_(use_single_buffer),
+      data_(new MemoryReference), other_state_(nullptr),
+      use_single_buffer_(use_single_buffer),
       use_growable_memory_(use_growable_memory)
 {
 }
@@ -52,7 +56,8 @@ SequenceState::SequenceState(
     const std::vector<int64_t>& shape, bool use_single_buffer,
     bool use_growable_memory)
     : name_(name), datatype_(datatype), shape_(shape),
-      data_(new MemoryReference), use_single_buffer_(use_single_buffer),
+      data_(new MemoryReference), other_state_(nullptr),
+      use_single_buffer_(use_single_buffer),
       use_growable_memory_(use_growable_memory)
 {
 }
@@ -294,36 +299,37 @@ SequenceStates::OutputState(
   }
 
   std::vector<int64_t> new_shape(shape, shape + dim_count);
-  *output_states_[name]->MutableDType() = datatype;
-  *output_states_[name]->MutableShape() = new_shape;
+  auto& output_state_r = output_state_itr->second;
+  *output_state_r->MutableDType() = datatype;
+  *output_state_r->MutableShape() = new_shape;
 
-  auto& output_state_r = output_states_[name];
-  size_t iter_advance =
-      std::distance(output_states_.begin(), output_states_.find(name));
-
-  // Find the input state corresponding to this output state.
-  auto input_states_itr = input_states_.begin();
-  std::advance(input_states_itr, iter_advance);
-  auto& input_state_r = input_states_[input_states_itr->first];
-  bool use_single_buffer = output_states_[name]->UseSingleBuffer();
-
-  if (output_state != nullptr) {
-    *output_state = output_states_[name].get();
+  SequenceState* input_state = output_state_r->OtherState();
+  if (input_state == nullptr) {
+    return Status(
+        Status::Code::INTERNAL,
+        "state '" + name + "' does not have a corresponding input state.");
   }
 
-  output_state_r->SetStateUpdateCallback(
-      [&output_state_r, &input_state_r, use_single_buffer]() {
+  SequenceState* output_state_ptr = output_state_r.get();
+  bool use_single_buffer = output_state_r->UseSingleBuffer();
+
+  if (output_state != nullptr) {
+    *output_state = output_state_ptr;
+  }
+
+  output_state_ptr->SetStateUpdateCallback(
+      [output_state_ptr, input_state, use_single_buffer]() {
         // Swap the internal memory if the size of the input and output state is
         // equal
 
         if (!use_single_buffer) {
-          if (output_state_r->Data()->TotalByteSize() ==
-              input_state_r->Data()->TotalByteSize()) {
-            std::shared_ptr<Memory> temp_memory = input_state_r->Data();
-            RETURN_IF_ERROR(input_state_r->RemoveAllData());
-            RETURN_IF_ERROR(input_state_r->SetData(output_state_r->Data()));
-            RETURN_IF_ERROR(output_state_r->RemoveAllData());
-            RETURN_IF_ERROR(output_state_r->SetData(temp_memory));
+          if (output_state_ptr->Data()->TotalByteSize() ==
+              input_state->Data()->TotalByteSize()) {
+            std::shared_ptr<Memory> temp_memory = input_state->Data();
+            RETURN_IF_ERROR(input_state->RemoveAllData());
+            RETURN_IF_ERROR(input_state->SetData(output_state_ptr->Data()));
+            RETURN_IF_ERROR(output_state_ptr->RemoveAllData());
+            RETURN_IF_ERROR(output_state_ptr->SetData(temp_memory));
           } else {
             // If the size of output state is different from the input state,
             // allocate a new memory for the input state with the same size as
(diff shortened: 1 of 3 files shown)


notandruu:
Ping — happy to make any changes needed.

PR #507, opened 2026-07-02
fix(TRI-1527): don't vendor libtritonserver stub into python wheel

mc-nv:
## Description

The 26.06 `tritonserver` wheel's in-process Python API crashes on the first C-API call:

```
python3 -c "from tritonserver._c.triton_bindings import TRITONSERVER_ServerOptions; TRITONSERVER_ServerOptions()"
# UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf3 in position 0
```

`auditwheel repair` vendored the core API **stub** (`libtritonserver.so`, ~77 KB of empty
functions built from `src/tritonserver_stub.cc`) into `tritonserver.libs/` instead of
leaving `libtritonserver.so` to resolve at runtime. The extension — built against the real
ABI — then calls empty stub functions, and pybind decodes a garbage return pointer
(machine code `f3 0f 1e fa` = `endbr64`) as UTF-8, raising `UnicodeDecodeError` on the
first `TRITONSERVER_*` call.

The stub exists only to satisfy symbols at link time; the real 34 MB library (with its
CUDA/backend machinery) is loaded from `/opt/tritonserver/lib` in the container at runtime
and must never be vendored.

## Changes

`python/build_wheel.py`:
- `auditwheel repair --exclude libtritonserver.so` — do not vendor the stub; keep
  `libtritonserver.so` as an unvendored `NEEDED` entry resolved at runtime.
- New `_ensure_runtime_libdir_rpath()` — auditwheel drops the original
  `/opt/tritonserver/lib` rpath, so re-add it via `patchelf` and repack with `wheel pack`
  (regenerates `RECORD`). No-op if already present.

## Test plan

- Reproduced on `nvcr.io/nvidia/tritonserver:26.06-py3` (amd64).
- With the fixed config, `import` + `TRITONSERVER_ServerOptions()` returns OK (real library
  resolved from `/opt/tritonserver/lib`).
- Mechanical test of the new helper (`wheel unpack` -> `patchelf --set-rpath` ->
  `wheel pack`): repacked wheel installs (RECORD valid) and imports cleanly.
- Pre-commit (isort/black/flake8) passes.

#### Related Issues:
- Resolves: TRI-1527

### DIFF
--- a/python/build_wheel.py
+++ b/python/build_wheel.py
@@ -34,7 +34,7 @@ import subprocess
 import sys
 import zipfile
 from distutils.dir_util import copy_tree
-from tempfile import mkstemp
+from tempfile import mkdtemp, mkstemp
 
 # ANSI colors for CI log readability (rendered by GitLab CI, harmlessly
 # inert in non-ANSI viewers). Suppressed when stderr is redirected to a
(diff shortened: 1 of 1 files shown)


python/build_wheel.py
```diff
@@ -219,6 +219,76 @@ def _wheel_has_so(wheel_path):
...
+            ["python3", "-m", "wheel", "unpack", wheel_path, "-d", tmp],
+            capture_output=True,
+            text=True,
+        )
+        fail_if(r.returncode != 0, f"wheel unpack failed: {r.stderr}")
+        unpacked = next(p for p in pathlib.Path(tmp).iterdir() if p.is_dir())
+        changed = False
+        for so in unpacked.rglob("*.so"):
+            needed = subprocess.run(
+                ["patchelf", "--print-needed", str(so)],
```
nv-rinig, line 261: should we pass in `check=True` here in case something goes wrong parsing/reading the file?

mc-nv: Redundant as it explicitly checking the `*.so` files

nv-rinig (approved):

whoisj (approved):

PR #518, opened 2026-07-30
fix: Do not autofill default_model_filename for Python-runtime PyTorch models

mattwittwer:
#### What does the PR do?

Since #484, `AutoCompleteBackendFields` autofills `default_model_filename: "model.pt"` for every `backend: "pytorch"` model with an empty platform — including models using the Python-based runtime (`runtime: "model.py"`). For those models a sibling `model.pt` is the weights companion to the `model.py` class, and the runtime gives `default_model_filename` top priority, so it calls `torch.jit.load()` on the weights file and fails at load with `PytorchStreamReader failed locating file constants.pkl`.

Restores the pre-#484 guard — skip filename autofill when `runtime` is `"model.py"` — and extends it to the AOTI platform branch (`model.pt2`), which the Python runtime cannot load either. Output is unchanged for every other config. #484's stated goal (correct .pt2 selection for torch_aoti) is preserved and now pinned by tests.

Adds `AutoCompleteBackendFieldsTest` to `repo_agent_test` (which already compiles `model_config_utils.cc`): the two Python-runtime regression cases plus six cases pinning existing autofill and version-dir sniffing behavior. The regression cases fail without the fix and pass with it.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:

https://github.com/triton-inference-server/server/pull/8913

#### Test plan:

Test case added [here](https://github.com/triton-inference-server/server/pull/8913) 

- CI Pipeline ID:

60326198

- closes GitHub issue: #xxx


src/model_config_utils.cc
```diff
@@ -1259,17 +1259,22 @@ AutoCompleteBackendFields(
   // When we know the backend is PyTorch, we set the platform and default model
   // filename as necessary.
   if (config->backend() == kPyTorchBackend) {
+    // A non-default runtime resolves its own model file. The Python-based
+    // PyTorch runtime ("model.py") treats a sibling "model.pt" as the weights
+    // companion to the model class, so autofilling this field would make it
+    // load the weights file as the model itself.
+    const bool autofill_filename = config->default_model_filename().empty() &&
```
yinggeh: `autofill_filename` sounds like a string type instead of boolean. Rename to `use_autofill_filename` instead.

mattwittwer: Updated

src/test/repo_agent_test.cc
```diff
@@ -2359,6 +2363,165 @@ TEST_F(TritonRepoAgentAPITest, TRITONREPOAGENT_AgentState)
...
+
+  std::string model_path_;
+  std::string version_path_;
+};
+
+// A model with backend "pytorch" + runtime "model.py" and both "model.py"
+// and a weights "model.pt" in the version directory must not have "model.pt"
+// autofilled as the default model filename, otherwise the runtime attempts
+// torch.jit.load() on the weights file.
+TEST_F(AutoCompleteBackendFieldsTest, PythonRuntimeLibTorchKeepsFilenameEmpty)
```
yinggeh: Is it possible to merge test cases into a for loop?

mattwittwer: Yes, updated

yinggeh (approved):

PR #527, opened 2026-09-11
build: migrate tritonserver wheel to hatchling with PEP 639 metadata

mc-nv:
#### What does the PR do?

- Pins the build backend so wheel metadata no longer depends on whatever the build image ships.
- `setuptools` <77 wrote a `License-File` field while still declaring `Metadata-Version: 2.1` — the combination PyPI/Kitmaker Warehouse rejects. The wheel now emits 2.4.
- Migrates to hatchling; the `BinaryDistribution` shim and the `stubgen` `build_py` subclass move to `python/hatch_build.py`.
- `license` becomes SPDX `BSD-3-Clause` + `license-files`; adds `requires-python = ">=3.12"`.

`pyproject.toml` and `python/hatch_build.py`.

#### Test plan:
Verified against a staged build tree: correct `cp312-cp312-linux_<arch>` tag, Metadata-Version 2.4, `License-Expression`, `License-File`, `Requires-Python`, `_api`/`_c` packages and `py.typed` all present, `twine check` passes.

#### Caveats:
`stubgen` **output content** needs the real compiled extension, so only the mechanism was verified locally — it must be confirmed in CI.

#### Related PRs:
- triton-inference-server/server#8970
- triton-inference-server/client#927
- triton-inference-server/model_analyzer#1038
- triton-inference-server/perf_analyzer#469
- triton-inference-server/triton_cli#130

#### Related Issues:
- Resolves: TRI-1775
<sup>
CI (internal): [#67460724](http://tritonserver.local/ci/pipelines/67460724)<br/>
</sup>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,13 @@
 name = "tritonserver"
 authors = [{ name = "NVIDIA Inc.", email = "sw-dl-triton@nvidia.com" }]
 description = "Triton Inference Server In-Process Python API"
-license = { file = "LICENSE.txt" }
+readme = { content-type = "text/markdown", text = """See the [Triton Inference Server](https://github.com/triton-inference-server/server) repository for package details. `tritonserver` provides the in-process Python API for embedding Triton directly in a Python application.""" }
+# PEP 639 SPDX expression, emitted as License-Expression at
+# Metadata-Version 2.4. The matching "License ::" trove classifier must stay
+# removed: declaring both is rejected outright by the build backend.
+license = "BSD-3-Clause"
+license-files = ["LICENSE.txt"]
+requires-python = ">=3.12"
 dynamic = ["version"]
 dependencies = ["numpy<2"]
 classifiers = [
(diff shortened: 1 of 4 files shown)


python/hatch_build.py
```diff
@@ -25,43 +24,34 @@
...
-    setup(distclass=BinaryDistribution, cmdclass={"build_py": BuildPyCommand})
+    def initialize(self, version, build_data):
+        build_data["pure_python"] = False
+        build_data["infer_tag"] = True
+        # Written into the package tree (self.root is the wheel build root)
+        # so hatchling collects the stubs alongside the extension.
+        subprocess.run(
+            ["stubgen", "-p", "tritonserver._c", "-o", self.root],
+            check=True,
+        )
```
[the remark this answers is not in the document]
mc-nv: Agreed, and this matches the caveat in the PR description: the stubgen path is the one thing that cannot be verified outside a real build. Locally I could only confirm the *mechanism* — stubgen runs from the isolated build env and its output is collected into the wheel — because the stub content needs the compiled extension.

Deferring the end-to-end packaging test rather than adding it here: it needs a built `triton_bindings.*.so`, so it belongs with the QA packaging tests rather than in this build-config change. Tracked under TRI-1775.

Separately, `f939daa` addresses a related gap you did not flag: this was invoking `python -m build` without `--wheel`, so it built an sdist and then the wheel from it. Now builds the wheel directly.

whoisj (approved):

pyproject.toml
```diff
@@ -41,38 +47,41 @@ classifiers = [
...
     "Topic :: Software Development :: Libraries",
     "Topic :: Utilities",
-    "License :: OSI Approved :: BSD License",
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.12",
     "Environment :: Console",
     "Natural Language :: English",
-    "Operating System :: OS Independent",
+    "Operating System :: POSIX :: Linux",
```
whoisj, line 54: is this a breaking change?

pyproject.toml
```diff
@@ -41,38 +47,41 @@ classifiers = [
     "Topic :: Scientific/Engineering :: Artificial Intelligence",
     "Topic :: Software Development :: Libraries",
     "Topic :: Utilities",
-    "License :: OSI Approved :: BSD License",
```
whoisj, line 44: no license classifier?

pyproject.toml
mc-nv: No, it's just a classifier update. 
`"Operating System :: OS Independent"` states that we are using pure-python application with no extension which is wrong, that's why it was updated

pyproject.toml
mc-nv: Since we changed the metadata-version some fields also been updated, in metadata 2.4 license field now have it's own key, and it was moved to line 35

